PR #21, opened 2014-12-10
Add maxRedirects option

floatdrop:
Move hard-coded maximum redirects number to options.


floatdrop:
Oops. Didn't mean to close it, but whatever.

PR #30, opened 2015-01-02
Set maxSockets to Infinity by default

### LINKED ISSUE #22: Default agent is limited with 5 sockets
~~It seems like there an open socket left after got has redirect reply from server. Setting `{agent: false}` fixes test, but feels like it should be working by default.~~

Default http agent in node is limited by 5 open sockets. Read more in [hyperquest readme](https://github.com/substack/hyperquest#pooling-is-evil).

floatdrop:
Setting agent for requests to new http(s).Agent({maxSockets: Infinity})

This will get same maxSockets value in 0.10 and upcoming 0.12 node as @julien-f suggested.

Closes #22

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,15 @@
 'use strict';
+
+var assign = require('object-assign');
+var agent = require('infinity-agent');
+var duplexify = require('duplexify');
 var http = require('http');
 var https = require('https');
-var urlLib = require('url');
-var zlib = require('zlib');
-var duplexify = require('duplexify');
-var assign = require('object-assign');
+var isReadableStream = require('isstream').isReadable;
 var read = require('read-all-stream');
 var timeout = require('timed-out');
-var isReadableStream = require('isstream').isReadable;
+var urlLib = require('url');
+var zlib = require('zlib');
 
 function got(url, opts, cb) {
 	if (typeof opts === 'function') {
@@ -54,6 +56,11 @@ function got(url, opts, cb) {
 		var parsedUrl = urlLib.parse(url);
 		var fn = parsedUrl.protocol === 'https:' ? https : http;
 		var arg = assign({}, parsedUrl, opts);
+
+		// TODO: remove this when Node 0.10 will be deprecated
+		if (arg.agent === undefined) {
+			arg.agent = agent(arg);
+		}
 
 		var req = fn.request(arg, function (response) {
 			var statusCode = response.statusCode;
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   ],
   "dependencies": {
     "duplexify": "^3.2.0",
+    "infinity-agent": "^0.1.0",
     "isstream": "^0.1.1",
     "object-assign": "^2.0.0",
     "read-all-stream": "^0.1.0",
--- a/readme.md
+++ b/readme.md
@@ -75,6 +75,16 @@ Type: `number`
 
 Milliseconds after which the request will be aborted and an error event with `ETIMEDOUT` code will be emitted.
 
+##### options.agent
+
+[http.Agent](http://nodejs.org/api/http.html#http_class_http_agent) instance.
+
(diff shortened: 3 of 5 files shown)


floatdrop:
Squashed commits. Ready for review.

readme.md
```diff
@@ -75,6 +75,16 @@ Type: `number`
...
 
+##### options.agent
+
+[http.Agent](http://nodejs.org/api/http.html#http_class_http_agent) instance.
+
+Node HTTP/HTTPS [Agent in 0.10.35](https://github.com/joyent/node/blob/v0.10.35-release/lib/http.js#L1261) by default limits number of open sockets to 5 — which is a little low. By default got will use `new Agent({maxSockets: Infinity})` like new [Agent in 0.11.14](https://github.com/joyent/node/blob/v0.11.14-release/lib/_http_agent.js#L110).
+
+You can read more about in [why pooling is evil](https://github.com/substack/hyperquest#pooling-is-evil).
+
+To use default agent just pass `undefined` to this option.
```
sindresorhus: Better to recommend `null` rather than `undefined`.

index.js
```diff
@@ -9,6 +9,9 @@ var read = require('read-all-stream');
 var timeout = require('timed-out');
 var isReadableStream = require('isstream').isReadable;
 
+var httpAgent = new http.Agent({maxSockets: Infinity});
+var httpsAgent = new https.Agent({maxSockets: Infinity});
```
sindresorhus: Can you add a `TODO` about removing this when Node 0.10 is not supported anymore?

test/test-https.js
```diff
@@ -51,9 +52,19 @@ tape('setup', function (t) {
...
+		t.ok(data);
+		t.end();
+	});
+});
+
+tape('make request to https server with ca', function (t) {
 	got(s.url, {
 		strictSSL: true,
-		ca: caRootCert,
+		agent: new https.Agent({ca: caRootCert}),
```
floatdrop: I'm only worried about this - `ca` option should be passed to `tls.connect`, but it does not.

floatdrop:
Rebased.

test/test-https.js
sindresorhus: Hmm, yeah, that's not good as users would expect that to work.

floatdrop: Node [`https.js`](https://github.com/joyent/node/blob/v0.10.35-release/lib/https.js#L109) has this code to achieve this:

```js
  if (typeof options.agent === 'undefined') {
    if (typeof options.ca === 'undefined' &&
        typeof options.cert === 'undefined' &&
        typeof options.ciphers === 'undefined' &&
        typeof options.key === 'undefined' &&
        typeof options.passphrase === 'undefined' &&
        typeof options.pfx === 'undefined' &&
        typeof options.rejectUnauthorized === 'undefined') {
      options.agent = globalAgent;
    } else {
      options.agent = new Agent(options);
    }
  }
```

Should we port this logic as well? It's, kind of, sucks.

sindresorhus: Lol. But yeah... Lets do it.

floatdrop:
Incapsulated Node logic of creating https agent to infinity-agent module.

index.js
```diff
@@ -1,13 +1,14 @@
...
+var assign				= require('object-assign');
+var agent				= require('infinity-agent');
+var duplexify			= require('duplexify');
+var http				= require('http');
+var https				= require('https');
+var isReadableStream	= require('isstream').isReadable;
+var read				= require('read-all-stream');
+var timeout				= require('timed-out');
+var urlLib				= require('url');
+var zlib				= require('zlib');
```
sindresorhus: formatting yak

package.json
```diff
@@ -32,6 +32,7 @@
   ],
   "dependencies": {
     "duplexify": "^3.2.0",
+    "infinity-agent": "^0.1.0",
```
sindresorhus: :heart:

index.js
floatdrop: fixed

sindresorhus:
```
$ npm i -g superb
$ superb
groovy
```

floatdrop:
\o/

PR #31, opened 2015-01-03
Discard response before throw

floatdrop:
This will not leave hanging sockets for 2 minutes (if server does not provide `Connection: close` header).

Downside of it - you will not get content of 3xx response, but almost always it does not contain anything.


sindresorhus:
I'm fine with that tradeoff.

PR #33, opened 2015-01-16
Various tweaks

kevva:
Some code tweaks.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,19 +13,18 @@ var prependHttp = require('prepend-http');
 
 function got(url, opts, cb) {
 	if (typeof opts === 'function') {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -34,25 +33,20 @@ function got(url, opts, cb) {
 
 	// returns a proxy stream to the response
 	// if no callback has been provided
-	var proxy;
 	if (!cb) {
 		proxy = duplexify();
 
-		// forward errors on the stream
```
sindresorhus: this comment is somewhat useful though

kevva: hm, yeah, thought it was kinda self explanatory, but we can keep it!

Issue #40, opened 2015-03-16
`json` option?


sindresorhus:
`request` has this. Could be a nice shortcut.

It basically `JSON.stringify()` the input and sets the header `'content-type': 'application/json'` and `JSON.parse()` the output.

@kevva @floatdrop @julien-f Thoughts?

kevva:
:+1: I like it.

floatdrop:
:+1: 

P.s. just don't make it silent like [request](https://github.com/request/request/blob/9f844e29862a26771352f2a4f31d658da872af62/request.js#L1181-L1183).

sindresorhus:
Hah, I saw that too. Of course not.

julien-f:
Good idea, :+1:

floatdrop:
If someone have ideas, how this option should behave in Stream mode, please share - I can't think nothing better than making it exclusive for callback mode.

sindresorhus:
@floatdrop :+1:

kevva:
@floatdrop, by using https://github.com/dominictarr/JSONStream.

floatdrop:
@kevva awesome module, but I tend to leave streaming on feature releases because:
 * Streaming JSON is rare
 * It needs some additional options (for `parse` and `stringify`/`stringifyObject` calls)
 * It will not break compatibility, if we add it later

Now I'm digging through [issues](https://github.com/request/request/issues/718) with `json`-like options in other modules and implementing it right might be harder, than it looks (or I'm just over-complicating).

kevva:
Yeah, haven't had a chance to look at it myself, but I agree we can leave this option for streams to future versions.

floatdrop:
After reading some issues with `json` option in request one idea came up. All this noise around `json` option comes from the fact, that enabling it have two separate effects:

 1. input - it enables serializing of `body` (and add headers, if they not set)
 2. output - parses incoming JSON

So, why don't just split them apart? Now, if we pass `Object` in `body`, we will get next error: `TypeError: first argument must be a string or Buffer`. What if we serialize it to JSON in this case (without setting any options)? Anyway we can't send it as is, so this seems reasonable. But two things needs to be considered:

 1. is it true, that most users want to serialize objects in `body` option instead of throwing error?
 2. if so, should we take care of `toString` property as well?

floatdrop:
Partially implemented in `2.7.0` :tada:

PR #44, opened 2015-04-05
Implement serializing Object in body to JSON

floatdrop:
First part of #40 - serializing `body` to JSON.

Passing `Object` to `body` option gives `TypeError: first argument must be a string or Buffer`, when it could serialize `Object` to JSON and send it.

__Cons:__ user will now about bad `body` much later, but if user passing `Object` to body (and it is not `Buffer`/`String`/`Stream`) then in most cases it should be stringified to JSON.

__Pros:__ shorter code and more flexible `json` option.

```js
got('jsonendpoint.com', {
    body: JSON.stringify({ data: 'dog' }), 
    headers: { 'content-type': 'application/json' }, 
function () { });

// vs

got('jsonendpoint.com', { 
    body: { data: 'dog' } }, 
function () { });
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,6 +23,10 @@ function GotError(message, nested) {
 util.inherits(GotError, NestedError);
 GotError.prototype.name = 'GotError';
 
+function isSendable(body) {
+	return Buffer.isBuffer(body) || typeof body === 'string' || body instanceof String;
+}
+
 function got(url, opts, cb) {
 	if (typeof opts === 'function') {
 		cb = opts;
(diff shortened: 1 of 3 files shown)


floatdrop:
In before question `Why serializing is not enabled only when json option is true?`:

There are cases, when server response is not JSON at all, but it expects JSON in request.

floatdrop:
After few more hours of thinking - writing `JSON.stringify` with right `header` is not that hard:

```js
got('jsonendpoint.com', {
    body: JSON.stringify({ data: 'dog' }), 
    headers: { 'content-type': 'application/json' }, 
function () { });
```

Auto-stringifying seems to me much more dangeuros, than one line in `got` call.

kevva:
> Auto-stringifying seems to me much more dangeuros, than one line in got call.

+1

sindresorhus:
:+1: Good calls all around :)

PR #45, opened 2015-04-05
Add json option

floatdrop:
Second part of #40

Parsing JSON in response.

Option ignores `Content-Type` header, because they are often not set by server. Errors from `JSON.parse` wrapped and pushed to callback with raw data.

Before:

```js
got('jsonendpoint.com', function (err, data) {
    if (err) { return cb(err); }

    var json;

    try {
        json = JSON.parse(data);
    } catch (e) {
        return cb(new Error('Reponse from jsonendpoint.com broken: ' + e.message));
    }
    
    // working with json
});
```

After:

```js
got('jsonendpoint.com', {json: true}, function (err, json) {
    if (err) { return cb(err); }

    // working with json
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,11 +40,13 @@ function got(url, opts, cb) {
 
 	var encoding = opts.encoding;
 	var body = opts.body;
+	var json = opts.json;
 	var proxy;
 	var redirectCount = 0;
 
 	delete opts.encoding;
 	delete opts.body;
+	delete opts.json;
 
 	if (body) {
 		opts.method = opts.method || 'POST';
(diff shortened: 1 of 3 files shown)


floatdrop:
Ready to merge. @sindresorhus @kevva @julien-f ?

floatdrop:
_To my future self:_ I intentionally leave stratification out of this option, because it will block next usecase: user wants to parse JSON response, but send Stream (or Buffer, or string) of data. 

So we could say: yup, we can check, that it is sendable type and do not stringify on it, but string is in both categories - it can be sended as it and it can be stringified to `"value"`.

kevva:
Great work!

PR #47, opened 2015-04-08
Parse non-200 responses when options.json is true

### LINKED ISSUE #46: Parsing non-200 responses
For now we [don't have](https://github.com/sindresorhus/got/blob/master/index.js#L108-L112) parsing on responses, that are not okay by them selves:

```js
read(res, encoding, function (err, data) {
	err = new GotError(url + ' response code is ' + statusCode + ' (' + status[statusCode] + ')', err);
	err.code = statusCode;
	cb(err, data, response);
});
```

But it is not convinient, when API returns JSON (like [github does](https://developer.github.com/v3/#client-errors)) and we need to parse it by hands.

On other side we have huge amount of services, that proxy'ed and return 50x codes with default nginx error page.

So question is - maybe we should parse JSON on non-200 responses silently? Anyway `got` is already in error state.

floatdrop:
Closes #46

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ var NestedError = require('nested-error-stacks');
 
 function GotError(message, nested) {
 	NestedError.call(this, message, nested);
-	objectAssign(this, nested);
+	objectAssign(this, nested, {nested: this.nested});
 }
 
 util.inherits(GotError, NestedError);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -17,7 +17,7 @@ var NestedError = require('nested-error-stacks');
 
 function GotError(message, nested) {
 	NestedError.call(this, message, nested);
-	objectAssign(this, nested);
+	objectAssign(this, nested, {nested: this.nested});
```
floatdrop: `Do erase nested errors.`

floatdrop:
Ready to merge. @kevva @sindresorhus

sindresorhus:
:+1: :shipit:

floatdrop:
@sindresorhus I'm planning to release patch release `2.7.1`, since in readme we never specified, that only successful responses will be parsed, so it is a bug. Is it ok?

sindresorhus:
:+1:

PR #48, opened 2015-04-19
Use backported agents from Node core

floatdrop:
So... Long story short: in Node 0.10 with `maxSockets` set to `Infinity` default Agent will have `shouldKeepAlive` flag set on - which is pretty bad (because sockets will be not reused), but remote system will think, that we will.

What have I done (!?):

 * [infinity-agent](https://github.com/floatdrop/infinity-agent) was bumped to `2.0.0` with backported Agent from Node core
 * Little [patch](https://github.com/floatdrop/infinity-agent/blob/master/http.js#L135-L140) was added to ported code, that disables `shouldKeepAlive` (same as [http module does](https://github.com/joyent/node/blob/master/lib/_http_client.js#L149-L152) in recent Node) in case of `maxSockets === Infinity`

Since `infinity-agent` now is just a backport of Node core Agent, [code under TODO](https://github.com/sindresorhus/got/compare/infinity-agent-2.0?expand=1#diff-168726dbe96b3ce427e7fedce31bb0bcR79) is now in module. It looks bad, but there is nothing, that we can do about it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ var urlLib = require('url');
 var util = require('util');
 var zlib = require('zlib');
 var objectAssign = require('object-assign');
-var agent = require('infinity-agent');
+var infinityAgent = require('infinity-agent');
 var duplexify = require('duplexify');
 var isStream = require('is-stream');
 var readAllStream = require('read-all-stream');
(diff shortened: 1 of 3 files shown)


floatdrop:
Also `infinity-agent` now ignores `http.Agent.defaultMaxSockets` and uses `Infinity` always.

sindresorhus:
Lgtm. Glad to see it's now more aligned with Node. I don't mind the verbosity. It's going away at some point anyways :)

PR #49, opened 2015-04-20
Add got-promise to related

floatdrop:


readme.md
```diff
@@ -159,8 +159,8 @@ got('todomvc.com', {
 
 ## Related
 
-[`gh-got`](https://github.com/sindresorhus/gh-got) - Convenience wrapper for interacting with the GitHub API
-
+ * [`gh-got`](https://github.com/sindresorhus/gh-got) - Convenience wrapper for interacting with the GitHub API
+ * [`got-promise`](https://github.com/floatdrop/got-promise) - got with Promises
```
sindresorhus: > - Promise wrapper

sindresorhus:
Generally not a fan of promise wrappers, but sure, since it's you :)

floatdrop:
@sindresorhus i'm not quite sure about it either. Should we add all wrappers here then?

sindresorhus:
We can evaluate on a case-by-case basis.

sindresorhus:
Re `got-promise`, I think it would be nicer if you exposed the promise as the main export, and dropped the `.promise` property. If people want a callback API they can just use vanilla `got`.

floatdrop:
@sindresorhus yeah, I thought about it, but in this way it is easier to get, what `got` call returning without having `gotPromise` variable around. Also it is easier to migrate from callback to Promises with this drop-in replacement. But yes - not quite common way of doing promisification.

sindresorhus:
> what got call returning without having gotPromise variable around.

Not sure what you mean. Mind showing an example?

> Also it is easier to migrate from callback to Promises with this drop-in replacement.

If you mean gradual migration, they could just use both temporarily. Don't really see the benefit here either.

floatdrop:
> Not sure what you mean. Mind showing an example?

```js
var got = require('got-promise');

got.promsie() // => Promise
got() // => Stream

// --- vs ---

var got = require('got');
var gotPromise = require('got-promise');

gotPromise() // => Promise
got() // => Stream
```

One line is not a huge benefit :dog:

We have promise wrapper around `request` and it was easier to use request inside promise wrapper, than installing another instance of it in project. Now with `got` this is not a problem anymore, so you are right.

sindresorhus:
If you insist on having the old API available I would at least do:

```js
var got = require('got-promise');

got() // => Promise
got.raw() // => Stream
```

PR #51, opened 2015-04-24
Support Object opts[,Function cb] signature

golyshevd:
Now, `got` supports also (opts[, cb]) signature. Url is not required.

Example:

```js
got({host: 'example.com', path: '/example'}, function (err, data, res) {
   // stuff
});
```

Convenient thing to keep (hostname, port, etc.) in project config separately, and no need to format `url`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,11 +39,10 @@ fs.createReadStream('index.html').pipe(got.post('todomvc.com'));
 
 It's a `GET` request by default, but can be changed in `options`.
 
-#### got(url, [options], [callback])
+#### got([url][, options][, callback])
 
 ##### url
 
-*Required*  
 Type: `string`
 
 The URL to request.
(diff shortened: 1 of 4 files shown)


floatdrop:
Let start from simple - split this PR: one for `query`, one for object as first argument.

golyshevd:
Removed query feature commit

floatdrop:
Okay, could you squash commits, provide clear description of PR (what it does, why you need it)?

golyshevd:
Ok, I will patch readme tonight and then squash commits after review

readme.md
```diff
@@ -39,11 +39,10 @@ fs.createReadStream('index.html').pipe(got.post('todomvc.com'));
 
 It's a `GET` request by default, but can be changed in `options`.
 
-#### got(url, [options], [callback])
+#### got([url][, options][, callback])
```
floatdrop: So `got()` is valid now?

golyshevd: I did not thought about it earlier but in fact, yes, it will be like `got({})`. Need check for "at least one argument"?

floatdrop: > `got([url][, options][, callback])`

I don't think this is a good API.

floatdrop:
I like idea of passing `url.parse` object as first argument - this will simplify usecase of wrapping got in other module, that already parsed url string. But this should now take so much changes to implement.

golyshevd:
What problem that it is few additional lines of code?

Simplest way to do it is to convert (opts, cb) to ('', opts, cb) but it is ugly in my opinion, and not quite right

floatdrop:
@golyshevd problem is that it is not a _few_ lines. Less code to maintain and debug is always better imo.

golyshevd:
You are absolutely right. But "many code" does not mean "complex code". In my opinion `got` becomes to be complex because of `got` is too polymorphic. But refactoring solutions is not a topic.

Personally, I need this feature in `got` and I have a plans to use `got` in future. But I have no idea how to do it with less diff

PR #52, opened 2015-04-25
Accept http.request object as first argument

floatdrop:
Cont. of https://github.com/sindresorhus/got/pull/51

If I want to pass `url.parse` result to `got` from my own parse logic - I need to do this hack:

```js
got('', { host: 'yada.com' }, function (err, data) {
    console.log(data);
});
```

Which is not nice and `got` will try to parse `''`.

This will make `got` compatible with `http.request` method:

```js
got({ host: 'yada.com' }, function (err, data) {
    console.log(data);
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,9 +70,11 @@ function got(url, opts, cb) {
 	}
 
 	function get(url, opts, cb) {
-		var parsedUrl = urlLib.parse(prependHttp(url));
+		var parsedUrl = typeof url === 'string' ? urlLib.parse(prependHttp(url)) : url;
 		var fn = parsedUrl.protocol === 'https:' ? https : http;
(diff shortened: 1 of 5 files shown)


floatdrop:
@sindresorhus @kevva @julien-f thoughts?

floatdrop:
By the way, placing `host` in options will break readirects because of [this line](https://github.com/sindresorhus/got/blob/master/index.js#L75).

sindresorhus:
> By the way, placing host in options will break readirects because of this line.

Can you open an issue about that?

sindresorhus:
:+1: Makes sense.

floatdrop:
woot

> Can you open an issue about that?

This is not critical at all I think, opened PR with failing test - https://github.com/sindresorhus/got/pull/54

PR #55, opened 2015-04-26
Broken redirects for host/port etc... in options

floatdrop:
This is ~~quite dirty~~ fix for overriding http options on redirect.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,8 +25,8 @@ util.inherits(GotError, NestedErrorStacks);
 GotError.prototype.name = 'GotError';
 
 function got(url, opts, cb) {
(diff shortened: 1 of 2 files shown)


floatdrop:
Will break this use case:

```js
got('host.com', { host: 'newhost.com' }).pipe(process.stdout);
```

But not `query` option.

floatdrop:
@sindresorhus any hints on readme path with this?

sindresorhus:
> got('host.com', { host: 'newhost.com' }).pipe(process.stdout);

Can't we just throw on this? I see no sane reason why this should be supported. What even is the expected outcome if it were allowed?

PR #56, opened 2015-04-27
fix redirects for host/port etc... in options

floatdrop:
Accidentally pushed #55 in master.

### DIFF
--- a/index.js
+++ b/index.js
@@ -123,6 +123,11 @@ function got(url, opts, cb) {
 					return;
 				}
 
+				delete opts.host;
+				delete opts.hostname;
+				delete opts.port;
+				delete opts.path;
+
 				get(urlLib.resolve(url, res.headers.location), opts, cb);
 				return;
 			}


sindresorhus:
> got('host.com', { host: 'newhost.com' }).pipe(process.stdout);

Can't we just throw on this? I see no sane reason why this should be supported. What even is the expected outcome if it were allowed?

floatdrop:
> Can't we just throw on this? I see no sane reason why this should be supported. What even is the expected outcome if it were allowed?

@sindresorhus I guess we can. Also we can:

 * drop this feature and don't do `objectAssign` at all. All 'bad' properties would be just ignored.
 * drop those properties on redirect (this will allow `got(conf.host, {path: '/mypath'})`)

As you say - `there is no sane reason to do so`, so probably there is no sane way to deal with it.

sindresorhus:
Sounds good :+1:

floatdrop:
@sindresorhus which one?

sindresorhus:
2

PR #57, opened 2015-04-28
auto-redirect only on GET and HEAD methods

floatdrop:
As [rfc2616 HTTP 1.1](http://tools.ietf.org/html/rfc2616#section-10.3.3) says:

```
If the 302 status code is received in response to a request other
than GET or HEAD, the user agent MUST NOT automatically redirect the
request unless it can be confirmed by the user, since this might
change the conditions under which the request was issued.
```

I think we should return error on POST redirect, because (in case of body as a Stream) we can not POST again same content and buffering it is not an option.

There are `302` and `303` codes, that require user approval - in case of HTTP client it should be an error I guess.

-

Interesting discussion about this in https://github.com/request/request/issues/29

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,6 +59,8 @@ function got(url, opts, cb) {
 		opts.method = opts.method || 'POST';
 	}
 
+	opts.method = opts.method || 'GET';
+
 	// returns a proxy stream to the response
 	// if no callback has been provided
 	if (!cb) {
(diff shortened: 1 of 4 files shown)


kevva:
:+1: for me. If we were to allow redirects for other methods we'd do that through an option I guess.

floatdrop:

sindresorhus:
:+1:

Issue #58, opened 2015-04-30
Preserve cookies on redirect


floatdrop:
While thinking about redirects, I notice, that `got` is ignoring cookies on redirects - and we have no way to do something about it in userland. 

~~We could resend `set-cookie` header as `cookie` on redirect - all gathered cookies on redirects path could be merged in `set-cookie` header in response.~~ (Cookies are host-sensitive and redirect can change host.)

sindresorhus:
I have no idea how this is supposed to work tbh. What does the spec say? What does `request` do?

floatdrop:
@sindresorhus [needle ignoring cookies on redirect by default](https://www.npmjs.com/package/needle#redirect-options) and request will [store them in global cookie jar](https://github.com/request/request/blob/master/request.js#L1099-L1106) which is not quite good idea (how can I clear it?).

Most browsers will set cookies on `30x` response, some sites are using this for authentication.

I have found only one thing about redirects and cookies in [RFC 2109](https://www.ietf.org/rfc/rfc2109.txt):

>   4.3.5  Sending Cookies in Unverifiable Transactions

>   Unverifiable transactions typically
   arise when a user agent automatically requests inlined or embedded
   entities or when it resolves redirection (3xx) responses from an
   origin server.

>   When it makes an unverifiable transaction, a user agent must enable a
   session only if a cookie with a domain attribute D was sent or
   received in its origin transaction, such that the host name in the
   Request-URI of the unverifiable transaction domain-matches D.

So we should set cookies, if it is safe to set it. This will be not easy feature to implement, I guess.

floatdrop:
In fact we have two problems:

1. Getting cookies from redirect response
2. Setting new cookies to redirect request

While first is easly solved with new `redirect` event:

```js
got('cookie.com')
    .on('redirect', function (res) {
        console.log(res.headers);
    });
```

Second is not quite straightforward. May be in this event we could pass headers as second argument:

```js
got('cookie.com')
    .on('redirect', function (res, next) {
        next.headers.cookie = cookie(res);
    });
```

In this case we will be able to move stuff about cookies (which I like to avoid) to userland.

@sindresorhus @kevva wdyt?

posabsolute:
I know this is old, but this still an issue? Can we pass cookies when using promises?

AlbertoFdzM:
I recently tried this.

I'm using a 3rd party service that uses 302 redirects with "Set-Cookie" headers to establish a session.

To clarify how it works:
[...]

PR #59, opened 2015-05-05
Implement redirect event

floatdrop:
Solution for #58 (for Stream mode only - but it is better than nothing).


kevva:
:+1:

sindresorhus:
:+1:

PR #65, opened 2015-05-10
Use node-tap for tests

floatdrop:
Switch for sweet reporter and errors diff:

screen shot 2015-05-10 at 11 48 11 am


sindresorhus:
Hot

PR #68, opened 2015-05-24
add `then-got` to related

tunnckoCore:
package since december

[see the usage](https://github.com/hybridables/then-got#usage) - promise and callback api in same time


sindresorhus:
Thanks, but we're making got promisified by default, so none of these should be needed for much longer.

PR #76, opened 2015-06-24
Add content-length header in POST

### LINKED ISSUE #71: Set Content-Length when body is a string or a buffer
If not already defined by the user of course.

floatdrop:
Set `content-length` header if:
 - not already defined
 - 'transfer-encoding' is not used
 - body is not Stream

Closes #71

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,6 +62,11 @@ function got(url, opts, cb) {
 
 	if (body) {
 		opts.method = opts.method || 'POST';
+
+		if (!opts.headers['content-length'] && !opts.headers['transfer-encoding'] && !isStream.readable(body)) {
+			var length = typeof body === 'string' ? Buffer.byteLength(body) : body.length;
+			opts.headers['content-length'] = length;
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -61,7 +61,16 @@ function got(url, opts, cb) {
...
 	if (body) {
+		if (typeof body !== 'string' && !Buffer.isBuffer(body) && !isStream.readable(body)) {
+			throw new GotError('options.body must be a ReadableStream, string or Buffer');
+		}
+
 		opts.method = opts.method || 'POST';
+
+		if (!opts.headers['transfer-encoding'] && !isStream.readable(body)) {
+			var length = typeof body === 'string' ? Buffer.byteLength(body) : body.length;
+			opts.headers['content-length'] = opts.headers['content-length'] || length;
```
julien-f: You could test the presence of `content-length` directly in the `if` above instead.

floatdrop: :+1:

julien-f:
LGTM :)

Issue #72, opened 2015-06-19
url.path vs url.pathname


julien-f:
It seems that `http(s).request(url)` uses `url.path` while `urlLib.format(url)` uses `url.pathname`.

I do not have the time to look further but I think it can cause some issues (for instance when displaying the URL in an error message).

floatdrop:
@julien-f nice catch. `url.format` is not quite good at formatting I think:

```js
url.format({protocol: 'http:', host: 'localhost', port: 6767});
// => http://localhost without port
```

So in error messages (that was generated from got with Object as first argument) we don't have port now. :(

May be it worth trying to replace `url.format` with some module?

julien-f:
It's because `host` is `${hostname}:`${port}`, therefore it takes precedence.

sindresorhus:
> It seems that http(s).request(url) uses url.path while urlLib.format(url) uses url.pathname.

Should we maybe open an issue on iojs? Seems like that should be consistent.

PR #77, opened 2015-06-25
Code cleanup and pathname issue

floatdrop:
There were some :hankey: code (by me) like this:

```js
arg.path = (arg.path ? arg.path.split('?')[0] : '') + '?' + (typeof query === 'string' ? query : querystring.stringify(query));
```

Or this:

```js
var parsedUrl = typeof url === 'string' ? urlLib.parse(prependHttp(url)) : url;
// ...
url = typeof url === 'string' ? prependHttp(url) : urlLib.format(url);
```

Which was hard to maintain and read. 

Main reason for this code - there was two different arguments for `get` function - first contains request options (as object __or__ as string) and other gets got options.

This logic was replaced by merging `url` and `opts` arguments in `got` function into one. Because (on redirect) we can reassign parsed request options.

In the process `redirect` event was fixed - it was emitting old `opts` that was not from redirect itself.

-

#72 was fixed by reassing  `path` to `pathname` (if `pathname` is present). I don't know, if this is right thing to do, but because of difference in `http.request` and `url.format` +  `url.parse` we have to normalise options or use another `url.format` implementation.

//cc @sindresorhus @sindresorhus @kevva

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,6 +48,10 @@ function got(url, opts, cb) {
 		'accept-encoding': 'gzip,deflate'
 	}, lowercaseKeys(opts.headers));
 
+	if (opts.pathname) {
+		opts.path = opts.pathname;
+	}
+
 	if (opts.query) {
 		if (typeof opts.query !== 'string') {
 			opts.query = querystring.stringify(opts.query);
--- a/index.js
+++ b/index.js
@@ -35,18 +35,32 @@ function got(url, opts, cb) {
 		opts = {};
 	}
 
-	opts = objectAssign({}, opts);
+	opts = objectAssign(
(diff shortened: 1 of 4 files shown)


floatdrop:
Side-effect: now you can specify `got` options in first argument:

```js
got({host: 'google.com', timeout: 2000});
```

sindresorhus:
> Side-effect: now you can specify got options in first argument:

Should this be documented?

sindresorhus:
Super nice cleanup and improvements @floatdrop :D

sindresorhus:
tumblr_ml9za9ecsb1qdlh1io1_400

PR #78, opened 2015-06-26
Handle empty response (204)

connyay:


floatdrop:
Technically - blank response (at least with `200` status code) is invalid response.

http://stackoverflow.com/questions/11970962/valid-json-in-response

connyay:
The situation I hit this with was on a 204 response. I started this PR only
checking if the status code was 204, but figured an empty 2xx would be
possible as well.

Thoughts?

floatdrop:
@connyay I think it is expected behaviour never less empty string is invalid JSON. [request](github.com/request/request) is behaving something like this.

sindresorhus:
I'm not sure tbh. Empty response on 200 is invalid, so we shouldn't do anything there. 204 is valid with blank response, so we should maybe handle that somehow (suggestions?). Maybe we should only try to parse JSON when content-type is "application/json"? I assume servers don't send that on 204?.

floatdrop:
@sindresorhus a lot of endpoints do not send `application/json` content-type in json responses. This was main point of strict `json` option - to parse body in any case.

@connyay could you do a 204 check instead?

Related: https://github.com/visionmedia/superagent/issues/255

connyay:
@floatdrop switched to check for 204

floatdrop:
@connyay cool! Can you squash it into one commit?

connyay:
@floatdrop should be good. 

:beers:

floatdrop:
:beers:

Issue #66, opened 2015-05-13
Promise API


floatdrop:
This is just a proposal - feel free to share ideas, criticize, etc.

#### Idea

Get Promise API into `got` with polyfill for Promise in Node.JS `0.10` like this:

```js
got('google.com'); // => Promise

got('google.com', function (err, body) {
    // ...
});
```

#### Motivation
 
:sparkles:Generators:sparkles: are awesome for flow-control and we finally can forget about callback hell and move on (with some help of [`co`](http://npmjs.com/co)-like libraries) to `yield`:

```js
co(function *() {
    var results = yield [got('google.com'), got('github.com')];
    console.log(results);
});
```

To support this, we should return Promise from got.

Also new [`fetch API`](https://hacks.mozilla.org/2015/03/this-api-is-so-fetching/) based on Promises landed in Chrome 42.

When ES7 will come to play, we will be able to go even further with `async`/`await`:

```js
async function () {
    let site = await got('google.com');
};
```

#### Problems

> Promise API conflicts with Stream API

Atm I have no idea how to merge them (and if it is possible, I would vote against it), so one way is to move Stream API under `stream` property, like this:

```js
var stream = got.stream('google.com');
```

This will be breaking change, so "brace yourselves, major release is coming" (hah). What we can do in `3.x` version, to warn users - add `promise` property first and deprecation warning of Stream API.

#### Resolve format

There are two ways, with which Promise can resolve request. This will affect, how results can be destructed into variables:

Resolve into Array of `response` and `body` (like [then-got](http://npmjs.com/then-got)):

```js
// Assume co-wrapper in all examples below ↓

var [response, body] = yield got('google.com');
```

Resolve into response Object with `body` property (like [got-promise](http://npmjs.com/got-promise)):

```js
var {body, headers} = yield got('google.com');
```

It seems like resolving to Object make much more sens - order of vars is not matters and user can access body, without response (or vice versa with different order).

-

TS: @sindresorhus
Related: https://github.com/request/request/issues/796

sindresorhus:
Agree with everything :star2: :+1: 

> It seems like resolving to Object make much more sens - order of vars is not matters and user can access body, without response (or vice versa with different order).

Yes, I would prefer an object too.

-

@kevva @julien-f @silverwind @arthurvr @shinnn

arthurvr:
I'm all for this! :+1:

> It seems like resolving to Object make much more sens - order of vars is not matters and user can access body, without response (or vice versa with different order).

[...]

PR #82, opened 2015-07-15
4.0.0

### LINKED ISSUE #79: Proxy support
_Previous issue https://github.com/sindresorhus/got/issues/7. I couldn't care less about this, but it's useful to have a place to refer people that open issues elsewhere._
## 

A good PR would be accepted, but won't happen from my part.

I would prefer for the solution to be an external module `got` can just depend on as I don't want to have to maintain it. Maybe https://github.com/mikeal/tunnel-agent.
## 

I strongly believe this is something that should be a part of Node.js and not every module doing HTTP. `+1` this issue if you agree: https://github.com/nodejs/node/issues/8381

floatdrop:
:dragon_face: Here be dragons!

### Promise API

As discussed in #66. [`pinkie-promise`](http://npmjs.com/pinkie-promise) was taken as fallback for native Promise.

### infinity-agent removed

Main reason to use `infinity-agent` was to fix internal `maxSockets` property on 0.10 version of NodeJS. But this approach is not quite configurable and not quite composable with custom agent, which could be necessary to fix #79 

So `infinity-agent` should be dropped in favore to readme section with workaround:

```js
require('http').globalAgent.maxSockets = Infinity;
require('https').globalAgent.maxSockets = Infinity;
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,10 +77,10 @@ function requestAsEventEmitter(opts) {
 			timedOut(req, opts.timeout);
 		}
 
-		ee.emit('request', req);
+		setImmediate(ee.emit.bind(ee), 'request', req);
 	}
 
-	setImmediate(get, opts); // quirk to attach event listeners
+	get(opts);
 	return ee;
 }
 
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 'use strict';
+var EventEmitter = require('events').EventEmitter;
 var http = require('http');
 var https = require('https');
 var urlLib = require('url');
@@ -14,6 +15,7 @@ var prependHttp = require('prepend-http');
 var lowercaseKeys = require('lowercase-keys');
 var isRedirect = require('is-redirect');
 var NestedErrorStacks = require('nested-error-stacks');
+var pinkiePromise = require('pinkie-promise');
 
 function GotError(message, nested) {
 	NestedErrorStacks.call(this, message, nested);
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -24,221 +25,157 @@ function GotError(message, nested) {
...
+				var unzip = zlib.createUnzip();
+				unzip.httpVersion = res.httpVersion;
+				unzip.headers = res.headers;
+				unzip.rawHeaders = res.rawHeaders;
+				unzip.trailers = res.trailers;
+				unzip.rawTrailers = res.rawTrailers;
+				unzip.setTimeout = res.setTimeout.bind(res);
+				unzip.statusCode = res.statusCode;
+				unzip.statusMessage = res.statusMessage;
+				unzip.socket = res.socket;
```
sindresorhus: Use `objectAssign` here?

index.js
```diff
@@ -24,221 +25,157 @@ function GotError(message, nested) {
...
-				body.pipe(req);
-			} else {
-				req.end(body);
-			}
+function asPromise(opts) {
+	var resolve, reject;
+
+	var promise = new pinkiePromise(function (_resolve, _reject) {
+		resolve = _resolve;
+		reject = _reject;
```
sindresorhus: Why?

index.js
```diff
@@ -248,13 +185,88 @@ function got(url, opts, cb) {
...
+function got(url, opts, cb) {
+	if (typeof opts === 'function') {
+		cb = opts;
+		opts = {};
+	}
+
+	opts = normalizeArguments(url, opts);
+
+	if (cb) {
+		return asCallback(opts, cb);
```
```js
asCallback(opts, cb);
return;
```

To make it clear it doesn't return the function as value.

readme.md
```diff
@@ -185,6 +186,18 @@ got('todomvc.com', {
 ```
 
 
+## Node 0.10
+
+It is a known issue with Node [http.Agent](https://nodejs.org/docs/v0.10.39/api/http.html#http_class_http_agent) and `agent.maxSockets`, which is set to `5`. This can cause low performance of application and (in rare cases) deadlocks. To avoid this you can set it manually:
```
sindresorhus: Make it clear it's fixed in Node 0.12 and higher, so it's only needed to do this if you care about Node 0.10 compatibility in your application.

sindresorhus:
Yay! \o/

-

Gave it a quick shim and generally looks good. Will give it a proper review tomorrow.

-

// @julien-f @kevva @shinnn @arthurvr

readme.md
floatdrop: Fixed a little bit - but still can be wrong/confusing.

p.s. This agent tuning does not affects compatibility (it will work on 0.10) - but, in some cases, performance.

index.js
floatdrop: I thought it was nice to read - return as callback, but to keep things consistent :+1:

index.js
floatdrop: To pass them to callback, I guess... :dog: (fixed)

index.js
floatdrop: It will copy internal fields - which breaks unzip stream. Maybe turn it to a module?

floatdrop:
@sindresorhus friendly ping :)

sindresorhus:
Sorry about the delay. I'm traveling in Iceland at the moment. Will be back on Friday.

index.js
```diff
@@ -273,4 +262,27 @@ function got(url, opts, cb) {
...
+});
+
+got.RequestError = createErrorClass('RequestError');
+got.ReadError = createErrorClass('ReadError');
+got.ParseError = createErrorClass('ParseError');
+
+got.HTTPError = createErrorClass('HTTPError', function() {
+	this.message = 'Response code ' + this.statusCode + ' (' + http.STATUS_CODES[this.statusCode] + ')';
+	this.statusMessage = http.STATUS_CODES[this.statusCode];
+});
```
sindresorhus: Can be simplified:

Also the space after `function`

```js
got.HTTPError = createErrorClass('HTTPError', function () {
  this.message = 'Response code ' + this.statusCode + ' (' + this.statusMessage + ')';
});
```

index.js
```diff
@@ -14,255 +12,246 @@ var timedOut = require('timed-out');
...
-			var statusCode = response.statusCode;
-			var res = response;
+			var statusCode = res.statusCode;
 
-			// auto-redirect only for GET and HEAD methods
-			if (isRedirect(statusCode) && 'location' in res.headers && (opts.method === 'GET' || opts.method === 'HEAD')) {
-				// discard response
-				res.resume();
+			if (statusCode < 200 || statusCode > 299) {
+				err = new got.HTTPError({statusCode: statusCode, url: url, method: opts.method});
```
sindresorhus: You pass in `url` and `method` here, but I don't see it used in the error construction.

index.js
```diff
@@ -14,255 +12,246 @@ var timedOut = require('timed-out');
...
-	delete opts.encoding;
-	delete opts.body;
-	delete opts.json;
-	delete opts.timeout;
+			res = unzipResponse(res);
 
-	if (json) {
-		opts.headers.accept = opts.headers.accept || 'application/json';
-	}
+			ee.emit('response', res);
```
```js
ee.emit('response', unzipResponse(res));
```

?

index.js
```diff
@@ -14,255 +12,246 @@ var timedOut = require('timed-out');
...
+	var redirectCount = 0;
 
-	opts = objectAssign(
-		{
-			protocol: 'http:'
-		},
-		typeof url === 'string' ? urlLib.parse(prependHttp(url)) : url,
-		opts
-	);
+	function get(opts) {
```
sindresorhus: We should probably not nest function statements. Make it an expression instead:

```js
var get = function (opts) {
```

index.js
```diff
@@ -14,255 +12,246 @@ var timedOut = require('timed-out');
...
 
-	if (json) {
-		opts.headers.accept = opts.headers.accept || 'application/json';
-	}
+			ee.emit('response', res);
+		}).once('error', function (err) {
+			ee.emit('error', new got.RequestError(err.message, {
+				url: url,
+				method: opts.method,
+				code: err.code
```
sindresorhus: `MaxRedirectsError` uses `statusCode`, while this uses `code`. We should make it consistent.

readme.md
```diff
@@ -151,6 +152,17 @@ The [response object](http://nodejs.org/api/http.html#http_http_incomingmessage)
...
+
+Got provides rich set of Errors to simplify error handling:
+
+#### got.RequestError
+#### got.ReadError
+#### got.ParseError
+#### got.HTTPError
+#### got.MaxRedirectsError
+
+Each class have `url` and `method` property.
```
sindresorhus: `ReadError` doesn't have a `method` property.

readme.md
```diff
@@ -151,6 +152,17 @@ The [response object](http://nodejs.org/api/http.html#http_http_incomingmessage)
...
 
+## Errors
+
+Got provides rich set of Errors to simplify error handling:
+
+#### got.RequestError
+#### got.ReadError
+#### got.ParseError
+#### got.HTTPError
+#### got.MaxRedirectsError
```
sindresorhus: We should probably document if any of these have extra properties.

sindresorhus:
Ok, found some time on the airport to review. Expect for some super minor nitpick this looks really good! I like how the code is a lot more readable now. Great work :)

Feel free to merge when the inline comments are resolved.

index.js
floatdrop: :+1:

index.js
floatdrop: :+1:

index.js
floatdrop: They assigned to Error object as properties.

index.js
floatdrop: Okay, but what's the difference?

floatdrop:
I feel like some more time should be spent on Errors (because of #86, #72 and https://github.com/floatdrop/create-error-class/issues/1), so new Errors will move to another PR.

@sindresorhus thanks for kind words :sparkles: I really miss 80-lines version of got, so tried to simplify things as much as I can.

index.js
floatdrop: @sindresorhus about this - `code` contains class of Error in Node (like `ETIMEDOUT` or such) and in other places status of response/request stored in `statusCode` property (https://github.com/trentm/node-bunyan#log-record-fields, https://github.com/jshttp/http-errors and so on). Let's discuss it in #83.

arthurvr:
Super cool work, @floatdrop! :star:

sindresorhus:
@floatdrop Anything else we want to get into 4.0.0? Breaking changes would be nice to get in now so 5.0.0 can be far in the future.

floatdrop:
@sindresorhus nothing, except new Errors.

Issue #84, opened 2015-07-18
Swallows errors inside of callback


ArtskydJ:
It seems like `got` silently swallows errors.
```js
var got = require('got')

got('http://josephdykstra.com', function (err, data) {
	console.log('i see this message')
	throw new Error('lolwut')
	console.log('but not this message')
})

process.on('uncaughtException', function(err) {
	console.log('error: ' + err)
})
```
```bash
C:\Users\Michael\Github\javascript\just-login-emailer\test>node test2.js
i see this message

C:\Users\Michael\Github\javascript\just-login-emailer\test>
```

floatdrop:
Thanks! This bug from `read-all-stream` and fixed in `3.0.1` version. Update please.

shinnn:
:+1:

PR #85, opened 2015-07-18
Added failing test for #84

ArtskydJ:
Added failing test for swallowing errors inside the callback.

### DIFF
--- a/test/test-error.js
+++ b/test/test-error.js
@@ -40,6 +40,23 @@ test('options.body error message', function (t) {
 	t.end();
 });
 
+test('throwing errors inside callback', function (t) {
+	got(s.url, function (err) {
+		t.pass('Callback was called');
+		throw new Error('wut');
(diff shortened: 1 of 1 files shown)


ArtskydJ:
@floatdrop @sindresorhus Why not merge the test to avoid regressions?

If you don't like the way it's written, feel free to rewrite it, but I think it's good to have regression tests.

And if not here, maybe in `read-all-stream`.

floatdrop:
@ArtskydJ you perfectly right, test was added in `read-all-stream` - https://github.com/floatdrop/read-all-stream/commit/e3e4e904a55888d9f51b8f627cfc1850f087b1c4#diff-1dd241c4cd3fd1dd89c570cee98b79ddR90

ArtskydJ:
@floatdrop Whoops, I should've seen that. Thanks a lot!

PR #88, opened 2015-07-24
Explicit Error classes

floatdrop:
Replacement for `nested-error-stacks`.

* All error messages now retained and additional properties added (like `host` and `method`).
* `code` property now comes from internal errors of Node (with `ECONNREFUSED` and so on values) and HTTP status code stored in `statusCode` property.
* `pathname` support dropped (since there is no such option in `http.request`).
* `statusMessage` added.


floatdrop:
@sindresorhus @julien-f please review.

PR #92, opened 2015-08-13
Return HTTPError if response body failed to parse as JSON

eush77:
Currently

```js
got('http://registry.npmjs.com/nonexistent', { json: true }, function (err) {
  console.error(err);
})
```

results in this rather cryptic error message:

```
{ [ParseError: Unexpected end of input]
  message: 'Unexpected end of input',
  code: undefined,
  host: 'registry.npmjs.org',
  hostname: 'registry.npmjs.org',
  method: 'GET',
  path: '/nonexistent' }
```

With this patch, `HTTPError` is returned instead if response is not a valid JSON.

```js
{ [HTTPError: Response code 404 (Not Found)]
  message: 'Response code 404 (Not Found)',
  code: undefined,
  host: 'registry.npmjs.com',
  hostname: 'registry.npmjs.com',
  method: 'GET',
  path: '/nonexistent',
  statusCode: 404,
  statusMessage: 'Not Found' }
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,7 @@ function asCallback(opts, cb) {
 				try {
 					data = JSON.parse(data);
 				} catch (e) {
-					err = new got.ParseError(e, opts);
+					err = err || new got.ParseError(e, opts);
 				}
 			}
 
--- a/test/test-json.js
+++ b/test/test-json.js
@@ -69,7 +69,8 @@ test('json option should catch errors on invalid non-200 responses', function (t
 	got(s.url + '/non200-invalid', {json: true}, function (err, json) {
 		t.ok(err);
 		t.deepEqual(json, 'Internal error');
-		t.equal(err.message, 'Unexpected token I');
+		t.equal(err.statusCode, 500);
+		t.equal(err.message, 'Response code 500 (Internal Server Error)');
 		t.equal(err.path, '/non200-invalid');
 		t.end();
 	});


floatdrop:
Yeah, errors at point when you have two sequential errors (`HTTPError` and then `ParseError`) are hard to reason about, but I still think that it should be `ParseError`, but maybe with additional fields about status code?

Reason behind it - some API (like github for example) will send with 404 a JSON payload, which you should parse. If JSON payload is broken - this is quite different error.

eush77:
@floatdrop Your reasoning makes sense to me.

I can suggest two ways of solving this problem:

1. Consulting Content-Type header for non-200 codes. Response is parsed as JSON if and only if its value is `application/json`. `json` option is simply ignored.

2. Additional option controlling whether response with non-200 codes should be JSON-decoded. Then request to GitHub API would have `{ json: true, jsonError: true }` and request to NPM API would have `{ json: true, jsonError: false }`. `{ json: true }` would be a shortcut for one or the other, whichever is more common in the real world.

sindresorhus:
An option for this is not going to happen.

floatdrop:
How about add `status` and `statusCode` to `PraseError`?

```js
{ [ParseError: Unexpected end of input]
  message: 'Unexpected end of input',
  code: undefined,
  host: 'registry.npmjs.org',
  hostname: 'registry.npmjs.org',
  method: 'GET',
  path: '/nonexistent'
  statusCode: 404,
  statusMessage: 'Not Found' }
```

julien-f:
By experience, the error should also contain the raw response.

julien-f:
I mean, content, but why not the response object too.

See [xmlrpc](/baalexander/node-xmlrpc/#to-debug-client-side) for inspiration.

floatdrop:
@julien-f you get body in `response.body` or second argument in callback mode, it does not disappear on error.

```js
got('google.com/broken', {json: true}, function (err, body, res) {
    // err is ParseError
    // body is String
    // res is Response
});
```

User can easily add body to error in the callback.

julien-f:
My bad, it is enough as it is :)

That leads me to another question, how can we promisify correctly such an API which has both an error and a result?

floatdrop:
@julien-f just like you said to do, add response to error - https://github.com/sindresorhus/got/blob/master/index.js#L114

eush77:
> How about add `status` and `statusCode` to `PraseError`?

This is in fact combining `HTTPError` and `ParseError` into one type. How about combine error messages as well? This way `err.toString()` would be more useful.

```js
{ [ParseError: Unexpected end of input (code 404)]
  message: 'Unexpected end of input (code 404)',
  code: undefined,
  host: 'registry.npmjs.org',
  hostname: 'registry.npmjs.org',
  method: 'GET',
  path: '/nonexistent'
  statusCode: 404,
  statusMessage: 'Not Found' }
```

Or `HTTPError: 404 Not Found (unexpected end of input)`.

floatdrop:
@eush77 I don't like to concatenate (or modify) error messages - it can become messy in long time run.

PR #112, opened 2015-10-14
Emit either error or reponse event in stream mode

### LINKED ISSUE #110: In stream mode emits response with empty body
(In stream mode) on error event `got` emits `error` and then `response` events (in this order). This seems kind of awkward - because `error` will read response body and this will lead to confusion in `request` event handler.

@sindresorhus thoughts?

floatdrop:
Closes #110


floatdrop:
Response event in stream mode will be overridden by error. I think this is most simple thing to do.

//cc @kevva

PR #114, opened 2015-10-14
Throw on authentication in url string

### LINKED ISSUE #106: Prevent usage of Basic Auth in URL
From: https://github.com/sindresorhus/got/issues/105#issuecomment-141696906

Passing Basic Auth in the URL is deprecated and not even possible anymore in Chrome and IE: http://serverfault.com/questions/371907/can-you-pass-user-pass-for-http-basic-authentication-in-url-parameters https://tools.ietf.org/html/rfc3986#section-3.2.1

In the next major version of `got` we should throw and tell the user to use the `auth` option instead.

floatdrop:
Closes #106

### DIFF
--- a/index.js
+++ b/index.js
@@ -182,9 +182,17 @@ function normalizeArguments(url, opts) {
 		throw new Error('Parameter `url` must be a string or object, not ' + typeof url);
 	}
 
+	if (typeof url === 'string') {
+		url = urlLib.parse(prependHttp(url));
+
+		if (url.auth) {
+			throw new Error('Basic authentication must be done with auth option');
+		}
+	}
+
 	opts = objectAssign(
 		{protocol: 'http:', path: ''},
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -182,9 +182,17 @@ function normalizeArguments(url, opts) {
 		throw new Error('Parameter `url` must be a string or object, not ' + typeof url);
 	}
 
+	if (typeof url === 'string') {
+		url = urlLib.parse(prependHttp(url));
+
+		if (url.auth) {
+			throw new Error('Basic authentication must be done with Authorization header');
```
sindresorhus: Why not just tell them to use the `http.request` `auth` option?

> Basic authentication must be done with the `auth` option

floatdrop: Dunno. `auth` option makes more sense.

PR #117, opened 2015-10-15
Deferred response event

floatdrop:
On request-heavy applications (with short `timeout` option) lots of requests will be rejected, because of `response` handlers.

For example, if `timeout` is `50ms` and `response` from `http` module was on `48ms` - chances are, that user will get `ETIMEDOUT` error.


sindresorhus:
:+1:

PR #118, opened 2015-10-16
Rewrite tests with async/await

### LINKED ISSUE #115: Convert tests to use the promise interface by default
Except for a few tests that specifically tests the callback interface.

We can then use async functions, which are so nice:

```js
test('retry - timeout errors', async t => {
	t.is(await got(`${s.url}/knock-twice`, {timeout: 1000}), 'who`s there?');
});
```

More info: https://github.com/sindresorhus/ava#async-function-support

floatdrop:
Closes #115 

P.s. I try'd to convert `_server.js` to provide `Promise` API, but failed. Maybe next time :)

### DIFF
--- a/test/arguments.js
+++ b/test/arguments.js
@@ -21,42 +21,30 @@ test.before('arguments - setup', t => {
 	s.listen(s.port, () => t.end());
 });
 
-test('arguments - url argument is required', t => {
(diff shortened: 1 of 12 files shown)


kevva:
:+1:

PR #123, opened 2015-10-30
Add the ability to explicitly omit certain headers.

julien-f:
For instance, `transfer-encoding` is automatically set by Node if `content-length` is not present.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,14 +60,6 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
-		// this method is not documented, prints a warning if used while
-		// not available
-		if (!req.removeHeader) {
-			req.removeHeader = function (headerName) {
-				console.error('could not remove', headerName, 'header');
-			};
-		}
-
 		// provide the ability to explicitly not send a header by setting
 		// it to `null`
 		var headerName;
--- a/index.js
+++ b/index.js
@@ -60,6 +60,23 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
+		// this method is not documented, prints a warning if used while
+		// not available
+		if (!req.removeHeader) {
+			req.removeHeader = function (headerName) {
+				console.error('could not remove', headerName, 'header');
+			};
+		}
+
+		// provide the ability to explicitly not send a header by setting
+		// it to `null`
+		var headerName;
+		for (headerName in opts.headers) {
+			if (opts.headers[headerName] === null) {
+				req.removeHeader(headerName);
+			}
+		}
+
 		req.once('error', function (err) {
 			if (retryCount < opts.retries) {
 				setTimeout(get, backoff(retryCount++), opts);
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -74,6 +74,14 @@ Type: `object`
 
 Any of the [`http.request`](http://nodejs.org/api/http.html#http_http_request_options_callback) options.
 
+###### headers
+
+Type: `object`
+
+Headers that will be sent wit a request.
```
sindresorhus: `with`

readme.md
```diff
@@ -74,6 +74,14 @@ Type: `object`
...
+
+Any header can be set to `null` to explicitly omit them, even if they are implicit headers that would be added by Node (i.e. `tranfer-encoding` if `content-length` is not present).
```
sindresorhus: `Node` => `Node.js`

index.js
```diff
@@ -60,6 +60,23 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
+		// this method is not documented, prints a warning if used while
```
sindresorhus: Looks documented to me: https://nodejs.org/api/http.html#http_response_removeheader_name

Can you find out in which Node.js version it was introduced?

index.js
```diff
@@ -245,11 +262,13 @@ function normalizeArguments(url, opts) {
...
 		if (isPlainObj(body)) {
-			opts.headers['content-type'] = opts.headers['content-type'] || 'application/x-www-form-urlencoded';
+			if (opts.headers['content-type'] === undefined) {
+				opts.headers['content-type'] = 'application/x-www-form-urlencoded';
+			}
 			body = opts.body = querystring.stringify(body);
 		}
 
-		if (!opts.headers['content-length'] && !opts.headers['transfer-encoding'] && !isStream.readable(body)) {
+		if (opts.headers['content-length'] === undefined && !opts.headers['transfer-encoding'] && !isStream.readable(body)) {
```
floatdrop: Should `!opts.headers['transfer-encoding']` be `opts.headers['transfer-encoding'] === undefined` then too?

julien-f: Nope, I put `=== undefined` because I want the user to be able to prevent a specific head (here `content-length`) to be defined.

sindresorhus:
Can you provide a realistic example of when someone would need this?

index.js
floatdrop: [Node.js 0.8.29](https://nodejs.org/docs/v0.8.28/api/http.html#http_response_removeheader_name) have this method - so it is safe to use.

sindresorhus: Alright, then we can just remove this check and warning.

julien-f: It is documented for `ServerResponse`, not for `ClientRequest` even if they both inherit from the same internal class`OutgoingMessage`.

julien-f:
I need it to send a request with a stream body I don't know the length of, without using chunked encoding.

Maybe there is a better to do it but I did not find it…

julien-f:
In fact it's a very bad use case because it does not work at all for what I wanted :p

Might still be useful for other use cases, I dunno.

sindresorhus:
Then I'm :-1:. @floatdrop ?

floatdrop:
We don't stumble on such problems yet. If it is not fixing any use case, I'd rather wait on it.

floatdrop:
@julien-f anyway thanks for pointing thing about `undefined` - it can be source of bugs.

PR #125, opened 2015-11-01
Make response property non enumerable on error

floatdrop:
Often in debugging I find rather difficult to read this output:

```js
got('google.com/wat').catch(err => console.log(err));

/*
{ [HTTPError: Response code 404 (Not Found)]
  message: 'Response code 404 (Not Found)',
  code: undefined,
  host: 'google.com',
  hostname: 'google.com',
  method: 'GET',
  path: '/wat',
  statusCode: 404,
  statusMessage: 'Not Found',
  response:
   IncomingMessage {
      <here goes huge amount of properties, like couple of screens>
   } }
*/
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,7 +130,10 @@ function asPromise(opts) {
 			}
 
 			if (err) {
-				err.response = response;
+				Object.defineProperty(err, 'response', {
+					value: response,
+					enumerable: false
+				});
 				reject(err);
 				return;
 			}
--- a/test/error.js
+++ b/test/error.js
@@ -15,12 +15,14 @@ test.before('setup', async t => {
 	await s.listen(s.port);
 });
 
-test('message', async t => {
+test('properties', async t => {
 	try {
 		await got(s.url);
 		t.fail('Exception was not thrown');
 	} catch (err) {
 		t.ok(err);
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -130,7 +130,10 @@ function asPromise(opts) {
 			}
 
 			if (err) {
-				err.response = response;
+				Object.defineProperty(err, 'response', {
+					value: response,
+					enumerable: false
```
sindresorhus: `enumerable` is false by default

sindresorhus:
:+1:

sindresorhus:
> code: undefined,

Is there any point in including the `code` prop when it's undefined?

index.js
floatdrop: Didn't know that, but I feel, that this way it is clearer why this code is here, compared with `Object.defineProperty(err, 'response', {value: response});` - someone definatly will PR `err.response = response;` :smile:

sindresorhus: Sure. Up to you. It's clear for me since I know that all the configs are false by default.

floatdrop:
> Is there any point in including the code prop when it's undefined?

Nope, this bugs me as well. Will do in separate PR.

PR #139, opened 2015-12-01
Accept backoff function in retry option

### LINKED ISSUE #138: Accept backoff function in `retries` option
From #136:

Now `retry` option accepts only number of retries. Idea is to pass function to `retries`, which will return delay of retry or `false` (if retry is not needed):

```js
// Constant delay of 10ms for 5 iterations
got('google.com', {
    retries: iter => iter < 5 && 10
});

// Infinity retries with constant delay - this is not possible now, 
// because delay will increase up to the MaxInt :(
got('google.com', {
    retries: () => 10
});
```

I think this will cover like 99% of usecases and will not introduce new option.

floatdrop:
Now `retries` accepts custom backoff funtion, which allows next usecases:

```js
// Constant delay of 10ms for 5 iterations
got('google.com', {
    retries: iter => iter < 5 && 10
});

// Infinity retries with constant delay - this is not possible now, 
// because delay will increase up to the MaxInt :(
got('google.com', {
    retries: () => 10
});
```

Closes #138

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,11 +19,6 @@ var nodeStatusCodes = require('node-status-codes');
 var isPlainObj = require('is-plain-obj');
 var parseJson = require('parse-json');
 
-function backoff(iter) {
-	var noise = Math.random() * 100;
-	return (1 << iter) * 1000 + noise;
-}
-
 function requestAsEventEmitter(opts) {
 	opts = opts || {};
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Looks good to me.

@maxvipon ?

maxvipon:
:+1:

PR #144, opened 2015-12-15
fix body types in readme

stevemao:


julien-f:
Why are these considered typos?

floatdrop:
@julien-f because all other types are in lowercase.

julien-f:
I see, I generally tend to follow [Flow conventions](http://devdocs.io/flow/quick-reference) myself.

stevemao:
The types were wrong. body's type is `string` and it's an instance of `String`.

PR #146, opened 2015-12-20
Pass error to retries option

### LINKED ISSUE #119: Retry intermittent errors only
Retrying connection errors mindlessly isn't great. Not entirely sure this list is complete, but I've had a lot of luck with it:
```js
var TRANSIENT_HTTP_ERROR_CODES = [
  'ETIMEDOUT',
  'ECONNRESET',
  'EADDRINUSE',
  'ESOCKETTIMEDOUT',
  // One could argue against retrying ECONNREFUSED,
  // but this often happens if a remote service is restarting (or temporarily crashed)
  'ECONNREFUSED',
  'EPIPE'
];
```
Retring when encountering `ENOTFOUND` isn't useful (DNS resolution error)

Also if error code is 5xx retry is often the right thing. But that might something to specify as an option.

-----
I look at this module as something pretty neat for REST apis, where as many others don't have promise and retry support (and are much bigger)... So far liking what I see...

floatdrop:
This will resolve any issues with questions about `retries` option.

Closes #119

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,7 +56,7 @@ function requestAsEventEmitter(opts) {
 		});
 
 		req.once('error', function (err) {
-			var backoff = opts.retries(++retryCount);
+			var backoff = opts.retries(++retryCount, err);
 			if (backoff) {
 				setTimeout(get, backoff, opts);
 				return;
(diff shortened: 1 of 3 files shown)


sindresorhus:
:+1: 

// @jonasfj @techniq

PR #150, opened 2016-01-11
Update readme.md

f0rmat1k:


sindresorhus:
The current is valid, but I guess this is a bit more readable.

Issue #156, opened 2016-01-15
Don't get any error output when invalid url is requested


mrgodhani:
When I request invalid url e.g "http://example.comhttp://example.com" It doesn't catch / stop the execution and doesn't give me any kind of error.

sindresorhus:
That's unfortunately how promises works. They are silent by default. You have to explicitly `.catch()` it.

sindresorhus:
You might find https://github.com/sindresorhus/loud-rejection useful ;)

mrgodhani:
I just figure out that it was due to retries. By default it retries 5 times. So I had to explicitly make to 0.

PR #157, opened 2016-01-16
Filter not-retryable Errors

floatdrop:
`ENOTFOUND` and `ENETUNREACH` are most common errors, that we get when we have typos in configs or having IPv4 to IPv6 routing problems. I think these errors (or maybe some more) should fail fast without retries.

---

Related to #119 and #156


sindresorhus:
:shipit:

PR #159, opened 2016-01-20
Basic proxy support

tgandrews:
This PR is intended to start a discussion, possibly get some extra help and feedback on the direction It aims to resolve the issue #79 

N.B. I am working on removing the spacing and line ending changes to make the PR more readable.

---
Added really basic proxy support. It works for the simple use case of
got("reddit.com") which will first make a http request and then be
redirected to https. This also supports requests that are redirected and
over to unsecure locations i.e. they start http and stay http.

Currently requests that require https to begin with are not supported.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,319 +19,375 @@ const nodeStatusCodes = require('node-status-codes');
 const isPlainObj = require('is-plain-obj');
 const isRetryAllowed = require('is-retry-allowed');
 
+const shouldProxy = !!process.env.HTTP_PROXY;
+const parsedProxy = shouldProxy ? urlParseLax(process.env.HTTP_PROXY) : {};
+
 function requestAsEventEmitter(opts) {
-	opts = opts || {};
+  opts = opts || {};
 
-	const ee = new EventEmitter();
-	let redirectCount = 0;
-	let retryCount = 0;
+  const ee = new EventEmitter();
+  let redirectCount = 0;
+  let retryCount = 0;
 
-	const get = opts => {
-		const fn = opts.protocol === 'https:' ? https : http;
+  const get = opts => {
+    const fn = opts.protocol === 'https:' ? https : http;
+    const req = fn.request(opts, res => {
+      const statusCode = res.statusCode;
 
-		const req = fn.request(opts, res => {
-			const statusCode = res.statusCode;
+      if (isRedirect(statusCode) && 'location' in res.headers && (opts.method === 'GET' || opts.method === 'HEAD')) {
+        res.resume();
 
-			if (isRedirect(statusCode) && 'location' in res.headers && (opts.method === 'GET' || opts.method === 'HEAD')) {
-				res.resume();
+        if (++redirectCount > 10) {
+          ee.emit('error', new got.MaxRedirectsError(statusCode, opts), null, res);
+          return;
+        }
 
-				if (++redirectCount > 10) {
-					ee.emit('error', new got.MaxRedirectsError(statusCode, opts), null, res);
-					return;
-				}
+        const redirectUrl = urlLib.resolve(urlLib.format(opts), res.headers.location);
+        const parsedRedirect = urlLib.parse(redirectUrl);
+        let redirectOpts = Object.assign({}, opts, parsedRedirect);
+        if (shouldProxy) {
+          if (parsedRedirect.protocol === 'https:') {
+						const connectOptions = { // establishing a tunnel
+              host: parsedProxy.hostname,
+              port: parsedProxy.port,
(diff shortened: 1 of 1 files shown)


sindresorhus:
From that issue:

> I would prefer for the solution to be an external module got can just depend on as I don't want to have to maintain it. Maybe https://github.com/mikeal/tunnel-agent.

tgandrews:
Sorry, I should have read the issue properly.

I will move it. I guess you would rather have it replace http/https?

Probably will make it easier from my PoV as well.

tgandrews:
Closing as will create a new one referencing a new npm lib.

sindresorhus:
> I guess you would rather have it replace http/https?

Yes, or even better, a custom http agent.

stevenvachon:
https://github.com/TooTallNate/node-proxy-agent ?

tgandrews:
I have a branch with this working. No tests yet. I will push on Monday once
I have some tests working.

alexsantos:
@tgandrews , I've seen your branch but what I miss is a way where _got_ uses proxies in a transparent way, without passing a opts.proxy value. This follows the need for proxies when using tools like npm-name, generator-node, etc... I've did some work on https://github.com/sindresorhus/got/issues/178 following the approach from request but @sindresorhus pointed out how wrong it could get :-1:

tgandrews:
@alexsantos I'll try and dig out the branch. I had a bit of down time in a new job (with big company network setup) but have not looked at this for a while.

alexsantos:
@tgandrews and @sindresorhus , I have another branch which I think is the minimalist solution for this. It depends on proxy-from-env and proxy-agent modules. Can you have a look here? https://github.com/alexsantos/got/tree/proxy-from-env/index.js
Basically, it's 6 lines of code...
```javascript
if (!opts.agent) {
  const proxy = getProxyForUrl(url);
  if (proxy) {
    opts.agent = new ProxyAgent(proxy);
  }
}
```
I could have done a PR but didn't want to bug Sindre again :-)

sindresorhus:
@alexsantos Yes, something like that is what I was thinking. Unfortunately the `proxy-agent` dependency takes up 7.4 MB, which is way too much.

sindresorhus:
// @floatdrop

floatdrop:
May be [same trick as in http2 issue](https://github.com/sindresorhus/got/issues/167#issuecomment-179783180) can be applied here as well?

sindresorhus:
@floatdrop Unfortunately not to achieve what most people requesting this want, as they want modules or modules of modules using got somewhere in the chain to support proxies transparently. Should be possible to find/create a smaller dependency though.

alexsantos:
@floatdrop , just to back @sindresorhus message, I came here because I want to use yeoman generators at work and I can't because of the lack of proxy support on got.

tgandrews:
@sindresorhus @alexsantos  My branch was using proxy-agent as well. I guess we need to support just HTTP and HTTPS proxies? If we are doing the whole whack then I doubt we are going to do better than node-proxy-agent

stevenvachon:
This feature is probably one of the things that makes `request` lib as big as it is.

alexsantos:
@tgandrews The size of node-proxy-agent is derived from the fact that it supports PAC and that module specfically needs degenerator/esprima which is a huge dependency. If we abandon PAC then it gets pretty small. I've been doing some work on that module so I'll might drop PAC and fork it to a smaller one.

tgandrews:
OK, I guess there are two options:
1. Fork node-proxy-agent and remove PAC support.
2. Work on node-pac-proxy-agent and reduce the external dependencies to make this library smaller.

And @alexsantos is looking(?) into number 1, so I'll have a look at number 2 then if I get some free time.

alexsantos:
@tgandrews , I already have a stripped version of node-pac-resolver (now with Promises) that gets rid of regenerator. https://github.com/alexsantos/node-pac-resolver/tree/nodev4-compat
Forking the whole node-proxy-agent would also bring this modules to `got` standards, like v4, xo, nyc, etc...

tgandrews:
@alexsantos Let me know if there is anything I can do to help.

stevenvachon:
Why fork node-proxy-agent when we could create a PR to bring it up to date? Or was that the intent?

alexsantos:
@stevenvachon , node-proxy-agent has pac-proxy support, which is a huge dependency for got.

stevenvachon:
Do we not need to support pac and socks proxies?

bhushankumarl:
Does it support socks5 proxy?
`socks5://127.0.0.1:3526`

Issue #165, opened 2016-02-02
Option to disable following redirects


ruyadorno:
Hello there, first of all, many thanks for all the hard work that went into this module :blush: 

I have a suggestion that may or may not make any sense for it to be implemented on **got** - that would be to add an option to disable the automatic following of redirects.

I'm using **got** in an express app that serves as a middleman/proxy between the client and other servers and I would love to have the possibility to just forward the 3xx status codes from the original server.

Thanks for your time and keep up with the good work :zap:

floatdrop:
@ruyadorno thanks for the nice issue and PR!

After some thinking I'm quite positive, that `got` is not the best solution to proxy request from other server to client – `got` will decompress content and send broken response with decompressed content and `content-encoding: gzip` (if you just passing headers by).

May be it will be a better solution to use native `http` module instead?

ruyadorno:
hi @floatdrop!

I understand your reasoning, I have thought the same thing and as a matter of fact I actually reimplemented three times the same micro-service to try different approaches but **got** was the clear winner for me.

The reason being that I actually want to fetch all the data from the server, parse it, modify and send it back to the user - **got** abstracts a lot of the hard work from that, such as gzipping on prod but not on local development, not to say the beautiful API it provides.

That said, I really don't mean to make any pressure, please take time to evaluate if this is a constructive addition to other users/use cases and feel free to close the PR if that's not the case :blush: 

:+1: thanks again!

dfilatov:
+1 to add this feature. I have some cases when I have to parse location of redirect.

floatdrop:
Released in `5.6.0` and `6.3.0` :tada:

dfilatov:
@floatdrop Thanks!

PR #166, opened 2016-02-02
Added option to toggle automatic following redirects

ruyadorno:
Hi @floatdrop and @sindresorhus,

I'm opening this PR with the goal of illustrating what such an option as described on #165 would look like.

Please review at your own discretion. 

Thanks in advance :+1:

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,7 @@ function requestAsEventEmitter(opts) {
 		const req = fn.request(opts, res => {
 			const statusCode = res.statusCode;
 
-			if (isRedirect(statusCode) && 'location' in res.headers && (opts.method === 'GET' || opts.method === 'HEAD')) {
+			if (isRedirect(statusCode) && opts.followRedirect && 'location' in res.headers && (opts.method === 'GET' || opts.method === 'HEAD')) {
 				res.resume();
 
 				if (++redirectCount > 10) {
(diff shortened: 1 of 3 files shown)


knksmith57:
+1

sindresorhus:
After much thinking I'm :+1: on this. @floatdrop You ok with this?

floatdrop:
@sindresorhus how about merge this with `maxRedirects` option? For example `-1` in `maxRedirects` will disable redirect following.

sindresorhus:
@floatdrop We don't have a `maxRedirects` option. ?

floatdrop:
@sindresorhus yep, just an idea.

sindresorhus:
I'm not a big fan of `-1` input, and in that case it could just be `0`, but still, I don't really see the need of a `maxRedirects` option, and since we have neither at the moment, I'd rather just expose a `followRedirects` option that is actually requested by multiple people.

ruyadorno:
_rebased_

RoobinGood:
@ruyadorno are you sure that making redirect a successful result for request is a good idea? As far as I know it is not usual behavior.

ruyadorno:
@RoobinGood that's a good point, I'm not really sure of anything :blush: 

It does make sense in the context of the original use case behind this idea, as described in #165 but at the end of the day I just want to access the original redirect HTTP response in case of a 300 error, I guess that could also be achieved by attaching the response to the thrown `got.HTTPError` object but then I'm even less sure that's a better option.

It's also worth noting that we're talking about an optional configuration, it won't change the way **got** behaves if you don't use the `followRedirect` config property.

floatdrop:
@sindresorhus @ruyadorno are we ready to merge this option?

ruyadorno:
👍 sure, it was a good point brought to the discussion by @RoobinGood but I still think the current implementation is better 😊 we're getting more people interested in having this option... let's merge this

PR #168, opened 2016-02-10
parse json based on content-type

Autarc:
Automatic content negotiation if the header is set (prevents missing the `json` option).

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,10 +97,11 @@ function asPromise(opts) {
 				.catch(err => reject(new got.ReadError(err, opts)))
 				.then(data => {
 					const statusCode = res.statusCode;
+					const contentType = res.headers['content-type'];
 
 					res.body = data;
 
-					if (opts.json && statusCode !== 204) {
+					if (statusCode !== 204 && (opts.json || (contentType && contentType.indexOf('application/json') > -1))) {
 						try {
 							res.body = JSON.parse(res.body);
 						} catch (e) {


floatdrop:
[`json`](https://github.com/sindresorhus/got#json) is intentionally non-conditional – some servers will not send `application/json` content-type even if they respond in JSON. If option is set to `true` it should always parse body.

Autarc:
Sure it shouldn't replace the existing option but just extend it for convenience. If someone expects the response of a request to be JSON because its mentioned in a documentation, it shouldn't be necessary to manually set it as the server already provides the hint.

sindresorhus:
@floatdrop Most servers are compliant, though, and it would be nice not having to set the JSON option for those. Maybe we should think of the JSON option as an override in the edge-case of the server being non-compliant?

julien-f:
I tend to agree with @floatdrop because getting an object when not expecting one is not very nice :)

Although an explicit option (`jon: 'auto'`) could do.

SamVerschueren:
But then it should be possible to set `json` to `false` in order to not parse it, even if the `Content-Type` is `application/json`. With the current implementation, that is not possible.

sindresorhus:
@SamVerschueren Yup

floatdrop:
> Most servers are compliant, though, and it would be nice not having to set the JSON option for those.

@sindresorhus yes, but if server changes behaviour (or something with proxy in front of it) – worst case scenario is that header and content will change. `got` will get you plain string object and (depends on your code) it can be silently processed (strings have props, right?). In current case you will get `ParseError` immediately.

@SamVerschueren this is possible (or I don't see something), but this is major change.

SamVerschueren:
If `opts.json` is `false` but the content type is `application/json`, the statement still resolves to true.

sindresorhus:
@floatdrop That's true, but if the server sends `application/json`, I think we should trust the server. The whole point of content-type's is that server can indicate what it supports. If it suddenly changes behavior it can break a lot of other things regardless of this change.

floatdrop:
After some time thinking – I'm sure, that automatic negation is not useful. Heres why:

When I want to get parsed data from endpoint I must do two things – ensure, that endpoint is returning json (almost always this easy to see in docs) and ensure, that `content-type` header is in place. With strict options I can skip last part and just type `json: true`.

Also I must worry about `404`/`503` pages, which almost _never_ contains json and spit HTML on to you (with `text/html` content-type of course). This leads not to ParseError, but to very strange behaviours.

May be we can figure out how to solve these issues and get automatic negation about content (`json: 'auto'` maybe as stated above). But this solution should be improved.

PR #171, opened 2016-02-15
Fix user-agent string RFC 2616 Section 14.43

### LINKED ISSUE #169: User-agent string warning with ':'
Default user-agent string in got contains ':' character. But many servers shows warnings that this character not allowed in user-agent string

birhoff:
Fixes #169.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ const createErrorClass = require('create-error-class');
 const nodeStatusCodes = require('node-status-codes');
 const isPlainObj = require('is-plain-obj');
 const isRetryAllowed = require('is-retry-allowed');
+const pkg = require('./package.json');
 
 function requestAsEventEmitter(opts) {
 	opts = opts || {};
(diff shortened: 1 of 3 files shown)


floatdrop:
@birhoff can you fix the tip [here](https://github.com/sindresorhus/got#tip) too?

index.js
```diff
@@ -202,7 +203,7 @@ function normalizeArguments(url, opts) {
 	);
 
 	opts.headers = Object.assign({
-		'user-agent': 'https://github.com/sindresorhus/got',
+		'user-agent': `sindresorhus-got/${version} (https://github.com/sindresorhus/got)`,
```
sindresorhus: Actually, since we include the URL, we can remove `sindresorhus-` from the first part.

birhoff:
this is not that one? https://github.com/sindresorhus/got/pull/171/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R269

floatdrop:
@birhoff oops :blush:

birhoff:
:up:

PR #172, opened 2016-02-16
Parse the response body only if it is not empty

### LINKED ISSUE #170: ParseError on response with no body
I'm using `got` to query a JSON API where some endpoints simply return `200` with no body.

`got` throws a [`ParseError`](https://github.com/sindresorhus/got/blob/3dc219e70474e54c4ec4b2c27468f6ff722f8c6b/index.js#L107) because it only skips parsing if the status code is `204`.

I wonder if it makes sense to skip parsing every time the body is empty, without checking the status code.

```js
if (opts.json && res.body) {
  // try to parse
}
```

I'm happy to send a pull request if you think the above change would be useful.

lpinca:
Fixes #170.

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,7 +101,7 @@ function asPromise(opts) {
 
 					res.body = data;
 
-					if (opts.json && statusCode !== 204) {
+					if (opts.json && res.body) {
 						try {
 							res.body = JSON.parse(res.body);
 						} catch (e) {
--- a/test/json.js
+++ b/test/json.js
@@ -15,8 +15,8 @@ test.before('setup', async () => {
 		res.end('/');
 	});
 
-	s.on('/204', (req, res) => {
-		res.statusCode = 204;
+	s.on('/no-body', (req, res) => {
+		res.statusCode = 200;
 		res.end();
 	});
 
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -101,7 +101,7 @@ function asPromise(opts) {
 
 					res.body = data;
 
-					if (opts.json && statusCode !== 204) {
+					if (opts.json && res.body) {
```
sindresorhus: Wonder if this makes more sense:

```js
if (opts.json && statusCode !== 204 && res.body) {
```

This way we don't try to read the body, even if there is one when the server tells us there shouldn't be one.

@floatdrop Thoughts?

lpinca: Does it make sense to return the unparsed body?
If I'm not mistaken if the status code is `204` and there is a body we only prevent it from being parsed.

floatdrop: I'm agree @sindresorhus – if status code is `204` we should not try to parse body, even if there is one.

lpinca: Ok, I will update it and add a couple of tests in a bit.

lpinca: I think we can remove the check for the `204` status code completely as no data chuck is fired on the client.
Here is a test case:

```js
'use strict';

const http = require('http');

http.createServer((req, res) => {
  res.statusCode = 204;
  res.end(JSON.stringify({ a: 1 }));
}).listen(3000);
```

```js
'use strict';

const http = require('http');

http.get('http://localhost:3000', res => {
  res.on('data', () => {
    throw new Error('Chunk retrieved');
  });
});
```
In short, there will never be something to parse if the status code is 204.

sindresorhus: :+1:

sindresorhus:
Thanks :)

PR #178, opened 2016-02-25
Proxy support

alexsantos:
This a PR for proxy support on got, using the environment variables:

- HTTP_PROXY / http_proxy
- HTTPS_PROXY / https_proxy
- NO_PROXY / no_proxy

For now, it supports these no_proxy use cases:
- hostame, hostname:port
- ipv4, ipv4/range
- ipv6 (no ipv6/range)
- *

Tests
There are 13 tests on test/proxy.js, covering both http and https hostnames, ipv4, ipv6 and socks. The https proxy has problems with self signed certificates, so no tests on that.

Please comment on this.

Alex


sindresorhus:
From #79:

> I would prefer for the solution to be an external module got can just depend on as I don't want to have to maintain it.

-

Also see https://github.com/sindresorhus/got/pull/159, where @tgandrews said he was working on a lib.

alexsantos:
Hi @sindresorhus , I have to agree on that (about the way request is doing it) :+1: 
I'll move this to an individual module.

Issue #176, opened 2016-02-24
Not usable in atom


artoale:
I'm developing an Atom plugin that sends requests through GOT. 
First of all, to import the module I have to do 
```javascript
import { allowUnsafeEval } from 'loophole';

allowUnsafeEval( () => {
  got = require('got');
});
```

Since GOT uses (indirectly) `eval`, which breaks ATOM csp policy. 
Now, if the request is successful, got works fine, but when a network error occurs (e.g. 404) I get:

```
At evalmachine.<anonymous>:1

ReferenceError: captureStackTrace is not defined
    at new HTTPError (evalmachine.<anonymous>:1:25)
    at /Users/aartoni/dev/atom-codebase/node_modules/got/index.js:105:11
    at BufferStream.<anonymous> (/Users/aartoni/dev/atom-codebase/node_modules/read-all-stream/index.js:64:3)
    at emitNone (events.js:72:20)
    at BufferStream.emit (events.js:166:7)
    at finishMaybe (/Users/aartoni/dev/atom-codebase/node_modules/readable-stream/lib/_stream_writable.js:511:14)
    at afterWrite (/Users/aartoni/dev/atom-codebase/node_modules/readable-stream/lib/_stream_writable.js:390:3)
    at doNTCallbackMany (node.js:463:18)
    at process._tickCallback (node.js:361:17)
```

I believe it all boils down to custom Errors issue in the hybrid electron environment, and I'd understand if you think it's not worth fixing, but maybe there are other viable solutions not using eval?

floatdrop:
`eval` is used in [`create-error-class`](https://github.com/floatdrop/create-error-class/blob/master/index.js#L30) to provide name to constructor. May be it can be achieved in other way or removed there. Can you reopen issue there?

julien-f:
@floatdrop This has been fixed in [make-error](https://github.com/julien-f/js-make-error), perhaps you should consider using it…

floatdrop:
@artoale fixed in `create-error-class@3.0.1`.

@julien-f I don't remember, why I ignored this module :blush: Can you do a PR?

julien-f:
Perhaps because it's not easy to find good modules on npm :p
Or perhaps I did not use the right keywords :)

Anyway, if you prefer make-error over create-error-class I believe you should deprecate latter in favor of the first.

artoale:
@floatdrop thanks for the super quick fix. Working perfectly now!

PR #179, opened 2016-02-25
Use make-error instead of create-error-class (see #176).

julien-f:
If you think it's worth it, I am not trying to impose my lib :laughing:

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ const urlParseLax = require('url-parse-lax');
 const lowercaseKeys = require('lowercase-keys');
 const isRedirect = require('is-redirect');
 const unzipResponse = require('unzip-response');
-const createErrorClass = require('create-error-class');
+const makeError = require('make-error');
 const nodeStatusCodes = require('node-status-codes');
 const isPlainObj = require('is-plain-obj');
 const isRetryAllowed = require('is-retry-allowed');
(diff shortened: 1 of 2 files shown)


sindresorhus:
@floatdrop Is this something you still want?

floatdrop:
@sindresorhus I don't know, they quite same to me.

sindresorhus:
@julien-f Would there be any actual benefits of switching?

julien-f:
@sindresorhus Nope, no benefits other than avoiding a duplicate library :)

By the way, I am getting fed up with abandoned repos were the maintainers (and often only npm owner) do not answer anymore so I created a GitHub organization (@JsCommunity) to avoid this kind of issues. Any opinion on this? (This might not be the best place to discuss about this, feel free to open [an issue there](https://github.com/JsCommunity/jscommunity.github.io/issues/new)).

sindresorhus:
Ok, then I don't really see the point of switching.

-

> Any opinion on this? (This might not be the best place to discuss about this, feel free to open an issue there).

Can you open an issue there called `Feedback` and mention me?

sindresorhus:
@floatdrop Now that we target Node.js 4, wouldn't it be better to actually subclass the errors?

For example:

```js
class HttpError extends Error {
	constructor(message) {
		super(message);
		this.name = this.constructor.name;
		this.message = 'Unicorns';
	}
}
```

floatdrop:
@sindresorhus to be honest, I have not yet used `class` in my code, so – dunno. I guess this is right decision, but looking at code in [`make-error`](https://github.com/JsCommunity/make-error/blob/master/index.js#L41-L72) it could have some bumps. 

@julien-f can you advise – how safe is to `extend` from `Error` class directly?

julien-f:
IMHO, if you want to be compatible with browsers, you should avoid extend `Error` for now.

sindresorhus:
@julien-f Why?

julien-f:
I don't know the ratio of browsers which correctly support inheriting from `Error`.

sindresorhus:
Yeah, I don't know either. Just assumed it would be ok since we already target Node.js 4 JS features.

We can do it sometime in the future though: https://github.com/sindresorhus/got/issues/190

PR #183, opened 2016-03-03
Add statusCode to ParseError, closes #182.

### LINKED ISSUE #182: ParseError should contain err.statusCode
`curl -v https://api.github.com/repos/震动/generator-test` should output a json but it does print an HTML pages, which gives a `ParseError` if we use `got()` with `options.json=true`. Though, I was expecting that the ParseError contained the `err.statusCode` property, which doesn't at the moment. It has to be accessed via `err.response.statusCode` which kind sucks :(

Thoughts?

satazor:
Let me know if everything is as expected.

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,7 +105,7 @@ function asPromise(opts) {
 						try {
 							res.body = JSON.parse(res.body);
 						} catch (e) {
-							throw new got.ParseError(e, opts, data);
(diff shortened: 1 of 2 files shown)


floatdrop:
I think `statusCode` and `statusMessage` should be metion in [readme](https://github.com/sindresorhus/got#errors) also.

PR #189, opened 2016-03-30
Add followRedirect option

RoobinGood:
Add option followRedirect to be able to not follow redirects.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,7 @@ function requestAsEventEmitter(opts) {
 		const req = fn.request(opts, res => {
 			const statusCode = res.statusCode;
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Duplicate of https://github.com/sindresorhus/got/pull/166. Can you elaborate on why you need this?

RoobinGood:
@sindresorhus I've just need this options in my tests on work project to check if the resource is reachable for user. If resource is not reachable server redirects user, so I need `followRedirect` is `false` to check status and location.
Also this options is represented in [`request`](https://github.com/mikeal/request). Previously I used `request` for testing, but `got` is lightly and easier. However I need some compatibility with `request`.

floatdrop:
Implemented in #166.

PR #192, opened 2016-04-05
Updated readme with a tip about getting got@6 working in the browser.

Darkle:


sindresorhus:
I don't see the point of this. Just use version 5 if you need to support older browsers. Browsers are not the main target of `got` anyways.

Darkle:
Fair enough.

PR #193, opened 2016-04-05
Update create-error-class dependency

Darkle:
I was running in to an issue with running got in the browser because create-error-class version 2.0.0 uses eval and I was trying to set a strict Content Security Policy - so updating the dependency.


floatdrop:
`+ got@5.5.1`

PR #198, opened 2016-05-11
Do not set default value for opts.retiries if timeout is present.

Firnis:
It quite isn't right to retry request 5 times when timeout is present in options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,8 @@ const isPlainObj = require('is-plain-obj');
 const isRetryAllowed = require('is-retry-allowed');
 const pkg = require('./package.json');
 
+const DEFAULT_RETRIES = 5;
+
 function requestAsEventEmitter(opts) {
 	opts = opts || {};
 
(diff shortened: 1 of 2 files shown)


SamVerschueren:
Why isn't that correct? A timeout is per individual call.

Let's say I have an endpoint and I set the timeout to 10 seconds. With this PR, this means that my retry count is set to 0 and will stop executing if the first call fails due to a network error. `timeout` and `retries` or 2 separate settings, I don't believe there is a link between those two.

Firnis:
But if your endPoint is not answering, then you will be waiting 5*10 seconds.
I see your point, but you can set count of retries manually if you need.
When I set some timeout, it means that I can't wait any longer.

sindresorhus:
This is not a solution we're interested in. I do however see that it how it works can be an issue. I don't really have a solution, but happy to discuss in an issue if you have a better proposal :)

PR #210, opened 2016-07-11
Handle errors thrown synchronously by http.request().

satazor:
```js
const got = require('got');

got('htttp://google.com')
.then((res) => console.log('res', res), (err) => console.log('error', err));
```

The code above fails globally because `htttp` is not a valid protocol. This PR fixes it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,36 +29,41 @@ function requestAsEventEmitter(opts) {
 
 	const get = opts => {
 		const fn = opts.protocol === 'https:' ? https : http;
+		let req;
 
-		const req = fn.request(opts, res => {
-			const statusCode = res.statusCode;
+		try {
+			req = fn.request(opts, res => {
+				const statusCode = res.statusCode;
 
-			if (redirectUrl) {
-				res.url = redirectUrl;
-			}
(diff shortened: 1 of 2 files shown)


floatdrop:
I believe [this code](https://github.com/sindresorhus/got/blob/master/index.js#L290-L294) should catch such errors (and we have [similar test](https://github.com/sindresorhus/got/blob/master/test/error.js#L33)).

Also this test is green:

```js
test('invalid url', async t => {
	got('htttp://google.com').then(t.fail).catch(err => {
		t.regex(err.message, /Protocol .+ not supported/);
		t.pass();
	});
});
```

Can you make a PR with failing test?

satazor:
@floatdrop sorry the test is wrong, I should have used `got.stream` since the bug is there.

How should we adjust this then?

satazor:
@floatdrop maybe it's ok to throw synchronously since this method does not return a promise?

floatdrop:
> maybe it's ok to throw synchronously since this method does not return a promise?

@satazor I think so.

satazor:
Agree! Closing this.

PR #214, opened 2016-07-31
Fix location encoding

luanmuniz:
See https://github.com/nodejs/node/issues/7928

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ function requestAsEventEmitter(opts) {
 					return;
 				}
 
-				redirectUrl = urlLib.resolve(urlLib.format(opts), Buffer.from(res.headers.location, 'binary').toString());
+				redirectUrl = urlLib.resolve(urlLib.format(opts), new Buffer(res.headers.location, 'binary').toString());
 				const redirectOpts = Object.assign({}, opts, urlLib.parse(redirectUrl));
 
 				ee.emit('redirect', res, redirectOpts);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Needs a test

luanmuniz:
On it

sindresorhus:
// @jamestalmage FYI, for [follow-redirects](https://github.com/olalonde/follow-redirects).

luanmuniz:
@sindresorhus Its fixed, but i'm having a small problem if compatibility here.

I found 2 ways to fix this:
`new Buffer(res.headers.location, 'binary').toString()`
and
`Buffer.from(res.headers.location, 'binary').toString()`

But i have the following problem: While the first one is [deprecated in v6](https://nodejs.org/api/buffer.html#buffer_new_buffer_arraybuffer_byteoffset_length) the second [doesn't exist in v4](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length) and since the tests check for 4, 5 and 6 i don't know how it should be.

Obs.: It's working with the first option
Obs2.: Don't merge it yet, i will write tests for this case

luanmuniz:
@sindresorhus Done. Ready to merge

luanmuniz:
@sindresorhus Anything else to do here?

floatdrop:
@luanmuniz lgtm, thanks :star:

PR #217, opened 2016-08-04
allow non-plain object as request body

### LINKED ISSUE #212: Request body can't be querystring.parse(str)
When the request is supplied with body that differs from `{ ... }` object, it throws

> options.body must be a ReadableStream, string, Buffer or plain Object

`is-plain-object` expects an object to inherit from either `null` or `Object` directly.

This makes 

```
got.post(url, { data: querystring.parse(query) });
```

to fail, because `querystring.parse(...)` object [inherits from empty object that inherits from `null`](https://github.com/nodejs/node/blob/v6.3.0/lib/querystring.js#L11-L12).

This looks more like a bug rather than zealous sanitization.

bisubus:
Closes #212


sindresorhus:
👍 

@floatdrop ?

floatdrop:
@sindresorhus test on this would be nice 🦄

sindresorhus:
Agreed.

@bisubus Would you mind adding a test and fixing the merge conflict?

bisubus:
@sindresorhus Sure thing, I'll get down to this PR a bit later.

bisubus:
@sindresorhus Added a couple of tests to match an object that can be passed as a body, since it is no longer validated by `is-plain-obj`.

floatdrop:
@bisubus thank you! :tada:

yeze322:
Was this PR reverted? Seems the validator is still using `is-plain-obj` for now, class instances cannot be used as `body`.

bisubus:
@yeze322 Yes, it looks like there was a regression in #297. A quick fix is `Object.assign({}, obj)`.

PR #220, opened 2016-08-14
Detect formdata body and set content-type header.

DylanPiercey:
This PR adds a check for [Form-Data](https://github.com/form-data/form-data) which makes it easier for users to submit form data bodies (no longer have to manually get headers from form-data).

This is similar to https://github.com/bitinn/node-fetch/pull/31, the idea is just to make it easier on end users since [Form-Data](https://github.com/form-data/form-data) is commonly used throughout the community. It will also make isomorphic code easier in the future with [isomorphic-form-data](https://github.com/form-data/isomorphic-form-data).

I realize this is missing a new test, I wasn't sure what the best way to go about it was.
If I got some direction I could go ahead with tests or someone else could take over from here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -242,7 +242,10 @@ function normalizeArguments(url, opts) {
 
 		opts.method = opts.method || 'POST';
 
-		if (isPlainObj(body)) {
+		if (isStream(body) && typeof body.getBoundary === 'function') {
+			// Special case for https://github.com/form-data/form-data
+			opts.headers['content-type'] = opts.headers['content-type'] || 'multipart/form-data; boundary=' + body.getBoundary();
+		} else if (isPlainObj(body)) {
 			opts.headers['content-type'] = opts.headers['content-type'] || 'application/x-www-form-urlencoded';
 			body = opts.body = querystring.stringify(body);
 		}
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   "devDependencies": {
     "ava": "^0.16.0",
     "coveralls": "^2.11.4",
+    "form-data": "^1.0.1",
     "get-port": "^2.0.0",
     "get-stream": "^2.3.0",
     "into-stream": "^2.0.0",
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -242,7 +242,10 @@ function normalizeArguments(url, opts) {
 
 		opts.method = opts.method || 'POST';
 
-		if (isPlainObj(body)) {
+		if (isStream(body) && typeof body.getBoundary === 'function') {
+			// Special case for https://github.com/form-data/form-data
+			opts.headers['content-type'] = opts.headers['content-type'] || 'multipart/form-data; boundary=' + body.getBoundary();
```
floatdrop: `content-type` is not the only header, that can be returned from `getHeaders` function. This line should contain call to `getHeaders` and assign user supplied headers over it.

sindresorhus:
The test should send a request and check that correct headers were actually sent. See other tests for how to do it.

sindresorhus:
ping @DylanPiercey :)

DylanPiercey:
@sindresorhus sorry forgot about this one, I'll try to get to it later today.

@floatdrop currently formdata getHeaders only returns content-type and that will not change so I figured this was the most appropriate api. Especially since in @2.0 the getHeaders function may have a deprecation warning.

DylanPiercey:
@sindresorhus does that look better? (Added some tests).

DylanPiercey:
Let me know if anything else is needed @sindresorhus 😀.

index.js
```js
`multipart/form-data; boundary=${body.getBoundary()}`;
```

sindresorhus:
LGTM. @floatdrop ?

floatdrop:
> Especially since in @2.0 the getHeaders function may have a deprecation warning.

Okay then.

> LGTM when #220 (comment) is fixed. @floatdrop ?

Lgtm. Waiting for https://github.com/sindresorhus/got/pull/220#discussion_r77452874

sindresorhus:
Actually, I'll just fix it when merging. Thanks @DylanPiercey :)

PR #230, opened 2016-10-19
Add `requestUrl` to response in streaming mode

### LINKED ISSUE #229: Add `url` property to `got.stream()` request and response objects
```js
require("got").stream("http://google.com/")
.on("request", request => console.log(request))
.on("response", response => console.log(response));
```
Pretty well every other http request library provides a `url` property in these objects. Why doesn't this one?

kevva:
Fixes #229.

### DIFF
--- a/index.js
+++ b/index.js
@@ -173,6 +173,7 @@ function asStream(opts) {
 		const statusCode = res.statusCode;
 
 		res.pipe(output);
+		res.requestUrl = opts.href || urlLib.resolve(urlLib.format(opts), opts.path);
 
 		if (statusCode < 200 || statusCode > 299) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
We now set it twice in different places. I think we could set it once in the `requestAsEventEmitter` instead, though haven't tested it.

kevva:
@sindresorhus, yeah, I thought about that as well. It'd work if we passed the initial URL every time we call `get()` inside `requestAsEventEmitter`. I just thought this was clearer.

kevva:
@sindresorhus, done.

sindresorhus:
@floatdrop LGTY?

floatdrop:
👍

PR #243, opened 2016-11-04
fix TypeError: binary is not a function

tommytroylin:
fix TypeError: binary is not a function
on node versions below 4.5

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,8 +40,13 @@ function requestAsEventEmitter(opts) {
 					ee.emit('error', new got.MaxRedirectsError(statusCode, opts), null, res);
 					return;
 				}
-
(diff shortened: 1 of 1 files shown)


sindresorhus:
Just do feature testing by checking if `Buffer.from` exists.

tommytroylin:
@sindresorhus Actually  `Buffer.from` function is exist on node 4.4.4 runtime
So `Buffer.from ? Buffer.from(...) : new Buffer(...)` won't work.

floatdrop:
@tommytroylin replaced `process.version` with future testing – we prefer to keep `process` variable out of code, because it will be shimmed in browserified version of `got`.

Issue #251, opened 2016-12-20
Got throws an HTTPError for 304 responses


lukechilds:
Currently implementing an API client that supports caching. It sends if-modified-since headers and will sometimes get a 304 response. If I get a 304, got throws an HTTPError. I can't catch the error and then do my stuff because the error doesn't contain the headers.

I noticed this in the readme:

>#### got.HTTPError
>
>When server response code is not 2xx. Contains `statusCode` and `statusMessage`.

Why isn't 304 allowed?

floatdrop:
@lukechilds `got` treats redirects as errors only on `POST`-like methods, otherwise it will follow redirect location (up to 5 redirects before throwing error). Can you post code to reproduce problem?

lukechilds:
304 isn't a redirect, it means the content hasn't been modified.

I'm doing a GET request with a date set in the if-modified-since header. The API is returning a 304 meaning the resource hasn't changed since the copy in my cache but this causes `got` to throw an error.

2 secs, just putting some code together.

floatdrop:
@lukechilds yup, you right – 304 case will be treated as error. But [error does contains headers](https://github.com/sindresorhus/got#errors) under `response` property:

> In Promise mode, the response is attached to the error.

So you can check for `statusCode` or headers in `error` object.

Maybe it is better to make this case valid and not throw error thou.

lukechilds:
```js
const got = require('got')

// Make API request
got('https://onionoo.torproject.org/summary?limit=1')
  .then(res => {

      // Log response info (should be 200)
      console.log(res.statusCode, res.headers)

      // Make another request with if-modified-since set
      // This should give us a 304
      return got('https://onionoo.torproject.org/summary?limit=1', {
        headers: {
          'if-modified-since': res.headers['last-modified']
        }
      })
  })

  // This should run, but got throws an error
  .then(res => console.log(res.statusCode, res.headers))

  // You can see we got a successful 304 response
  .catch(console.log)
```

lukechilds:
Oops, just seen your reply.

And the headers aren't in the error object, that's my problem.

floatdrop:
@lukechilds `response` property is hidden from `console.log` (because it is huge). Try:

```js
.catch((err) => console.log(err.response))
```

lukechilds:
Ahhh, thank you, I didn't realise that. Well that gives me a way to resolve this on my end. It definitely seems odd to me that it throws on a 304 though.

[...]

PR #252, opened 2016-12-20
Don't throw HTTPError on 304 responses

lukechilds:
Resolves https://github.com/sindresorhus/got/issues/251

### DIFF
--- a/test/http.js
+++ b/test/http.js
@@ -15,6 +15,11 @@ test.before('setup', async () => {
 		res.end();
 	});
 
+	s.on('/304', (req, res) => {
+		res.statusCode = 304;
+		res.end();
+	});
+
 	s.on('/404', (req, res) => {
 		setTimeout(() => {
 			res.statusCode = 404;
@@ -54,6 +59,14 @@ test('error with code', async t => {
 		t.is(err.statusCode, 404);
 		t.is(err.response.body, 'not');
 	}
+});
+
+test('status code 304 doesn\'t throw', async t => {
+	const p = got(`${s.url}/304`);
+	await t.notThrows(p);
+	const response = await p;
+	t.is(response.statusCode, 304);
+	t.is(response.body, '');
 });
 
 test('buffer on encoding === null', async t => {
--- a/test/http.js
+++ b/test/http.js
@@ -21,10 +21,8 @@ test.before('setup', async () => {
 	});
 
 	s.on('/404', (req, res) => {
-		setTimeout(() => {
-			res.statusCode = 404;
-			res.end('not');
-		}, 10);
+		res.statusCode = 404;
+		res.end('not');
 	});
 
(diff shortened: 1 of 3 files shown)


lukechilds:
@sindresorhus Continuing discussion from here: https://github.com/sindresorhus/got/issues/251#issuecomment-285131422

I've added a test 👍

When writing it I noticed that the [mock 404 route](https://github.com/sindresorhus/got/blob/a6907b439919166a3f5d9d18cdef3afc5440245f/test/http.js#L18-L23) is wrapped with `setTimeout`? Why is this? Tests still pass if I comment out `setTimout`. Should I be doing that for my 304 route?

Regarding the readme update, is it really necessary to mention a 304 response will have an empty body in the readme? That's normal HTTP behaviour, if people are sending `if-modified-since` headers then they are gonna be expecting to handle 304s.

sindresorhus:
> When writing it I noticed that the mock 404 route is wrapped with setTimeout? Why is this? Tests still pass if I comment out setTimout. Should I be doing that for my 304 route?

Nah, you don't need. According to `git blame`, it was added to workaround some flakyness. Probably related to older Node.js versions. https://github.com/sindresorhus/got/commit/46d0d4a4f404d633beb4a5235cf2fec1e9b4fc1a

@floatdrop Let's just remove that setTimeout now? I don't think it applies anymore.

> Regarding the readme update, is it really necessary to mention a 304 response will have an empty body in the readme? That's normal HTTP behaviour, if people are sending if-modified-since headers then they are gonna be expecting to handle 304s.

Expecting users to know something is like asking for support questions. Doesn't hurt to have a quick reminder about it.

test/http.js
```diff
@@ -56,6 +61,16 @@ test('error with code', async t => {
...
 });
 
+test('304 doesn\'t throw', async t => {
+	try {
+		const response = await got(`${s.url}/304`);
+		t.is(response.statusCode, 304);
+		t.is(response.body, '');
+	} catch (err) {
+		t.fail('Exception was thrown');
+	}
```
sindresorhus: Use `t.notThrows` rather than try/catch:

```js
const p = got(`${s.url}/304`)
await t.notThrows(p);
const response = await p;
t.is(response.statusCode, 304);
t.is(response.body, '');
```

test/http.js
```diff
@@ -56,6 +61,16 @@ test('error with code', async t => {
 	}
 });
 
+test('304 doesn\'t throw', async t => {
```
sindresorhus: > status code 304 doesn't throw

floatdrop:
@sindresorhus yeah, it can be removed now. Tests were failing without it on 0.10/0.12 I think.

lukechilds:
Ok, I've made those changes.

@sindresorhus is the wording/position in the readme ok?

sindresorhus:
Yup. Looks good. Thanks again :)

PR #253, opened 2016-12-20
use `safe-buffer`

ReklatsMasters:
[`safe-buffer`](https://www.npmjs.com/package/safe-buffer) implements `Buffer.from` (` Buffer.alloc`, ... ) where they are missing.


floatdrop:
@ReklatsMasters sweet! Can you add `Node 4.4.4` in `.travis.yml` to verify missing method case?

PR #254, opened 2016-12-22
Strengthen expect undefined tests

AlexTes:
Since we're checking for an undefined property, it becomes relevant to
check the rest of the object does look the way we expect. If the returned body object breaks, through parsing problems or otherwise, this test could break but pass.

I ran into this working on a larger PR coming up :3.

* I don't care for the comment I added. If you feel the reason is clear, I could remove it.
* If you feel there's a way to address this issue in a cleaner way, I'd happily implement it.
* If you feel breaking something like parsing but this test still passing is not something these specific tests should care about because of the assumption another test somewhere _should_ break too. I can get behind that, and you can close the PR. I like the idea of fixing tests you'd expect to break but don't, and making them break. On the other hand, you shouldn't test _all_ of the dependencies every test relies on. Your call!

### DIFF
--- a/test/post.js
+++ b/test/post.js
@@ -88,6 +88,9 @@ test('content-length header with Stream body', async t => {
 		body: intoStream(['wow']),
 		json: true
 	});
+	// since we're checking for undefined, let's make sure the rest of the
+	// reponse is there by checking another prop we expect to be there
+	t.is(body['transfer-encoding'], 'chunked', 'likely failed to get headers');
 	t.is(body['content-length'], undefined);
 });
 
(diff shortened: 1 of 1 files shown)


AlexTes:
bitmoji

floatdrop:
> I don't care for the comment I added. If you feel the reason is clear, I could remove it.

@AlexTes Yeah, tests should be descriptive by themselves. 👍

AlexTes:
@floatdrop Added two words to the descriptive error message. Removed the comment. I feel this would also work.

Issue #174, opened 2016-02-19
JSON option when posting data


paglias:
Would it be possible to use the `json: true` option when posting data to set `Content-Type` to `application/json` and stringify the body with `JSON.stringify`?

floatdrop:
@paglias it is possible, but `json` option is responsible only for toggle parsing response from server and it will be kinda major change to make it stringify the body for request also.

spencerhakim:
It would be a major change, but not necessarily a bad one. At least having the option to enable it would be nice. Like, if `json` is an object instead of just `true` seems like a reasonable compromise.

sindresorhus:
@spencerhakim Can you provide a code example of how it would look like?

sindresorhus:
Reopening, but we're not likely to do another major in the near future.

spencerhakim:
Current way:
```js
var options = {
  json: true, // if truthy, parse *response* as JSON
  headers: {
    'Content-type': 'application/json'
  },
  body: JSON.stringify({ // have to manually stringify object in body
    name: 'whatever',
    category: 'toy',
    description: 'Hello World'
  })
};

request.post(`http://example.com/api/createThing`, options)
  .then((resp) => {
    // ...
  });
```

Possible new way (should be functionally equivalent to the above):
```js
var options = {
  json: { // typeof json === 'object', not 'boolean', so send body as JSON instead of form-encoded
    name: 'whatever',
    category: 'toy',
    description: 'Hello World'
  }
};

request.post(`http://example.com/api/createThing`, options)
  .then((resp) => {
    // ...
  });
```

Coming from `request`, I was surprised that `got` didn't follow the same semantics for handling JSON or offer any way of automatically serializing sent objects as JSON. This approach should be backwards compatible with current usage, so I don't think it requires a major version bump.

sindresorhus:
Wouldn't it make more sense to pass the object literal in `body`?

spencerhakim:
Yes, but that approach would likely require a major version bump.

bishtawi:
@sindresorhus What about making the `json` attribute an object that takes `req` and `res` as parameters. Setting `req` to `true` means the request body should be JSON encoded, setting `res` to `true` means the response body should be JSON parsed.

[...]

PR #264, opened 2017-01-22
Solving the json option confusion by replacing it with a parse option

AlexTes:
# Solving the JSON option confusion
## Preamble
You're looking at my first significant PR hooray 🎉! Got is awesome and inspired me to submit it. I’ve put a lot of thought into this PR and its code; nevertheless, the project comes first. Feel free to be completely honest. Also quite excited after spending hours creating and rebasing. Most commits have an elaborate description, I hope they will be a joy to read.

On to the PR!

Got wants to support parsing of responses and sending of encoded bodies. The current JSON option achieves one of these but seems to confuse people. After *a lot* of comparing libraries and thinking here’s what I believe we can do!

Summarizing from #174 and my thoughts here’s why the JSON option currently confuses:
* It’s the only option that helps parse although you can parse a response many ways.
* Some popular libraries parse based on headers by default.
* Some popular libraries use JSON as the body encoding by default or have a shortcut for it.

I think from #174 the easiest solution is to bundle all parsing under one parse option and drop the confusing JSON option. To my thinking, this would lead to better expected behaviour (and is a little more flexible going forward).

But that’s not where it ends!

I'm thinking of a next *PR* that would add some nice default parsing based on the Content-Type header (that's what that’s for after all). Seems like a nice thing to do 😄. Might also make Got too complex, let me know what you think!

I also created #265 because - as I hinted earlier - I think the body encoding default is confusing people. I have added a few suggestions to ease use there too.


AlexTes:
Just noticed there was a misguided commit in there. Dropped it and rewrote history to improve the PR. I think we can improve the test that confused me. The details will be in the PR under this comment.

sindresorhus:
Very good PR and reasoning @AlexTes. Sorry not getting to this sooner.

I'm not entirely sold on the change itself though. The reason we added a `json` option is that JSON is the 95% use-case and we wanted to optimize for that. I've honestly never used any other encoding than plain and JSON. I would rather want to fix any confusion with the JSON option than replacing it with something more flexible.

AlexTes:
Fair enough @sindresorhus. It's 😢 to close my first PR but I'll get over it 😉, I knew this chance as there from the beginning when I chose not to discuss my solution first. 
It was a fun and good way to familiarize myself with got.
Thanks for looking over the long PR and helping me in merging many other PRs by now.

sindresorhus:
Yeah, always hard to close PRs that had a lot of work gone into them, but that's why opening an issue first is often beneficial.

PR #267, opened 2017-01-23
Make (more) sure the error we get is the one we expect

AlexTes:
Since we set the statusCode to 500 in the changed test, our code triggers a HTTPError. Regardless of the parsing conditional's outcome that may or may not throw. Because of this, the test would pass even if the code does not throw a ParseError. This change makes sure the error is the got.ParseError we expect. By no means perfect but good enough to save me and hopefully others from confusion.

It is a minor change. Hopefully, you too feel this would improve, not complicate the code 😄 .

### DIFF
--- a/test/json.js
+++ b/test/json.js
@@ -74,13 +74,8 @@ test('catches errors on invalid non-200 responses', async t => {
 });
 
 test('should have statusCode in err', async t => {
-	try {
-		await got(`${s.url}/non200-invalid`, {json: true});
-		t.fail('Exception was not thrown');
-	} catch (err) {
-		t.true(err instanceof got.ParseError);
-		t.is(err.statusCode, 500);
-	}
+	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}), got.ParseError);
+	t.is(err.statusCode, 500);
 });
 
 test.after('cleanup', async () => {
--- a/test/json.js
+++ b/test/json.js
@@ -78,6 +78,7 @@ test('should have statusCode in err', async t => {
 		await got(`${s.url}/non200-invalid`, {json: true});
 		t.fail('Exception was not thrown');
 	} catch (err) {
+		t.true(err instanceof got.ParseError);
 		t.is(err.statusCode, 500);
 	}
 });


sindresorhus:
Always happy to improve the tests. Thanks for the PR. I think it would be better to switch the whole test to use `t.throws()` instead though:

```js
test('should have statusCode in err', async t => {
	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}), got.ParseError);
	t.is(err.statusCode, 500);
});
```

AlexTes:
@sindresorhus hi Sindre!
Me too! Should've thought of that. I'll update this one and then have a look around for others that could be improved that way 😺 .

AlexTes:
I had no idea Ava's `.throws` could take an error to compare to, that's perfect!
Thanks for the suggestion.
If I understand ava's blowing up correctly we shouldn't wrap the main got function in an anonymous function because it's caught at some point and returned in the form of a rejected promise.

Sidenote: 
breaking the test causes the following message:
```
json › should have statusCode in err
  Cannot read property 'host' of undefined`
```
That seems less than ideal. I'm not sure why that happens or how to improve it. I'll do some digging  ^_^.

sindresorhus:
> If I understand ava's blowing up correctly we shouldn't wrap the main got function in an anonymous function because it's caught at some point and returned in the form of a rejected promise.

Yes, no need, since we're passing in a promise, which is handled by `t.throws`.

> breaking the test causes the following message:

How are you breaking the test?

AlexTes:
@sindresorhus
```js
test('should have statusCode in err', async t => {
	const err = await t.throws(got(`${s.url}/non200-invalid`), got.ParseError);
	t.is(err.statusCode, 500);
});
```
I'm having a very hard time tracing the cause of the AssertionError `.throws` seems to throw.

sindresorhus:
@AlexTes Hmm. Weird. See: https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-chrome-devtools.md for debugging.

AlexTes:
@sindresorhus I know ava pretty well. She's amazing! Was already messing around with that but I couldn't seem to break T_T.

Regardless I almost got it. No need for you to spend time on this. Go be awesome!
If you're interested it's going wrong over at:
```js
function stdError(error, opts) {
	if (error.code !== undefined) {
		this.code = error.code;
	}

	Object.assign(this, {
		message: error.message,
		host: opts.host,
		hostname: opts.hostname,
		method: opts.method,
		path: opts.path
	});
}

got.HTTPError = createErrorClass('HTTPError', function (statusCode, opts) {
	stdError.call(this, {}, opts);
	this.statusCode = statusCode;
	this.message = `Response code ${this.statusCode} (${this.statusMessage})`;
});
```

AlexTes:
I finally understand @sindresorhus 😓 . The check is still valid. After adding some defaults a `ParseError` no longer blows up with access of undefined and ava correctly checks we are looking at an instance of a `ParseError`. 

Now to the silly bit. Ava / the assertion lib manages to hijack the error's name, `ParseError` and turns it into an `AssertionError`. The rest however, the `createErrorClass` logic overwrites. In other words, when the test fails we end up with a very strange error message (in verbose mode):
```
  ✖ json › should have statusCode in err Response code 500 (Internal Server Error)

  1 test failed [23:32:50]


  1. json › should have statusCode in err
  AssertionError: Response code 500 (Internal Server Error)
```

This is error created by the `createErrorClass` logic in my previous comment.
Catching and rethrowing something makes little sense.
We could go back to the code I originally PR'd.
We could leave it like this. The code is the most clear. It does go downhill fast when the test fails.. which is kinda what it should do. Ava seems to prefer the error's message property to what I can optionally pass so that doesn't help. I would have to add a commit that sets defaults for the `ParseError` constructor since for a reason I'm not entirely sure about `.throws` ends up invoking the constructor passed to compare to.

I have a bit of a headache, but it was a good lesson! Hope to keep contributing 😄 .

sindresorhus:
> Submit a PR to Ava! When .throws catches an error that is not judged valid by optional second arg we use the optional message from the third arg if there is any.

This ;) Could you open an issue (or PR) on AVA about this?

sindresorhus:
Oh wow. Good debugging! :)

sindresorhus:
> Hope to keep contributing 😄 .

Please do! I'm in https://gitter.im/sindresorhus/meta if you have any questions. Sorry about the delay. I was busy in real life.

AlexTes:
> This ;) Could you open an issue (or PR) on AVA about this?

On it! I'll try and open a solid PR 😁.

> Oh wow. Good debugging! :)

ty ty 😊 

> Please do! I'm in [...]

Awesome! See you there.

PR #273, opened 2017-02-14
Make ParseError checking more clear

AlexTes:
JSON test 'should have statusCode in err' requires an instance of
ParseError to compare to but has no `opts` or `data` to pass. This commit
makes those arguments optional allowing for empty ParseError
construction.

### DIFF
--- a/test/json.js
+++ b/test/json.js
@@ -74,7 +74,8 @@ test('catches errors on invalid non-200 responses', async t => {
 });
 
 test('should have statusCode in err', async t => {
-	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}), got.ParseError);
+	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}));
+	t.is(err.constructor, got.ParseError);
 	t.is(err.statusCode, 500);
 });
 


floatdrop:
Why not pass those dummy values to ParseError in test?

AlexTes:
That would be much nicer @floatdrop!
I'd much rather have the test explicitly use less-sensible arguments than making the actual code more lenient by making the less-sensible arguments defaults.

I didn't do this straight off because I didn't know it was possible. Looking at the ava docs for `.throws`: 
> error can be an error constructor, error message, regex matched against the error message, or validation function.

Regardless, the following works:
```js
	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}), 
		got.ParseError(new Error(), 500, {}, ''));
```

But to my understanding that's passing an instance, not any of the above, and to make things worse, once I comply with the linting rules and use the 'new' keyword things stop working and I get confused.

Started digging around in the [node source](https://github.com/nodejs/node/blob/master/lib/assert.js#L330), and played around with argument binding, but neither have worked out yet and sadly I don't have more time this morning. I'll try and continue this soon!

Thanks for the suggestion @floatdrop.

sindresorhus:
Maybe this would work, or if not, we could always just check the individual properties:

```js
const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}));
t.deepEqual(err, got.ParseError(new Error(), 500, {}, ''));
```

AlexTes:
Figured out why it worked without the `new`, because the custom error constructor returns `undefined` in that case. Which if you're interested fails [this check](https://github.com/nodejs/node/blob/master/lib/assert.js#L362) which means node's assert throws nothing, and ava returns the result obtained [here](https://github.com/avajs/ava/blob/master/lib/assert.js#L107) which is the HTTPError which also has a property `statusCode` set to 500.

Now that I understand some of the internals of `.throws` in NodeJS and AVA that AVA PR we talked about in #267 should quickly follow 😁 .

Errors are notoriously inaccessible, you can't iterate them, I tried anyway, but deepEqual is out.

I really like @sindresorhus 's suggestion to just check props then, the way I see it that means we have three okay options. Every error should have a constructor, a message, and a name. That would respectively make the output when some non ParseError was thrown:
Using constructor:
error_constructor
Using message:
error_message
Using name:
error_name

Sorry for taking so long to get back to you two.

test/json.js
```diff
@@ -74,7 +74,8 @@ test('catches errors on invalid non-200 responses', async t => {
 });
 
 test('should have statusCode in err', async t => {
-	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}), got.ParseError);
+	const err = await t.throws(got(`${s.url}/non200-invalid`, {json: true}));
+	t.is(err.constructor, new got.ParseError(new Error(), 500, {}, '').constructor);
```
sindresorhus: Why not just `t.is(err.constructor, got.ParseError);`?

AlexTes: Fair question. Too long ago to remember what the issue was. Fairly sure I tried. But maybe I wrongly assumed `t.is` would run into the same problem as `t.throws`. I'll try tonight!

AlexTes: You were super right. That works fine. Updated the PR.

sindresorhus:
👍

PR #277, opened 2017-03-08
Add protocol to stdError

### LINKED ISSUE #249: Add protocol to error messages
https://github.com/sindresorhus/got/blob/master/index.js#L339 has a bunch of request settings copied into the error object. The protocol would also be helpful.

AlexTes:
Adds the protocol to the stdError object got uses for all its errors.
Closes #249.

(These are _so_ much fun to do. If only my lunch break was longer.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -334,7 +334,8 @@ function stdError(error, opts) {
 		host: opts.host,
 		hostname: opts.hostname,
 		method: opts.method,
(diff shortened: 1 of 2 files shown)


AlexTes:
Fixed accidently basing this branch on my other branch instead of master. Sorry 🙈.

PR #283, opened 2017-03-15
Add response data to error

utter-step:
Some API services return additional info on errors (e.g. names of fields with errors), it would be nice to be able to get this info, when handling error.


utter-step:
Found `error.request`, sorry.)

PR #284, opened 2017-03-23
Add `cache` option

### LINKED ISSUE #281: Respect and handle cache headers
Would be very handy if Got had builtin support for response caching. Most APIs you interact with implement either `ETag` or `Last-Modified` headers for caching and it's annoying to implement manually, but usually required for efficiency and rate limiting reasons. It could be a Got wrapper, but in reality most requests will need this, so I think it should be in core. We should not be opinionated in how you cache the request however. We could simply say that it supports any module that exposes `.set()`, `.get()`, `.has()` methods. Most will probably want to use a LRU cache for in-memory storage, or Redis higher scale requirements.

Relevant GitHub docs: https://developer.github.com/v3/#rate-limiting

Prior art:

- https://github.com/export-mike/f-etag/blob/master/src/index.js
- https://github.com/gitterHQ/request-http-cache
- https://github.com/kaelzhang/node-modified
- https://github.com/ForbesLindesay/http-basic/blob/febede32150b99874f2c327a19b206b73cdec377/lib/cache-utils.js
- https://github.com/spalger/http-asset
- https://github.com/developit/request-easy-cache (only normal caching)

---

// @floatdrop @AlexTes  @lukechilds @julien-f @reconbot

lukechilds:
Resolves #281 

To keep track of what's implemented:

- [x] Allow user to pass in a cache store (copy `Map()` API)
- [x] Cache responses that are cacheable
- [x] Return cached responses immediately if still fresh
- [x] If expired send `If-Modified-Since`/`If-None-Match` header
- [x] Handle potential 304 response
- [x] Tests
- [x] Refactor into external module
- [x] Write some cache adapters
- [x] Documentation

Ok I hacked this together last night, I'm aware there are a few things that need tidying up / potentially unhandled errors etc. Just submitting this PR for some feedback.

I'm caching the url/headers/body of all responses that satisfy:

```js
(opts.cache && statusCode === 200 && req.method === 'GET' && (expired(response.headers) === false || 'etag' in response.headers))
```

This is working, I can view them in the cache. The issue is to serve them up I need to create a Node.js response (or similar) object to be passed through Got's `asPromise()`/`asStream()` fns and to make sure all the usual response API methods are available on the returned object. I was hoping `http-response-object` was gonna do the job here but it looks like it only implements a basic subset of the Node.js response object.

When I emit the response-like object I get:

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: inputStream.once is not a function
```

I think this is coming from `getStream()` because `http-response-object` doesn't implement some stream stuff from the Node.js response object.

Any thoughts? I'm kind of out of my depth here in regards to streams and the native response object. Is there is simpler way to do this? Or maybe a straight forward way to create a fake request object?

// @sindresorhus @reconbot @AlexTes


lukechilds:
Got it working, `http-response-object` doesn't implement the Readable Stream interface. Works with this PR: https://github.com/ForbesLindesay/http-response-object/pull/4

Is `http-response-object` "good enough" to return instead of a native response? If not this may need to be a Got wrapper that just exposes url/headers/body and none of the Node.js response stuff.

lukechilds:
Ok, I've done some digging and a Node.js response is an instance of [http.IncomingMessage](https://nodejs.org/api/http.html#http_class_http_incomingmessage). So to make the cached version compatible with the Node.js response there's quite a few things we need to support as well as some methods/properties that we can't support on the cached response (e.g `response.socket`).

We'll need to either add support to all these things in `http-response-object` or if @ForbesLindesay thinks this is beyond the scope of his module I'm happy to create a new one.

There is also the issue that we'll need to manually keep the cached response up to date with `http.IncomingMessage`. For example if someone uses a new version of Node.js that has a new method on `response`, they'll be able to use that new method on normal responses but it won't exist on cached responses until it's implemented in our response package.

This seems like kind of a hacky solution, maybe it would be better to limit the cached responses to only basic data, although this is probably not gonna be obvious to people unless they read the docs. Or alternatively just do this as a wrapper that never exposes the Node.js response.

I'll wait for feedback before going any further with this.

sindresorhus:
Maybe https://github.com/pornel/http-cache-semantics might be of some use. It seems to have a pretty extensive coverage of all caching possibilities and quirks in HTTP. Even if it's just for inspiration.

sindresorhus:
> maybe it would be better to limit the cached responses to only basic data

Yes, let's limit it to the basic data. statusCode, headers, body, url.

> I'm happy to create a new one.

Seems like you need to do that regardless as you have yet to get a response on your PR.

index.js
```diff
@@ -58,6 +61,20 @@ function requestAsEventEmitter(opts) {
 				response.url = redirectUrl || requestUrl;
 				response.requestUrl = requestUrl;
 
+				if (opts.cache && statusCode === 200 && req.method === 'GET' && (expired(response.headers) === false || 'etag' in response.headers)) {
+					const stream = opts.encoding === null ? getStream.buffer(response) : getStream(response, opts);
```
```js
const encoding = opts.encoding === null ? 'buffer' : opts.encoding;
const stream = getStream(stream, {encoding});
```

lukechilds: 👌

lukechilds:
`http-cache-semantics` looks perfect 😍. I wish I'd seen that before writing `expired`...

lukechilds:
Just an idea but `http-response-object` gives us a TTL so we know how long a cache item can be stored for. LevelUP is just get/set with no concept of a TTL. If it's stored, it's stored forever. This isn't a massive issue, we can manually delete items from the cache if we see they're stale on a future request or if LevelUP is backed by a LRU store then it'll take care of itself.

However this functionality is being written in a way that will allow someone to use any cache store with Got and I think TTL is useful information to pass through so users have the option of using it. What about if we change the `cache.put` signature to `cache.put(key, val, ttl, cb)` rather than the current LevelUP signature of `cache.put(key, val, cb)`.

Then we could provide a LevelUP cache plugin for Got:

```js
const got = require('got');
const gotCacheLevelup = require('got-cache-levelup');
const db = gotCacheLevelup('mydb', { db: require('redisdown'), host: 'localhost', port: 6379 });

got('foo.com', {cache: db});
```

Also, I was just using callbacks so it works OOB with a LevelUP db instance. If are setting our own spec for cache "plugins" then we could require a Promise interface which would be a bit neater.

Thoughts?

sindresorhus:
Instead of adding an additional parameter to `.set()`. Couldn't we just create additional keys for meta data. So if there's a key called `https://sindresorhus.com`, we could also create a `__ttl__$__https://sindresorhus.com` (Or something), with the TTL value. I would really prefer to keep a simple glue API, so user could just input for example `Map()` directly without anything extra. Alternatively we could store the TTL in the data, so instead of putting the data `{foo: true}`, we put `{ttl: 123123, data: {foo: true}`.

> then we could require a Promise interface which would be a bit neater.

We can just run the input through `Promise.resolve()`, so the user can submit any value or a Promise, and it's awaited if it's a Promise.

lukechilds:
We can still pass the TTL as the third argument to set and match the `Map()` API. `map.set` will just ignore the TTL:

```js
const map = new Map();
map.set('foo', 'bar', 100);
map.get('foo');
// "bar"
```

So `got('foo.com', {cache: new Map()});` would work fine. I'm not too sure what you mean regarding creating a new cache entry for the TTL, I don't think I explained myself very well. I mean it's useful for other people to use the TTL in storage adapters, not that it's useful to store.

e.g: If `some-cache-store` had native TTL functionality to clear old values we could do:

```js
const cache = require('some-cache-store');

const gotCachePlugin = {
  set: (key, val, ttl) => cache.put(key, val, ttl),
  get: key => cache.get(key),
};

module.exports = gotCachePlugin;
```

> We can just run the input through Promise.resolve(), so the user can submit any value or a Promise, and it's awaited if it's a Promise.

Sounds good 👌

sindresorhus:
> I'm not too sure what you mean regarding creating a new cache entry for the TTL, I don't think I explained myself very well. I mean it's useful for other people to use the TTL in storage adapters, not that it's useful to store.

I was proposing we just support TTL natively, so users can use any kind of "dumb" storage and get TTL support without doing anything extra.

sindresorhus:
I guess I'm misunderstanding how this all works. I assumed we get a TTL from the server, save it, and the next time the same resource is requested, we check the cache, check whether the TTL has expired or not, and than either uses the cache, or deletes the cache entry and TTL, and refetches from the server.

lukechilds:
Nah, that's not quite how it works. We make a request, save the response along with the cache policy generated by `http-response-object`, then the next time the resource is requested we check the cache, check whether the cache policy is still valid, then either use the cached response or refetch. We don't need to store the TTL, everything required to check that the cached response is fresh is handled by `policy.satisfiesWithoutRevalidation()` from `http-response-object`.

The TTL would be useful so the DB can clear out stale cache entries ASAP rather than waiting for them to be requested again before checking.

e.g If you make lots of requests to different URLs that you won't request again, you'll end up with loads of entries filling up your cache.

sindresorhus:
Ok. Makes sense. Thanks for elaborating :)

lukechilds:
What do you think would be the best way to handle binary responses? I'm struggling to get my head round it.

The simple solution would be b64 encode => cache => b64 decode. But that seems like it could have pretty major performance implications. If you're using an image API for example then reading the cache is gonna block the event loop while decoding potentially MBs of data.

Another solution would be to store as a hex string in the cache. That could then be loaded directly into a buffer but it's gonna take up loads of space in the cache.

Any ideas?

sindresorhus:
I think that should be up to the consumer. If they choose to get a `buffer` encoding, it's up to them to properly handle it. With in-memory caching, it's easy as you can just save the buffer directly. When it's saved to disk, however, it's a little bit harder as you need a data format/database that supports blobs, or serialize it yourself.

lukechilds:
Oh yeah, of course 👌. For some reason I was thinking we needed to serialize all the data we're caching to a JSON string, but that's just specific to the LevelUP implementation.

lukechilds:
Is it safe for`getStream()` to be called twice on cached requests? Once to [get a buffer](https://github.com/lukechilds/got/blob/7d4365ec2569a4a5adbcca03e62bba2605f089dd/index.js#L67) to cache and then again [inside `asPromise()`](https://github.com/lukechilds/got/blob/7d4365ec2569a4a5adbcca03e62bba2605f089dd/index.js#L152) to be returned to the user?

It seems to be working ok but I'm not sure if maybe reading from the same stream twice could cause problems. Maybe I should be cloning the stream and using one for the cache and another to pass to `asPromise()`/`asStream()` like this: https://github.com/nodejs/readable-stream/issues/202

sindresorhus:
Streams can only be read from once. If it's works now it's probably because you're sending something small which is still buffered.

Instead, you should do the caching later on in the code. For example, we wouldn't want to cache an error response. I would do the caching here: https://github.com/lukechilds/got/blob/7d4365ec2569a4a5adbcca03e62bba2605f089dd/index.js#L173 And hook into the pipeline here: https://github.com/lukechilds/got/blob/7d4365ec2569a4a5adbcca03e62bba2605f089dd/index.js#L227

lukechilds:
I've moved the caching around like you suggested, however after taking another look at this I think maybe it was better how it was before. I think (I might be wrong) it's ok to read a stream multiple times, as long as your read at the same time, not after it's already started being read.

e.g this is fine: 

```js
const stream = got.stream('todomvc.com');
getStream(stream).then(txt => console.log(txt.length));
getStream(stream).then(txt => console.log(txt.length));
// 254806
// 254806
```
This isn't:

```js
const stream = got.stream('todomvc.com');
getStream(stream).then(txt => {
	console.log(txt.length)
	getStream(stream).then(txt => console.log(txt.length));
});
// 254806
// 0
```

The previous method I had wasn't doing anything else with the stream inside the `then()`. The other read was happening outside. This had the benefit of having all the cache logic once inside `get()` rather than checking twice inside both `asPromise()` and `asStream()`. It also meant that we would only ever try to cache responses from `get()` and not `getFromCache()`. Now the caching is inside `asPromise()`/`asStream()` I need to check whether the response came from  `get()`/`getFromCache()` and make sure we don't try and cache responses that were already served from the cache.

Does this make sense? Do you think it was ok how it was previously or is there a reason it needs to be like this?

sindresorhus:
Ok, I looked more into this. You can get away with multiple listeners if you add them in the same tick. Since the stream will start pumping right when it has a subscriber, the second one might miss something if it's added in a different tick. So, if you can add two listeners at once, the it should be ok. Alternatively, you could use a PassThrough stream to duplicate it: http://stackoverflow.com/a/19561718/64949

lukechilds:
Yep, I think that's the best solution. I don't think a PassThrough stream would work well in this scenario as it's not just a stream but an instance of http.IncomingMessage. So we can't just return a new stream, we'd need to copy all the relevant properties from the response to the duplicated stream which would be kind of messy.

lukechilds:
Ok, this seems to be pretty much sorted and working.

- All cacheable responses are cached.
- Cache API is compatible with Map so `new Map()` works OOB.
- Fresh cache entries are served directly from the cache.
- Stale cache entries with either Last-Modified or ETag headers are revalidated.
- 304 responses from the revalidation requests will use the cached body (user doesn't have to manually handle 304)
- Revalidation responses are re-cached (headers may have changed)
- Uses [http-cache-semantics](https://github.com/pornel/http-cache-semantics) to check responses are cacheable/fresh/revalidated so should conform fully to HTTP RFC 7234

You can view `test/cache.js` for an overview of the new behaviour.

The revalidation code (https://github.com/sindresorhus/got/pull/284/commits/07a403c1528abda166275b626a89c3e3eee90f8e) is a bit messy, I'm working on refactoring that into something more readable. Also there are a few other things I wanna tidy up and a couple of edge cases I wanna test.

Is there any functionality you think I've missed or have I covered everything?

lukechilds:
Ok, after having a think about this it occurred to me that I'm going about this the wrong way. Rather than trying to bake all the caching logic directly into Got without breaking any existing Got functionality, it would be much simpler to just extract what I've got so far into a separate module that wraps `http.request` with caching support. Then, rather than bloating Got with about 33% extra code we can just require the cache module and wrap `http.request` with it if `opts.cache` is set.

It'll return an event emitter and pass all request events through if a request needs to be made for the resource. That means it will automatically work with all of Got's existing functionality.

So just to keep you in the loop, there won't be any activity on this PR for a while, don't worry though, I'm still working on it, I'll just be focusing on getting the standalone cache module working. Once the cache module's implemented it should just be +~3 line change for all the same functionality we've discussed. 🎉

sindresorhus:
@lukechilds Just wanted to say that I'm totally 👍 on this approach.

lukechilds:
@sindresorhus Great 🎉 

Btw, in my last message I said:

> It'll return an event emitter and pass all request events through if a request needs to be made for the resource. That means it will automatically work with all of Got's existing functionality.

However I think it would be better to leave the request events and allow listening to them via a single `request` event for the event listeners to be attached. e.g:

```js
cacheableRequest(http.request, opts, cb)
  .on('error', handleCacheError)
  .on('request', req => {
    req.on('error', handleRequestError);
    req.end();
  });
```

Does that seem like the better solution to you?

test/cache.js
```diff
@@ -0,0 +1,197 @@
...
+			res.statusCode = 304;
+			responseBody = null;
+		}
+
+		res.end(responseBody);
+	});
+
+	let cacheThenNoStoreIndex = 0;
+	s.on('/cache-then-no-store-on-revalidate', (req, res) => {
+		const cc = cacheThenNoStoreIndex === 0 ? 'public, max-age=0' : 'public, no-cache, no-store';
```
vadimdemedes: I'd rename this to `cacheControl`.

test/cache.js
```diff
@@ -0,0 +1,197 @@
...
+	});
+
+	await s.listen(s.port);
+});
+
+test('Non cacheable responses are not cached', async t => {
+	const endpoint = '/no-store';
+	const cache = new Map();
+
+	const firstResponseInt = parseInt((await got(s.url + endpoint, {cache})).body, 10);
```
vadimdemedes: I think xo had a rule favoring `Number(x)` over `parseInt(x, 10)`.

test/cache.js
```diff
@@ -0,0 +1,197 @@
...
+test('Binary responses are cached', async t => {
+	const endpoint = '/cache';
+	const cache = new Map();
+	const encoding = null;
+
+	const firstResponse = await got(s.url + endpoint, {cache, encoding});
+	const secondResponse = await got(s.url + endpoint, {cache, encoding});
+
+	t.is(cache.size, 1);
+	t.true(firstResponse.body instanceof Buffer);
```
vadimdemedes: => `Buffer.isBuffer(firstResponse.body)`

test/cache.js
```diff
@@ -0,0 +1,197 @@
...
+
+test('TTL is passed to cache', async t => {
+	const endpoint = '/cache';
+	const store = new Map();
+	const cache = {
+		get: store.get.bind(store),
+		set: (key, val, ttl) => {
+			t.true(typeof ttl === 'number');
+			t.true(ttl > 0);
+			return store.set(key, val, ttl);
```
vadimdemedes: Since `store` isn't actually needed, perhaps the following would be simpler?

```js
const cache = {
  get: () => {},
  delete: () => {},
  set: (key, val, ttl) => {
    // your set() code
  }
};
```

test/cache.js
```diff
@@ -0,0 +1,197 @@
...
+	t.is(firstResponse.body, secondResponse.body);
+});
+
+test('Stale cache entries that can\'t be revalidate are deleted from cache', async t => {
+	const endpoint = '/cache-then-no-store-on-revalidate';
+	const cache = new Map();
+
+	const firstResponse = await got(s.url + endpoint, {cache});
+	t.is(cache.size, 1);
+	const secondResponse = await got(s.url + endpoint, {cache, log: true});
```
vadimdemedes: Could you add a newline above too?

lukechilds:
@vadimdemedes thanks for reviewing, a lot of these tests are actually duplicated in `cached-request` so I need to go though and remove some of them from Got.

sindresorhus:
> Does that seem like the better solution to you?

It's unclear to me what the difference is?

lukechilds:
@sindresorhus so originally I was thinking `cacheableRequest` should have the same API as `http.request` so they would be completely interchangeable, just a drop in upgrade. e.g:

```js
cacheableRequest(http.request, opts, cb)
    .on('error', handleRequestError);
    .end();
```

However I realised afterwards that was going to be pretty messy to implement because we don't know if we actually need to make a request until after `cacheableRequest` has returned. So we'd have to manually patch all methods events over to the emitter we've already returned if/when we have the request object. 😐

So instead I decided to just return the request object inside a request event. Then you can do all your normal request stuff in the `request` event. And handle cache specific event on the main event emitter:

```js
cacheableRequest(http.request, opts, cb)
  .on('error', handleError)
  .on('request', req => {
    req.on('error', handleRequestError);
    req.end();
  });
```

sindresorhus:
@lukechilds Ah ok, got it. Makes sense.

sindresorhus:
@lukechilds What's the status on this? Could you fix the merge conflict?

lukechilds:
Current status is that I'm finishing off storage adapters (in a separate repo) and then writing docs. If it's all the same to you I'll focus on getting that done first and clean up the merge conflict after. Now that everything's handled in external modules the changes to the Got codebase are minimal.

lukechilds:
@sindresorhus current status is that I'm finishing off storage adapters (in a separate repo) and then writing docs. If it's all the same to you I'll focus on getting that done first and clean up the merge conflict after. Now that everything's handled in external modules the changes to the Got codebase are minimal.

lukechilds:
[Keyv](https://github.com/lukechilds/keyv) and [`cacheable-request`](https://github.com/lukechilds/cacheable-request) are now stable and I've updated this PR to use them and fixed all merge conflicts.

@sindresorhus I've built a load of error handling into Keyv/`cacheable-request` but not sure how to expose it through Got.

Should I create a `got.CacheError` class? Currently `cacheable-request` emits the raw error straight from the DB client so they could have different properties depending on which storage adapter is used.

lukechilds:
Would this be enough:

```js
got.CacheError = class extends StdError {
	constructor(error, opts) {
		super(error.message, error, opts);
		this.name = 'CacheError';
	}
};
```

Or should I also attach the underlying DB error, with something like:

```js
got.CacheError = class extends StdError {
	constructor(error, opts) {
		super(error.message, error, opts);
		this.name = 'CacheError';
    this.error = error;
	}
};
```

Seems the DB error would be useful to have but accessing it as `err.error` just seems wrong.

jotto:
Echoing the comment here: https://github.com/sindresorhus/got/issues/281#issuecomment-288059738

> I'd prefer the second, needs more documentation but keeps got tiny.

I'd like to gently point out that `got` is currently 0.16M, but if `cacheable-request` is added, it gains 0.39M (0.53M total). My favorite thing about got is how tiny it is.

lukechilds:
@jotto How are you calculating Got's size?

In a fresh directory I get:

```shell
$ npm install got && du -hd 0 node_modules
584K	node_modules
```

That's without `cacheable-request`.

Thanks for bringing this up though, `cacheable-request` is currently 708K. There's currently no `.npmignore` used in `cacheable-request` or any other libs modules I've published that it depends on. I'll fix that which should make a significant reduction in size.

jotto:
@lukechilds I used https://github.com/siddharthkp/cost-of-modules

Size of npm package is, historically not an issue but suddenly more important since "serverless" AKA function as a service (AWS Lambda etc.) has some size constraints.

lukechilds:
Is this ok to be merged?

index.js
```diff
@@ -190,62 +192,66 @@ function requestAsEventEmitter(opts) {
...
+						const headersSize = Buffer.byteLength(req._header);
+						uploaded = req.connection.bytesWritten - headersSize;
 
-					// Don't emit events with unchanged progress and
-					// prevent last event from being emitted, because
-					// it's emitted when `response` is emitted
-					if (uploaded === lastUploaded || uploaded === uploadBodySize) {
-						return;
-					}
+            // Prevent the known issue of `bytesWritten` being larger than body size
```
sindresorhus: These comments are not aligned with the below code.

readme.md
```diff
@@ -170,6 +171,13 @@ Decompress the response automatically.
 
 If this is disabled, a compressed response is returned as a `Buffer`. This may be useful if you want to handle decompression yourself or stream the raw compressed data.
 
+###### cache
+
+Type: `object`<br>
+Default: `false`
+
+Cache adapter instance for storing cached data.
```
sindresorhus: Link to the cache section here.

readme.md
```diff
@@ -170,6 +171,13 @@ Decompress the response automatically.
 
 If this is disabled, a compressed response is returned as a `Buffer`. This may be useful if you want to handle decompression yourself or stream the raw compressed data.
 
+###### cache
+
+Type: `object`<br>
```
sindresorhus: object => Object

readme.md
```diff
@@ -253,6 +261,10 @@ Each error contains (if available) `statusCode`, `statusMessage`, `host`, `hostn
 
 In Promise mode, the `response` is attached to the error.
 
+#### got.CacheError
+
+When a cache method fails.
```
sindresorhus: Can you include *when* it could fail?

readme.md
```diff
@@ -282,6 +294,57 @@ When given an unsupported protocol.
...
+
+You can use the JavaScript `Map` type as an in memory cache:
+
+```js
+const got = require('got');
+const map = new Map();
+
+(async () => {
+    let response = await got('todomvc.com', {cache: map});
+    console.log(response.fromCache);
```
sindresorhus: `. fromCache` and any other extra properties on the response needs to be documented.

readme.md
```diff
@@ -282,6 +294,57 @@ When given an unsupported protocol.
...
+    response = await got('todomvc.com', {cache: map});
+    console.log(response.fromCache);
+    //=> true
+})();
+```
+
+Got uses [Keyv](https://github.com/lukechilds/keyv) internally to support a wide range of storage adapters. For something more scalable you could use an [official Keyv storage adapter](https://github.com/lukechilds/keyv#official-storage-adapters):
+
+```
+npm install --save @keyv/redis
```
sindresorhus: `--save` is no longer needed.

readme.md
```diff
@@ -282,6 +294,57 @@ When given an unsupported protocol.
...
+const KeyvRedis = require('@keyv/redis');
+
+const redis = new KeyvRedis('redis://user:pass@localhost:6379');
+
+got('todomvc.com', {cache: redis});
+```
+
+Got supports anything that follows the Map API so it's easy to write your own storage adapter or use a third-party solution.
+
+e.g The following are all valid storage adapters
```
sindresorhus: > For example, the following are all valid storage adapters:

sindresorhus:
Looks really good :)

Tip: Use this view for reviewing (to ignore whitespace changes): https://github.com/sindresorhus/got/pull/284/files?w=1

lukechilds:
@sindresorhus thanks for taking the time to go over this!

All latest requested changes have been implemented.

Two things I'm not quite sure about:

**New Properties**

> `.fromCache` and any other extra properties on the response needs to be documented.
https://github.com/sindresorhus/got/pull/284#discussion_r142334164

Is this ok: https://github.com/sindresorhus/got/pull/284/commits/4d9ab592dfb34ec43977c2a44e3bf4744336760e

Pulled straight from `cacheable-request` docs:
https://github.com/lukechilds/cacheable-request#cb

**Cache link**

There are currently two headings titled "cache", which means the href `#cache` goes to the api docs, and `#cache-1` goes to the main cache section.

Is this ok? Obviously works fine as is but wondering if re-ordering or something in the future could break the other link.

https://github.com/lukechilds/got/tree/cache#cache
https://github.com/lukechilds/got/tree/cache#cache-1

sindresorhus:
> There are currently two headings titled "cache", which means the href #cache goes to the api docs, and #cache-1 goes to the main cache section.

I think you can use an HTML link to force it:

```markdown
<a name="cache"></a>
### Cache
```

lukechilds:
Is this ok?

https://github.com/sindresorhus/got/pull/284/commits/0a75785812c9acf33cc1dc8b8a4ad863a243b6c6

> Rather say it will be responselike when it's from the cache, but it should behave exactly like the original

It's worth noting, it won't behave *exactly* like the original, because the original has references to the socket and stuff which I obviously can't do with a cached response. However none of those core Node.js HTTP response stream methods are actually explicitly mentioned in the Got docs so you could consider them undocumented properties from Got's perspective and argue the behaviour is exactly the same.

Anyway, I linked to the [responselike](https://github.com/lukechilds/responselike) module so they can view more information on exactly what properties is does/doesn't support.

lukechilds:
Hey, just incase you're about to merge this, can I just make a quick tweak to Keyv first. It won't change any current behaviour but it will store the data in a way that will make it much easier for me to make changes to Keyv in the future without affecting 3rd party users like Got.

Wanna get this in before Keyv gets wide scale usage, it's literally only a tiny change, just not backwards compatible, working on it now.

lukechilds:
Sorted!

Details on why the change was needed here: https://github.com/lukechilds/keyv/releases/tag/v3.0.0

sindresorhus:
Landed! Thank you so much for working on this 🙌

superhighfive

lukechilds:
My pleasure!

PR #285, opened 2017-03-23
Require with extension for json files

alexandrusavin:
Packaging apps (webpack etc.) default to js extensions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ const unzipResponse = require('unzip-response');
 const createErrorClass = require('create-error-class');
 const isRetryAllowed = require('is-retry-allowed');
 const Buffer = require('safe-buffer').Buffer;
-const pkg = require('./package');
+const pkg = require('./package.json');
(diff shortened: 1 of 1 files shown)


sindresorhus:
Sorry, I'm not going to change something because of Webpack. It's up to Webpack to support the Node.js ecosystem. Open an issue on Webpack ;)

AlexTes:
Why not use [read-pkg](https://github.com/sindresorhus/read-pkg) @sindresorhus ?
I like reading JSON to be explicit, but that's just intuition. Interestingly both `read-pkg` and [load-json-file](https://github.com/sindresorhus/load-json-file) are yours, why not use them here?

sindresorhus:
@AlexTes `require` works perfectly for this use-case. We want to read the package.json once and cache it for the lifetime of the process. `read-pkg` would read it for every Got usage. `require` is also faster and less dependencies.

PR #287, opened 2017-03-25
Add ability to cancel in-progress requests

### LINKED ISSUE #274: .cancel() method on the promise
Would is be possible to expose the [http.ClientRequest](https://nodejs.org/api/http.html#http_class_http_clientrequest) as `promise.ClientRequest` or similar, so one can `.abort()` pending requests?

I could use this for https://github.com/sindresorhus/is-online/pull/39.

AlexTes:
* [x] Test canceling in progress request.
  * [x] Call `.cancel()` at the right moment.
* [x] Check if it's possible to abort during an in-progress request.
* [x] Fix the PCancelable rejection blowing up instead of hitting our catch handler.
* [x] Document cancel option

The documentation on aborting is unfortunately pretty skimpy. It appears Node destroys the socket on abort, so it sounds like we can abort during any phase of the request.

---
Fixes #274

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ const createErrorClass = require('create-error-class');
 const isRetryAllowed = require('is-retry-allowed');
 const Buffer = require('safe-buffer').Buffer;
 const isPlainObj = require('is-plain-obj');
+const PCancelable = require('p-cancelable');
 const pkg = require('./package');
 
 function requestAsEventEmitter(opts) {
@@ -93,10 +94,23 @@ function requestAsEventEmitter(opts) {
 }
 
 function asPromise(opts) {
-	return new Promise((resolve, reject) => {
+	return new PCancelable((onCancel, resolve, reject) => {
 		const ee = requestAsEventEmitter(opts);
+		let cancelOnRequest = false;
+
+		onCancel(() => {
+			cancelOnRequest = true;
+		});
 
 		ee.on('request', req => {
+			if (cancelOnRequest) {
+				req.abort();
+			}
+
+			onCancel(() => {
+				req.abort();
+			});
+
 			if (isStream(opts.body)) {
 				opts.body.pipe(req);
 				opts.body = undefined;
@@ -357,6 +371,7 @@ function stdError(error, opts) {
 
 got.RequestError = createErrorClass('RequestError', stdError);
 got.ReadError = createErrorClass('ReadError', stdError);
+got.InvalidCancelError = createErrorClass('InvalidCancelError', stdError);
 got.ParseError = createErrorClass('ParseError', function (e, statusCode, opts, data) {
 	stdError.call(this, e, opts);
 	this.statusCode = statusCode;
--- a/index.js
+++ b/index.js
@@ -371,7 +371,6 @@ function stdError(error, opts) {
 
 got.RequestError = createErrorClass('RequestError', stdError);
 got.ReadError = createErrorClass('ReadError', stdError);
-got.InvalidCancelError = createErrorClass('InvalidCancelError', stdError);
 got.ParseError = createErrorClass('ParseError', function (e, statusCode, opts, data) {
 	stdError.call(this, e, opts);
 	this.statusCode = statusCode;
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   "devDependencies": {
     "ava": "^0.18.0",
     "coveralls": "^2.11.4",
+    "delay": "^2.0.0",
     "form-data": "^2.1.1",
     "get-port": "^3.0.0",
     "into-stream": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
   "devDependencies": {
     "ava": "^0.18.0",
     "coveralls": "^2.11.4",
-    "delay": "^2.0.0",
     "form-data": "^2.1.1",
     "get-port": "^3.0.0",
     "into-stream": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "is-retry-allowed": "^1.0.0",
     "is-stream": "^1.0.0",
     "lowercase-keys": "^1.0.0",
+    "p-cancelable": "^0.2.0",
     "safe-buffer": "^5.0.1",
     "timed-out": "^4.0.0",
     "unzip-response": "^2.0.1",
--- a/readme.md
+++ b/readme.md
@@ -359,6 +359,7 @@ request(`https://${config.host}/production/`, {
 });
 ```
 
+
 ## Tips
 
 ### User Agent
--- a/readme.md
+++ b/readme.md
@@ -12,7 +12,7 @@
 
 A nicer interface to the built-in [`http`](http://nodejs.org/api/http.html) module.
 
-It supports following redirects, promises, streams, retries, automagically handling gzip/deflate and some convenience options.
+It supports following redirects, promises, streams, retries, automagically handling gzip/deflate, canceling of requests, and some convenience options.
 
 Created because [`request`](https://github.com/request/request) is bloated *(several megabytes!)*.
 
@@ -217,6 +217,9 @@ When server redirects you more than 10 times.
 
 When given an unsupported protocol.
 
+## Aborting the request
+
+The promise returned by Got has a `.cancel()` function which, when called, aborts the request.
 
 ## Proxies
 
@@ -355,7 +358,6 @@ request(`https://${config.host}/production/`, {
 	// All usual `got` options
 });
 ```
-
 
 ## Tips
 
--- a/readme.md
+++ b/readme.md
@@ -217,9 +217,11 @@ When server redirects you more than 10 times.
 
 When given an unsupported protocol.
 
+
 ## Aborting the request
 
 The promise returned by Got has a `.cancel()` function which, when called, aborts the request.
+
 
 ## Proxies
 
--- a/test/http.js
+++ b/test/http.js
@@ -147,7 +147,7 @@ test.serial('cancel in-progress request', async t => {
 	});
 
 	await t.throws(p, PCancelable.CancelError);
-	await t.notThrows(cancelPromise);
+	await t.notThrows(() => cancelPromise);
 });
 
 test.serial('cancel immediately', async t => {
@@ -155,7 +155,7 @@ test.serial('cancel immediately', async t => {
 	p.cancel();
 
 	await t.throws(p, PCancelable.CancelError);
-	await t.notThrows(cancelPromise);
+	await t.notThrows(() => cancelPromise);
 });
 
 test.after('cleanup', async () => {
--- a/test/http.js
+++ b/test/http.js
@@ -1,13 +1,8 @@
-import stream from 'stream';
 import test from 'ava';
-import getStream from 'get-stream';
-import PCancelable from 'p-cancelable';
 import got from '../';
 import {createServer} from './helpers/server';
 
-const Readable = stream.Readable;
 let s;
-let cancelPromise = null;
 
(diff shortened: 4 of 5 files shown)


sindresorhus:
Yes, I think we can abort the request whenever, so I think all we have to do is make `promise.cancel()` call `request.abort()`.

sindresorhus:
> Fix the PCancelable rejection blowing up instead of hitting our catch handler. 

Where is this happening?

AlexTes:
Hi, thanks for the feedback, happy to get it 😁. 
```js
$ ava test/http.js
body is a stream, piping
calling cancel

  1 failed

  cancel in-progress request promise
  node_modules/p-cancelable/index.js:65

   64:     this._canceled = true;
   65:     this._reject(new CancelError());
   66:   }

  Error: undefined

  Promise.cancel (node_modules/p-cancelable/index.js:65:16)
  Timeout._onTimeout (test/http.js:131:18)

  The .only() modifier is used in some tests. 11 tests were not run
```

sindresorhus:
Would you be able to submit a [failing test](https://github.com/avajs/ava#failing-tests) for `p-cancelable`? I never got time to dig into this. A failing test would make it much easier to investigate.

AlexTes:
`p-cancelable` has been updated with a fix.

Final notes:
* not sure about canceling before the request event is emitted. used an approach that waits for the request to start then tells Node to abort.
* p-cancelable simply returns on canceling a settled promise. I worry about the use-case where a user tries to cancel a settled promise but won't know the request actually went through. I can imagine that's problematic especially when a user doesn't care much about a response but is going for more of an RPC type deal.

AlexTes:
Thanks, forgot to save the dep update!

sindresorhus:
> not sure about canceling before the request event is emitted. used an approach that waits for the request to start then tells Node to abort.

Is that really necessary? I think you can abort a request whenever.

> p-cancelable immediately when asked to cancel a settled promise.

It immediately does what?

AlexTes:
> Is that really necessary? I think you can abort a request whenever.

Let me make sure and test.

> It immediately does what?

Sorry. '`p-cancelable` returns immediately when asked to cancel a settled promise.'

AlexTes:
Heyo, I predicted an issue but forgot to make an effort to expand the test to actually catch the issue. I shelved it because the pretty approach doesn't work due to.. long story. Anyway, last two commits show the issue and apply the stash I had to solve it.

sindresorhus:
> Anyway, last two commits show the issue - not a pretty test - and apply the stash I had that solves the issue.

Ah ok. I get it now.

index.js
```diff
@@ -87,10 +88,23 @@ function requestAsEventEmitter(opts) {
...
+		});
 
 		ee.on('request', req => {
+			if (cancelImmediately) {
+				req.abort();
+			}
+
+			onCancel(() => {
+				req.abort();
+			});
```
sindresorhus: Test pass without this. Can you add a test to cover it?

test/http.js
```diff
@@ -110,6 +120,40 @@ test('response contains url', async t => {
...
+test('cancel immediately', async t => {
+	const p = got(`${s.url}/store-id`);
+	p.cancel();
+	await t.throws(p, PCancelable.CancelError);
+
+	await delay(100);
+	t.false(cancelRequestCompleted);
+});
+
+test.failing('cancel throws after request completed', async t => {
```
sindresorhus: What is this about?

AlexTes: > p-cancelable returns immediately when asked to cancel a settled promise. I worry about the use-case where a user tries to cancel a settled promise but won't know the request went through. I can imagine that's problematic especially when a user doesn't care much about a response but is going for more of an RPC type deal.

index.js
AlexTes: Hmmm. Alright. That means coming up with a way to easily test the request actually gets canceled. Don't want to simply repeat what I did in the previous test. I'll come up with something clean. Give me a couple minutes.

test/http.js
sindresorhus: It works exactly like trying to reject a settled promise. When I promise is settled, it's by the name of it *settled*. Same with cancel. If the user use `.cancel()`, they should be aware they might call it too late and handle that themselves if they need. We can document this well. They can even check the promise whether it was canceled or not with `promise.canceled` (Which is `false` if you try canceling it after the promise is settled). I think `p.cancel()` throwing based on timing could lead to race issues and would be surprising to the user.

Please call me out if you think I'm totally wrong in this. I'm happy to discuss.

sindresorhus:
@AlexTes In case you missed it, note: https://github.com/sindresorhus/got/pull/287#discussion_r115069123

test/http.js
AlexTes: I did a little testing and a little thinking. Trying to resolve an already settled promise seems to do nothing in the case of Node too. A note in the documentation is enough.
I think you're right on this.

AlexTes:
Not sure where to leave the documentation for this. It's a sort of indirect API. I took my best guess.

test/http.js
```diff
@@ -115,6 +133,31 @@ test('response contains url', async t => {
...
+	await t.throws(p, PCancelable.CancelError);
+	await t.notThrows(() => cancelPromise);
+});
+
+test.serial('cancel immediately', async t => {
+	const p = got(`${s.url}/no-res-store-id`, {body: '', retries: 0});
+	p.cancel();
+
+	await t.throws(p, PCancelable.CancelError);
+	await t.notThrows(() => cancelPromise);
```
sindresorhus:
```js
await t.notThrows(cancelPromise);
```

test/http.js
```diff
@@ -115,6 +133,31 @@ test('response contains url', async t => {
 	t.is((await got(s.url)).url, `${s.url}/`);
 });
 
+test.serial('cancel in-progress request', async t => {
+	const body = new Readable({
+		read() {}
+	});
+	body.push(String(1));
```
sindresorhus:
```js
body.push('1');
```

test/http.js
```diff
@@ -29,6 +34,19 @@ test.before('setup', async () => {
 		res.end('recent');
 	});
 
+	s.on('/no-res-store-id', (req, res) => {
+		cancelPromise = new Promise((resolve, reject) => {
+			req.on('aborted', resolve);
+			res.on('finish', () => {
+				reject(new Error('request finished instead of aborting'));
```
sindresorhus: `request` => `Request`

test/http.js
```diff
@@ -1,8 +1,13 @@
+import stream from 'stream';
+import PCancelable from 'p-cancelable';
```
sindresorhus: Place this after `from 'ava'`

readme.md
```diff
@@ -356,6 +356,9 @@ request(`https://${config.host}/production/`, {
 });
 ```
 
+## Aborting the request
```
sindresorhus: Can you place this above the `Proxies` section?

sindresorhus: Maybe we should say `Canceling the request` instead?

readme.md
```diff
@@ -356,6 +356,9 @@ request(`https://${config.host}/production/`, {
...
+
+The promise returned by got features a `.cancel` function. It takes no arguments and will abort the request, destroying the underlying socket if called.
```
sindresorhus: got => Got

.cancel => .cancel()

---

I would make it more succinct:

> The promise returned by Got has a `.cancel()` function, which when called, aborts the request.

index.js
```diff
@@ -357,6 +371,7 @@ function stdError(error, opts) {
 
 got.RequestError = createErrorClass('RequestError', stdError);
 got.ReadError = createErrorClass('ReadError', stdError);
+got.InvalidCancelError = createErrorClass('InvalidCancelError', stdError);
```
sindresorhus: We don't need this anymore.

test/http.js
```diff
@@ -115,6 +133,31 @@ test('response contains url', async t => {
...
+
+	const p = got(`${s.url}/no-res-store-id`, {body});
+
+	setImmediate(() => {
+		p.cancel();
+		body.push(null); // Complete the request that should now be canceled
+	});
+
+	await t.throws(p, PCancelable.CancelError);
+	await t.notThrows(() => cancelPromise);
```
sindresorhus:
```js
await t.notThrows(cancelPromise);
```

sindresorhus:
Can you mention cancelation in the readme intro?

readme.md
AlexTes: Done :3.
I don't think we should. It sounds worse, granted, but Node consistently calls it 'abort'.

package.json
```diff
@@ -63,6 +64,7 @@
   "devDependencies": {
     "ava": "^0.18.0",
     "coveralls": "^2.11.4",
+    "delay": "^2.0.0",
```
AlexTes: 🤦‍♀️

readme.md
sindresorhus: Alright. Let's do it.

test/http.js
AlexTes: Cool, so we alphabetize after complying with the import plugin rules?

sindresorhus: No, I just prefer the order:

- builtins
- ava (placed here because it's the main external import)
- external
- local

sindresorhus:
Can you add commits instead of rebasing? Makes it easier to see what changed. We'll squash on merge anyways.

AlexTes:
Sure, will do that in the future. I asked you on Gitter when you squash and when you don't. The only reason I was rebasing is because I wasn't sure if you'd squash.

AlexTes:
I'll see about reviewing my own PR in the morning next time I'm coding late. Too many slip ups from code that changed heavily during the hour I was trying to find a strong way to write these tests.

AlexTes:
Alright, ready for re-review @sindresorhus.

test/http.js
AlexTes: I just tried this. The test fails without. Turned it back into what it was. I remember now I wrote it your suggested way first. Not sure why this demands to be lazily evaluated. My JS-foo is not strong enough (yet).

sindresorhus: @AlexTes `t.notThrows` either accepts a promise that rejects or a function that throws. AVA is actually catching a bug in your test here as `cancelPromise` is `null`, not a Promise. Try: `console.log(cancelPromise === null);`

AlexTes: Of course. I'll fix it.

sindresorhus: Ping :)

AlexTes:
So I finally finished trying out all ideas I could come up with for solid tests. The in-flight one looks quite solid to me now. It starts a streamed body request, doesn't close the stream until the got triggers the cancel. To make sure we abort on cancel the test then verifies the server emitted an abort event, not a finish event.

The immediate cancel test is much harder. First off, by 'immediate cancel' I mean a user canceling a request before got's underlying event emitter has even emitted its request event. I tried all ideas I had, but the core issue remains. When you cancel a request, you end up with a rejected promise and pretty much lose the opportunity to get any further feedback. After extensive testing, it seems the 'connection' event is the event that hits a Node HTTP server first. This event doesn't fire with an immediate cancel. In other words, unless more of got's internals are exposed it appears impossible to make sure got ever triggered the abort. The compromise I've currently gone with is to make the following assumption: if the server has not received a request within 1000ms, the abort was executed client-side.

Any better method of testing is very welcome, as are of course any pointers to clean up the code more. Sorry for taking so long, this PR has proven to be a very tough one to crack.

sindresorhus:
> Any better method of testing is very welcome, as are of course any pointers to clean up the code more.

I can't think of any better way either.

sindresorhus:
This is great stuff @AlexTes. Really appreciate all the hard work and thoughts you put into it. 🙌

superhighfive

PR #289, opened 2017-03-26
Add WHATWG URL support

### LINKED ISSUE #228: WHATWG-URL support
[`URL`](https://developer.mozilla.org/en/docs/Web/API/URL) / [whatwg-url](https://npmjs.com/whatwg-url) is the future.

AlexTes:
So I've gotten stuck implementing this.
The way I imagine things, this PR is a _bad_ idea or I've been way overthinking the whole time. With this PR the nuances between the way Node handles a [Node urlObject](https://nodejs.org/api/url.html#url_url_strings_and_url_objects) or [whatwg urlObject](https://nodejs.org/api/url.html#url_the_whatwg_url_api) would be lost. I tried to dig into the Node v8 nightly to confirm but couldn't find anything. I assumed this can't be done because this PR trivially supports whatwg-urls on any Node and I thought that wasn't possible.

The second suggestion was keeping the URL object intact and supporting whatwg-url on Node >= v8. This makes less sense the more I think about it. Some issues I can think of:
* Any place you read from or write to the URL you'd have to be aware of which type you're dealing with. (If you normalize you're effectively doing what this PR is)
* We currently support all of the `http.request` options, which we use to overwrite the URL object. 
* How does one pass node's `http.request` options that are not Node URL properties? If you have to serialize the whatwg-url in order to do that what's the point keeping it intact in the first place? The current Node implementation runs it through [url.parse](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) anyway. If that is going to stay the same in v8, this PR seems like an adequate solution.

I guess the core question I'm asking is: does it matter if you serialize a whatwg-url then parse it with node's url.parse or directly pass it to node? I can't tell from skimming the whatwg-url spec.

Note to self: if we're going with normalizing whatwg-url's take care of sneaking in auth. 
Sidenote to self: dealing with features supported in some versions of Node is tricky 😓.

---

Fixes #228

### DIFF
--- a/index.js
+++ b/index.js
@@ -209,6 +209,10 @@ function normalizeArguments(url, opts) {
 
 	if (isURL.lenient(url)) {
 		url = urlParseLax(url.href);
+
+		if (url.auth) {
+			throw new Error('Basic authentication must be done with auth option');
+		}
 	}
 
 	opts = Object.assign(
--- a/index.js
+++ b/index.js
@@ -16,6 +16,7 @@ const unzipResponse = require('unzip-response');
 const createErrorClass = require('create-error-class');
 const isRetryAllowed = require('is-retry-allowed');
 const Buffer = require('safe-buffer').Buffer;
+const isURL = require('isurl');
 const pkg = require('./package');
 
 function requestAsEventEmitter(opts) {
@@ -204,6 +205,10 @@ function normalizeArguments(url, opts) {
 		if (url.auth) {
 			throw new Error('Basic authentication must be done with auth option');
 		}
+	}
+
+	if (isURL.lenient(url)) {
+		url = urlParseLax(url.href);
 	}
 
 	opts = Object.assign(
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "pem": "^1.4.4",
     "pify": "^2.3.0",
     "tempfile": "^1.1.1",
-    "whatwg-url": "^4.7.1",
+    "universal-url": "^1.0.0-alpha",
     "xo": "^0.18.0"
   },
   "ava": {
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "is-redirect": "^1.0.0",
     "is-retry-allowed": "^1.0.0",
     "is-stream": "^1.0.0",
-    "isurl": "^1.0.0-alpha3",
+    "isurl": "^1.0.0-alpha5",
     "lowercase-keys": "^1.0.0",
     "safe-buffer": "^5.0.1",
     "timed-out": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "is-redirect": "^1.0.0",
     "is-retry-allowed": "^1.0.0",
     "is-stream": "^1.0.0",
+    "isurl": "^1.0.0-alpha3",
     "lowercase-keys": "^1.0.0",
     "safe-buffer": "^5.0.1",
     "timed-out": "^4.0.0",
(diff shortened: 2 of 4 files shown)


sindresorhus:
I think this is the correct approach. We just want to support WHATWG URLs now, not use them internally. Can you also document this?

package.json
```diff
@@ -48,6 +48,7 @@
     "is-redirect": "^1.0.0",
     "is-retry-allowed": "^1.0.0",
     "is-stream": "^1.0.0",
+    "isurl": "^1.0.0-alpha3",
```
sindresorhus: It's `1.0.0-alpha5` now

package.json
```diff
@@ -64,6 +65,7 @@
     "pem": "^1.4.4",
     "pify": "^2.3.0",
     "tempfile": "^1.1.1",
+    "whatwg-url": "^4.6.0",
```
sindresorhus: And `^4.7.1`

package.json
AlexTes: This one doesn't work 😓 .

sindresorhus: // @stevenvachon

sindresorhus:
@stevenvachon Can you review?

sindresorhus:
@AlexTes Tests are failing

package.json
stevenvachon: `isURL.lenient(url)`

test/arguments.js
```diff
@@ -1,3 +1,4 @@
+import {URL} from 'whatwg-url';
```
stevenvachon: It might be more complete to use [universal-url](https://npmjs.com/universal-url) here so that the Node v7 `URL` is included in tests when available.

index.js
```diff
@@ -206,6 +207,10 @@ function normalizeArguments(url, opts) {
 		}
 	}
 
+	if (isURL(url)) {
+		url = urlParseLax(url.href);
```
stevenvachon: There's no auth (`username`,`password`) check in this block like there is for string input.

readme.md
```diff
@@ -62,7 +62,7 @@ Returns a Promise for a `response` object with a `body` property, a `url` proper
 
 Type: `string`, `object`
 
-The URL to request or a [`http.request` options](https://nodejs.org/api/http.html#http_http_request_options_callback) object.
+The URL to request as simple string, a [`http.request` options](https://nodejs.org/api/http.html#http_http_request_options_callback), or a [`WHATWG-URL`](https://nodejs.org/api/url.html#url_class_url).
```
stevenvachon: This might make more sense:
> , or a [WHATWG `URL`](https://nodejs.org/api/url.html#url_class_url).

index.js
stevenvachon: Also, reparsing the url is not the best idea for performance.

AlexTes:
Thanks for the feedback! Will update tomorrow 😄.

test/arguments.js
AlexTes: Sure. It'd be nice to test both the native and shim. I'm swapping it out. Just to make sure, how does the rest feel about going from a lib that's seen tons of use to something that's still in alpha?

AlexTes:
@sindresorhus @stevenvachon ready for review 😁!

package.json
```diff
@@ -69,6 +70,8 @@
     "nyc": "^10.0.0",
     "pem": "^1.4.4",
     "pify": "^2.3.0",
+    "universal-url": "^1.0.0-alpha",
```
stevenvachon: Alphabetical order.

AlexTes: Sorry 😅. It was late.

PR #295, opened 2017-05-03
Check protocol before sending request

### LINKED ISSUE #294: Got throws on unsupported protocol
Hey, we've recently started using the latest version of [probe-image-size](https://github.com/nodeca/probe-image-size), which under the hood uses `got`.

Since then, we've seen a bunch of those:

```js
_http_client.js:55
    throw new Error('Protocol "' + protocol + '" not supported. ' +
    ^
2017-05-03T12:32:25.058531056Z 
Error: Protocol "c:" not supported. Expected "http:"
    at new ClientRequest (_http_client.js:55:11)
    at Object.exports.request (http.js:31:10)
    at opts (/usr/src/app/node_modules/got/index.js:33:18)
    at ClientRequest.fn.request.res (/usr/src/app/node_modules/got/index.js:51:5)
    at ClientRequest.g (events.js:286:16)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:472:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:105:23)
    at TLSSocket.socketOnData (_http_client.js:361:20)
    at emitOne (events.js:96:13)
    at TLSSocket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at TLSSocket.Readable.push (_stream_readable.js:130:10)
    at TLSWrap.onread (net.js:542:20)
```

We're running an application that requires scraping a wide variety of web pages, which in turn contain an even wider variety of images, which we probe using `got`.

While image urls starting with `c:` are not common at all and will likely not resolve to anything helpful, we'd like to not return 5xx errors when such weird images are found on a page, and fail more helpfully instead.

Adding a simple url validity check before requesting them would solve this.

I already implemented such a check in our application code, which now fails early instead of trying to probe an url that uses a scheme different from `http://` or `https://`.

This isn't enough, as such errors are still coming in, which I've identified as coming from redirects, e.g. <https://tinyurl.com/lbf3aq9> is valid itself, but will 301 redirect to the fictitious and invalid `c://users/pioul/Downloads/icon.jpg`, which will throw.

This fix would probably be as simple as adding a check [around here](https://github.com/sindresorhus/got/blob/7b12b7582d71abc652013d3b967473a776abf481/index.js#L32) – what do you think?

pioul:
Fixes #294 by making sure an url has a valid protocol (http/https) before sending a request.

Let me know if any other tweak would make sense before merging this PR!

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,6 +28,11 @@ function requestAsEventEmitter(opts) {
 	let redirectUrl;
 
 	const get = opts => {
+		if (opts.protocol !== 'http:' && opts.protocol !== 'https:') {
+			ee.emit('error', new got.UnsupportedProtocolError(opts));
+			return;
+		}
+
 		const fn = opts.protocol === 'https:' ? https : http;
 
 		const req = fn.request(opts, res => {
@@ -208,11 +213,13 @@ function normalizeArguments(url, opts) {
 
 	opts = Object.assign(
 		{
-			protocol: 'http:',
 			path: '',
 			retries: 5
 		},
 		url,
+		{
+			protocol: url.protocol || 'http:' // Override both null/undefined with default protocol
+		},
 		opts
 	);
 
@@ -362,6 +369,11 @@ got.MaxRedirectsError = createErrorClass('MaxRedirectsError', function (statusCo
 	this.statusCode = statusCode;
 	this.statusMessage = http.STATUS_CODES[this.statusCode];
 	this.message = 'Redirected 10 times. Aborting.';
+});
+
+got.UnsupportedProtocolError = createErrorClass('UnsupportedProtocolError', function (opts) {
+	stdError.call(this, {}, opts);
+	this.message = `Unsupported protocol "${opts.protocol}"`;
 });
 
 module.exports = got;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -208,11 +213,13 @@ function normalizeArguments(url, opts) {
...
-			protocol: 'http:',
 			path: '',
 			retries: 5
 		},
 		url,
+		{
+			protocol: url.protocol || 'http:' // Override both null/undefined with default protocol
+		},
 		opts
 	);
```
pioul: `urlParseLax` returns `null` for all properties that weren't parsed, so `protocol: null` used to take over the default `protocol: 'http:'` all the time for protocol-less urls; this change remedies that.

sindresorhus:
Thanks @pioul :)

pioul:
👍 @sindresorhus Of course! Just realizing now that there's a bit of doc I haven't updated if you want to quickly add a line there: https://github.com/sindresorhus/got#errors

By the way, how often are you releasing? I'd love to take advantage of that addition in a dependent ([probe-image-size](https://github.com/nodeca/probe-image-size)) once the next release is out :)

sindresorhus:
Readme updated. I missed it too.

Right now we're targeting a new major release in a couple of weeks.

pioul:
Thanks, sounds great!

PR #297, opened 2017-05-04
Change stringifying options behavior

### LINKED ISSUE #174: JSON option when posting data

AlexTes:
Changes the behavior of the stringifying options.

* [x] don't accept plain object unless `form` or `json` is true
* [x] stringify if `json` is true
* [x] stringify if `form` is true
* [x] stringify as querystring and parse `res` as JSON if `form` and `json` are true

Things I changed because I found the code hard to deal with, and confusing without.
* clean up input validation using Sindre's suggestion from #212.
* simpler use of `body` variable. It's a shorthand for the body passed by the user. `opts.body` is what we mutate into the body we want to send.
* 'throws on bad input' test is maybe overkill

As always details are in the commit messages.

Sidenotes: I went for minimal change based on the proposal from #174. I think this can be quite a bit cleaner still. I'm considering a second PR. Specifically what I feel could be better:
* make 'body is string' case explicit instead of implicit fallback
* use early returns to limit code paths and sharpen reduce redundant input checking
* some tests use destructuring others don't _for the exact same response property_.

---
Fixes #174


sindresorhus:
@floatdrop What do you think of these changes? See my proposal in https://github.com/sindresorhus/got/issues/174#issuecomment-298292987

sindresorhus:
> I'm considering a second PR. Specifically what I feel could be better:

:+1:

AlexTes:
I got an avocado rainbow 😲 ✊!

index.js
```diff
@@ -236,31 +237,37 @@ function normalizeArguments(url, opts) {
...
-			throw new Error('options.body must be a ReadableStream, string, Buffer or plain Object');
+	const body = opts.body;
+	if (body !== null && body !== undefined) {
+		const headers = opts.headers;
+		if (!isStream(body) && typeof body !== 'string' && !Buffer.isBuffer(body) && !(opts.form || opts.json)) {
+			throw new TypeError('options.body must be a ReadableStream, string, Buffer or plain Object');
 		}
 
-		opts.method = opts.method || 'POST';
+		if ((opts.form || opts.json) && !(isObj(body) && typeof body !== 'function')) {
```
floatdrop: Is `typeof body !== 'funciton'` is necessary? Looks like `isPlainObj(function() {})` returns `false`.

index.js
```diff
@@ -236,31 +237,37 @@ function normalizeArguments(url, opts) {
...
+			throw new TypeError('options.body must be a plain Object');
```
floatdrop: `options.body must be a plain Object when options.form or options.json is used`

index.js
```diff
@@ -236,31 +237,37 @@ function normalizeArguments(url, opts) {
...
-			const length = typeof body === 'string' ? Buffer.byteLength(body) : body.length;
-			opts.headers['content-length'] = length;
+		if (headers['content-length'] === undefined && headers['transfer-encoding'] === undefined && !isStream(body)) {
+			const length = typeof opts.body === 'string' ? Buffer.byteLength(opts.body) : opts.body.length;
+			headers['content-length'] = length;
 		}
-	}
 
-	opts.method = (opts.method || 'GET').toUpperCase();
+		opts.method = opts.method || 'POST';
```
floatdrop: `opts.method` should be uppercased in this if-branch too.

index.js
AlexTes: Nice catch. I was confusing `is-obj` and `is-plain-obj`.

sindresorhus:
Thank you for yet another excellent pull request @AlexTes. I'd like to invite you to join the project if you're interested? 🙏

AlexTes:
@sindresorhus 🙇 🙇 . You just made my day. And it's only 10 AM!

PR #301, opened 2017-05-08
Only throw ParseError on 2xx responses

### LINKED ISSUE #278: Handling both ParseError and HTTPError simultaneously
[Previous discussion](https://github.com/SamVerschueren/travis-got/pull/2#pullrequestreview-25498565) /cc @SamVerschueren @floatdrop @sindresorhus

To summarize, Travis CI’s API returns HTML on `403 Forbidden` responses. For example, try:

```sh
curl -i -H "Authorization: token fail" https://api.travis-ci.org/repos/SamVerschueren/travis-got 
```

When using got’s `json` option this causes a `ParseError` (instead of an `HTTPError`) to be thrown. To avoid this, I suggested to catch the error, check the status code for `403`, and re-throw an `HTTPError`. While this works, it requires us to manually create an error which is an undocumented API of got, as far as I can tell.

@SamVerschueren suggested to disable got’s `json` option and parse the body ourselves. Again, while this works, we would still have to manually throw a `ParseError` in case the body can’t be parsed.

So, either way, we would have to to manually create and throw a `ParseError` or an `HTTPError`.

> I would recommend looking into how other HTTP libs handle this, like `request`. We might be doing something wrong, or not.

@sindresorhus request currently just [logs a debug message](https://github.com/request/request/blob/v2.80.1/request.js#L1116-L1120). This behavior has also been discussed in https://github.com/request/request/issues/440.

sonicdoe:
Resolves #278.

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,7 +122,9 @@ function asPromise(opts) {
 						try {
 							res.body = JSON.parse(res.body);
 						} catch (e) {
-							throw new got.ParseError(e, statusCode, opts, data);
+							if (statusCode >= 200 && statusCode < 300) {
+								throw new got.ParseError(e, statusCode, opts, data);
+							}
 						}
 					}
 
--- a/readme.md
+++ b/readme.md
@@ -203,7 +203,7 @@ When reading from response stream fails.
 
 #### got.ParseError
 
-When `json` option is enabled and `JSON.parse` fails.
+When `json` option is enabled, server response code is 2xx, and `JSON.parse` fails.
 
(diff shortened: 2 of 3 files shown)


floatdrop:
We should add a note about parsing error policy, I guess.

sindresorhus:
@sonicdoe Can you add a note in the readme about the behavior?

sonicdoe:
Of course, done.

sindresorhus:
Thank you! :)

PR #302, opened 2017-05-10
Include response headers and redirect URLs in error object

### LINKED ISSUE #234: Get intermediate redirect urls
It would be useful if there were a way to get all redirected urls in a request. So if you request url `a`, but it redirects to `b` which redirects to `c` and finally to `d`, if there were a `urls` property on the response that would return `['a', 'b', 'c', 'd']`. There's the `redirect` event that could be used to capture these urls but that seems to only work when using streams.

djmadeira:
Fixes #234.

- Adds a response headers object to HTTPError
- Adds an array of redirects that were followed to the response object and MaxRedirectsError

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,7 +46,7 @@ function requestAsEventEmitter(opts) {
 				res.resume();
 
 				if (redirects.length >= 10) {
-					ee.emit('error', new got.MaxRedirectsError(statusCode, opts, redirects), null, res);
+					ee.emit('error', new got.MaxRedirectsError(statusCode, redirects, opts), null, res);
 					return;
 				}
 
@@ -379,7 +379,7 @@ got.HTTPError = createErrorClass('HTTPError', function (statusCode, headers, opt
 	this.headers = headers;
 });
 
-got.MaxRedirectsError = createErrorClass('MaxRedirectsError', function (statusCode, opts, redirectUrls) {
+got.MaxRedirectsError = createErrorClass('MaxRedirectsError', function (statusCode, redirectUrls, opts) {
 	stdError.call(this, {}, opts);
 	this.statusCode = statusCode;
 	this.statusMessage = http.STATUS_CODES[this.statusCode];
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -207,11 +207,11 @@ When `json` option is enabled and `JSON.parse` fails.
 
 #### got.HTTPError
 
-When server response code is not 2xx. Contains `statusCode` and `statusMessage`.
+When server response code is not 2xx. Contains `statusCode`, `statusMessage`, and `redirectUrls`.
 
 #### got.MaxRedirectsError
 
-When server redirects you more than 10 times.
+When server redirects you more than 10 times. Has `redirectUrls` property, which is an array of the URLs got was redirected to before giving up.
```
sindresorhus: `Has `redirectUrls` property` => `Includes a `redirectUrls` property`

got => Got

index.js
```diff
@@ -365,18 +371,20 @@ got.ParseError = createErrorClass('ParseError', function (e, statusCode, opts, d
...
+got.HTTPError = createErrorClass('HTTPError', function (statusCode, headers, opts) {
 	stdError.call(this, {}, opts);
 	this.statusCode = statusCode;
 	this.statusMessage = http.STATUS_CODES[this.statusCode];
 	this.message = `Response code ${this.statusCode} (${this.statusMessage})`;
+	this.headers = headers;
 });
 
-got.MaxRedirectsError = createErrorClass('MaxRedirectsError', function (statusCode, opts) {
+got.MaxRedirectsError = createErrorClass('MaxRedirectsError', function (statusCode, opts, redirectUrls) {
```
sindresorhus: `opts` should be last

readme.md
```diff
@@ -207,11 +207,11 @@ When `json` option is enabled and `JSON.parse` fails.
 
 #### got.HTTPError
 
-When server response code is not 2xx. Contains `statusCode` and `statusMessage`.
+When server response code is not 2xx. Contains `statusCode`, `statusMessage`, and `redirectUrls`.
```
sindresorhus: Unrelated, but change this to:

```
Includes `statusCode`, `statusMessage`, and `redirectUrls` properties.
```

djmadeira:
@sindresorhus addressed comments.

PR #320, opened 2017-05-31
Add `decompress` option

### LINKED ISSUE #319: Allow opting out of decompression
Automatically decompressing the response is really helpful for the majority of use cases. However, there may be some situations where you want to deal with the compressed data, currently there's no way to do this apart from letting Got decompress the data and then manually re-compressing it yourself afterwards. Obviously that's very inefficient.

I don't expect there to be a huge amount of use cases that would benefit from this, however it's a very simple (+2 lines) backwards compatible change so I think it's worth adding.

One use case that would massively benefit from this is implementing a reverse proxy server with Got.

This option would allow you to reverse proxy gzip responses with Got by just passing the stream directly through. Combined with the upcoming cache support (#284), I think a caching reverse proxy would be a very powerful use of Got.

lukechilds:
Adds `opts.decompress` option. By default it's `true` and functionality remains the same. Setting `opts.decompress` to false allows you to opt out of Got's automatic decompression.

Resolves #319

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,6 +89,10 @@ function requestAsEventEmitter(opts) {
 					typeof decompressResponse === 'function' &&
 					req.method !== 'HEAD' ? decompressResponse(res) : res;
 
+				if (!opts.decompress && ['gzip', 'deflate'].indexOf(res.headers['content-encoding']) !== -1) {
+					opts.encoding = null;
+				}
+
 				response.redirectUrls = redirects;
 
 				ee.emit('response', response);
--- a/index.js
+++ b/index.js
@@ -85,7 +85,8 @@ function requestAsEventEmitter(opts) {
 			}
 
 			setImmediate(() => {
-				const response = typeof decompressResponse === 'function' &&
+				const response = opts.decompress === true &&
+					typeof decompressResponse === 'function' &&
 					req.method !== 'HEAD' ? decompressResponse(res) : res;
 
 				response.redirectUrls = redirects;
@@ -280,6 +281,7 @@ function normalizeArguments(url, opts) {
 		{
 			path: '',
 			retries: 2,
+			decompress: true,
 			useElectronNet: true
 		},
 		url,
--- a/test/gzip.js
+++ b/test/gzip.js
@@ -6,6 +6,7 @@ import got from '..';
 import {createServer} from './helpers/server';
 
 const testContent = 'Compressible response content.\n';
+const testContentUncompressed = 'Uncompressed response content.\n';
 
 let s;
 let gzipData;
@@ -41,6 +42,12 @@ test.before('setup', async () => {
 		res.end(gzipData.slice(0, -1));
 	});
 
+	s.on('/uncompressed', (req, res) => {
+		res.statusCode = 200;
+		res.setHeader('Content-Type', 'text/plain');
+		res.end(testContentUncompressed);
+	});
+
 	await s.listen(s.port);
 });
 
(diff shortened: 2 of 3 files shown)


lukechilds:
Also, is the existing `typeof decompressResponse === 'function'` check necessary or can it be removed?

Won't decompressResponse` always be a function?

floatdrop:

sindresorhus:
> Won't decompressResponse always be a function?

It was added as the `decompress-response` package is not included when Got is browserified, so we need to handle it not existing.

lukechilds:
> the `decompress-response` package is not included when Got is browserified

Ahh, makes sense 👍

sindresorhus:
> One use case that would massively benefit from this is implementing a reverse proxy server with Got.

Can you include the use-case in the options definition? Not everyone realize what the option might be useful for.

lukechilds:
Yeah, sure.

I added a more generic explanation:

>###### decompress
>
>Type: `boolean`<br>
>Default: `true`
>
>Defines if compressed responses should be decompressed automatically.
>
>If this is disabled the body is returned as a compressed Buffer. This may be useful if you want to handle decompression yourself or stream the raw compressed data.

That ok?

test/gzip.js
```diff
@@ -56,6 +59,11 @@ test('handles gzip error', async t => {
 	t.is(err.name, 'ReadError');
 });
 
+test('decompress option opts out of decompressing', async t => {
+	const response = await got(s.url, {decompress: false, encoding: null});
```
sindresorhus: This doesn't match the docs. Here you explicitly set `encoding: null`, but in the docs you say it's returned as a compressed Buffer when `decompress: false`.

lukechilds: Woops, sorry, completely missed that.

Encoding should be set to null internally if `decompress` is `false`. I'll update.

lukechilds: Fixed in https://github.com/sindresorhus/got/pull/320/commits/c99b2500f46673fd4ffa13d8e0e698d6d4b74e50

index.js
```diff
@@ -289,6 +291,10 @@ function normalizeArguments(url, opts) {
 		opts
 	);
 
+	if (!opts.decompress) {
+		opts.encoding = null;
+	}
```
sindresorhus: This causes a non-compressed response to be a Buffer, even a plain text response. Is that really wanted/expected?

lukechilds: *facepalm*. Of course, sorry.

What's the best solution here?

Copy/paste:

```js
if (['gzip', 'deflate'].indexOf(res.headers['content-encoding']) === -1)
```

straight from `decompress-response`?

Or maybe publish it separately as `responseIsGzipped()` and require in both Got/`decompress-response` so they're both in sync?

sindresorhus: > Copy/paste:

👍

lukechilds: Fixed. Also added test to check `opts.decompress` doesn't alter encoding of uncompressed responses.

index.js
```diff
@@ -85,9 +85,14 @@ function requestAsEventEmitter(opts) {
 			}
 
 			setImmediate(() => {
-				const response = typeof decompressResponse === 'function' &&
+				const response = opts.decompress === true &&
+					typeof decompressResponse === 'function' &&
 					req.method !== 'HEAD' ? decompressResponse(res) : res;
 
+				if (!opts.decompress && ['gzip', 'deflate'].indexOf(res.headers['content-encoding']) !== -1) {
```
lukechilds: @sindresorhus is `!opts.decompress` ok here? It's a boolean. Or should I do `opts.decompress === false`?

sindresorhus: What you have is good.

lukechilds:
Is there anything else needed for this?

sindresorhus:
🍰

PR #322, opened 2017-06-01
Add progress events

vadimdemedes:
This PR adds upload and download progress events.

```js
got.post('https://website.com', {body: fs.createReadStream('file')})
  .on('uploadProgress', progress => {
    // `progress` looks like:
    // {
    //   percent: 0.24,
    //   transferred: 24518,
    //   total: 50000
    // }
    }
  })
  .on('downloadProgress', progress => {
    // `progress` has the same shape as in `uploadProgress` event
  });
```

Initial upload/download progress event with `percent` equal to `0` is always emitted. If it's not possible to retrieve the total upload/download body size, `percent` and `size.total` are equal to `null`.

#### Hack for receiving upload progress events when uploading buffers or strings

When sending `Buffer` as `body`, there are only 2 `uploadProgress` events, because the whole buffer gets written to the connection, so `req.connection.bytesWritten` becomes equal to buffer length. To work around this limitation and receive meaningful progress events, `intoStream` module is used to convert buffer into stream.

One more quirk. Since after conversion, there's no more access to buffer's size, I'm making it accessible by assigning it to `_buffer` on the stream here https://github.com/sindresorhus/got/pull/322/files#diff-168726dbe96b3ce427e7fedce31bb0bcR525 and getting it later again here https://github.com/sindresorhus/got/pull/322/files#diff-168726dbe96b3ce427e7fedce31bb0bcR76.

#### TODO

- [x] Write a description for this PR
- [x] Explain the hack behind wrapping body buffer into a stream when uploading
- [x] Update readme


index.js
```diff
@@ -111,7 +226,42 @@ function requestAsEventEmitter(opts) {
...
+					if (uploaded === lastUploaded || uploaded === uploadBodySize) {
+						return;
+					}
+
+					ee.emit('uploadProgress', {
+						percent: uploadBodySize ? uploaded / uploadBodySize : 0,
+						transferred: uploaded,
+						total: uploadBodySize
+					});
+				}, 150);
```
sindresorhus: Why did we go with `150`? Can you move `150` into a variable with a descriptive name of why it's 150.

index.js
```diff
@@ -24,6 +28,62 @@ const pkg = require('./package');
...
+			return;
+		}
+
+		if (typeof body === 'string') {
+			resolve(Buffer.byteLength(body));
+			return;
+		}
+
+		if (isFormData(body)) {
+			body.getLength((err, size) => {
```
sindresorhus: Would be nicer to use `pify` here and on line 66, so we don't have to wrap the whole thing in the Promise constructor.

index.js
```diff
@@ -121,8 +271,17 @@ function requestAsEventEmitter(opts) {
 	};
 
 	setImmediate(() => {
-		get(opts);
+		getBodySize(opts)
+			.then(size => {
+				uploadBodySize = size;
+
```
sindresorhus: Remove the empty line here.

index.js
```diff
@@ -191,7 +352,17 @@ function asPromise(opts) {
 		});
 
 		ee.on('error', reject);
+		ee.on('uploadProgress', proxy.emit.bind(proxy, 'uploadProgress'));
+		ee.on('downloadProgress', proxy.emit.bind(proxy, 'downloadProgress'));
 	}));
+
+	promise.on = (name, fn) => {
+		proxy.on(name, fn);
+
```
sindresorhus: Remove the empty line here.

index.js
```diff
@@ -388,7 +568,7 @@ function got(url, opts) {
 	}
 }
 
-got.stream = (url, opts) => asStream(normalizeArguments(url, opts));
+got.stream = (url, opts) => asStream(Object.assign({}, normalizeArguments(url, opts), {stream: true}));
```
sindresorhus: Would be better to do the `Object.assign` thing inside the `asStream` function.

test/progress.js
```diff
@@ -0,0 +1,197 @@
...
+	t.true(events.length >= 2);
+
+	const hasBodySize = typeof bodySize === 'number';
+	let lastEvent = events.shift();
+
+	if (!hasBodySize) {
+		t.is(lastEvent.percent, 0);
+	}
+
+	events.forEach((event, index) => {
```
sindresorhus:
```js
for (const [index, event] of events.entries()) {
```

test/progress.js
```diff
@@ -0,0 +1,197 @@
...
+	const req = got.stream.post(`${s.url}/upload`)
+		.on('uploadProgress', e => events.push(e));
+
+	intoStream(file)
+		.pipe(req)
+		.on('data', () => {})
+		.on('end', () => {
+			checkEvents(t, events);
+			t.end();
+		});
```
sindresorhus: You could use https://github.com/sindresorhus/get-stream here to be able to use async/await instead of `test.cb`.

test/progress.js
```diff
@@ -0,0 +1,197 @@
...
+});
+
+test('upload progress - form data', async t => {
+	const events = [];
+
+	const body = new FormData();
+	body.append('key', 'value');
+	body.append('file', file);
+
+	const size = await pify(body.getLength.bind(body))();
```
sindresorhus: `pify` binds to the original object by default, so the bind here is moot I think.

vadimdemedes: Nope, it fails without bind.

index.js
vadimdemedes: We just picked a random number for frequency of upload progress events.

vadimdemedes:
Thanks @sindresorhus, feedback addressed. Had to force push, because of a conflict.

index.js
sindresorhus: Should definitely be clearer in the code then. 150 is very specific, wonder why we didn't just go with 100.

sindresorhus:
@vadimdemedes Thanks for the quick turn-around. You missed: https://github.com/sindresorhus/got/pull/322#discussion_r125171040

sindresorhus:
Can you apply https://github.com/sindresorhus/got/pull/322#discussion_r125171040 to the other `test.cb` tests too?

vadimdemedes:
@sindresorhus Oh yeah, missed the other ones.

index.js
vadimdemedes: Np ;)

readme.md
```diff
@@ -202,6 +203,21 @@ got.stream('github.com')
 
 `redirect` event to get the response object of a redirect. The second argument is options for the next request to the redirect location.
 
+##### .on('uploadProgress', progress)
+##### .on('downloadProgress', progress)
```
kevva: Should probably clarify that these can be used with the `Promise` interface too. It's a little ambiguous now since they're documented under `got.stream`.

sindresorhus:
@AlexTes @floatdrop I'll merge this on monday unless it gets any objections by then ;)

index.js
```diff
@@ -3,27 +3,64 @@ const EventEmitter = require('events');
...
 const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);
 
+const isFormData = body => isStream(body) && typeof body.getBoundary === 'function';
+
+const getBodySize = opts => {
+	const body = opts.body;
+
+	if (opts.headers['content-length']) {
+		return Number(opts.headers['content-length']);
```
SamVerschueren: Just generally interested what the benefit is of wrapping this in a `Number` object?

readme.md
```diff
@@ -202,6 +203,21 @@ got.stream('github.com')
...
+
+Progress events for uploading (sending request) and downloading (receiving response). The `progress` argument is an object like:
+
+```js
+{
+	percent: 0.24,
```
SamVerschueren: `1024` bytes send against `10240` in total means 10% and thus `percent` should be `0.1`.

index.js
```diff
@@ -46,7 +85,17 @@ function requestAsEventEmitter(opts) {
 			fn = electron.net || electron.remote.net;
 		}
 
+		let progressInterval;
+
 		const req = fn.request(opts, res => {
+			clearInterval(progressInterval);
+
+			ee.emit('uploadProgress', {
+				percent: 1,
```
SamVerschueren: Is `percent` emitted because `total` can be `null` when streaming?

index.js
sindresorhus: It's not wrapping it in a Number object, it's coercing it to a number.

SamVerschueren: Any benefit over parseInt? I never used Number directly so that's why I'm asking :).

sindresorhus: Mostly just consistency. I also use Boolean() to coerce something to a Boolean. Same with String().

index.js
vadimdemedes: > Why is `percent` set to 1? Probably missing something here :).

It's calculating `percent` to be from `0` to `1`. So `1` means it's the last upload progress event and it's done.

> I'm also wondering why it is emitted?

Initially it was only for convenience. But there are also cases where `total` can be unavailable (like you said, when streaming), so got makes sure that `percent` is always available.

sindresorhus:
Yay! Vadim, your sambuca-powered work is finally merged. Thanks for working on it. Good times.

leo-excited

sindresorhus:
@floatdrop @AlexTes Feel free to add your review still. I'm not gonna do a release yet.

index.js
```diff
@@ -111,7 +203,44 @@ function requestAsEventEmitter(opts) {
...
+
+			req.connection.on('connect', () => {
+				const uploadEventFrequency = 150;
+
+				progressInterval = setInterval(() => {
+					const lastUploaded = uploaded;
+					const headersSize = Buffer.byteLength(req._header);
+					uploaded = req.connection.bytesWritten - headersSize;
+
+					// Prevent the known issue of `bytesWritten` being larger than body size
```
AlexTes: @vadimdemedes if you have a link to an issue explaining what exactly is amiss and how simply ignoring any bytes uploaded beyond the expected body size is the right solution. Do share, I'd be happy to add it to the comment. If you don't let me know and I'll go looking myself.

index.js
```diff
@@ -111,7 +203,44 @@ function requestAsEventEmitter(opts) {
...
 
+		ee.on('request', req => {
+			ee.emit('uploadProgress', {
+				percent: 0,
+				transferred: 0,
+				total: uploadBodySize
+			});
+
+			req.connection.on('connect', () => {
+				const uploadEventFrequency = 150;
```
AlexTes: Just curious, is there a reason behind this number? 6 2/3 times a second?

index.js
```diff
@@ -111,7 +203,44 @@ function requestAsEventEmitter(opts) {
...
+
+				progressInterval = setInterval(() => {
+					const lastUploaded = uploaded;
+					const headersSize = Buffer.byteLength(req._header);
```
AlexTes: Can header size change during upload? Otherwise, I don't see the point of putting this inside the interval or even the on 'connect'.

vadimdemedes: It can't change during upload, but there's no way to know when headers are sent, is there? So I placed it here to correctly calculate the body size.

index.js
vadimdemedes: No reason. It was just a guess on how to report progress events not too frequently, but still enough.

index.js
vadimdemedes: There's no explanation, I just observed this bug randomly and couldn't find a source of the issue.

vadimdemedes:
@sindresorhus haha, best times!

I don't know why, but when I searched "sambucca" on Giphy, this is what it gave me, so now I'm posting this instead of my favorite drink to celebrate the merge:

marianoguerra:
is there any plan to make a release with this change?

sindresorhus:
Yes

PR #323, opened 2017-06-01
Fix ParseError being thrown for 304

djmadeira:
Noticed while scanning the code for an unrelated issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -164,21 +164,26 @@ function asPromise(opts) {
 				.then(data => {
 					const statusCode = res.statusCode;
 					const limitStatusCode = opts.followRedirect ? 299 : 399;
+					let error;
 
 					res.body = data;
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
> I think I'll kick this back over to the issue for further discussion.

@AlexTes Which issue?

sindresorhus:
@AlexTes :arrow_up:

AlexTes:
@sindresorhus upon re-review I suggest we just close this. If we get a body on a 304 response (wrong behavior), and the user has requested json parsing of the response body, we do not throw in case we fail to parse the body (arguably wrong behavior).

Until someone wants to discuss this (@djmadeira ?) I think there's no reason to change anything. I think in this PR it was assumed 304's could have bodies that got should be parsing, they should not but in case anyone feels we should parse them regardless throw when failing they should open an issue to discuss it first. I for one disagree with that approach but would happily discuss 😄.

PR #329, opened 2017-06-08
`url` arg optimizations

stevenvachon:


kevva:
LGTM 👍

AlexTes:
Neat. Now that I look at it again the way I wrote it is silly.

One note. This would in a sense be a breaking change. `url-to-options` turns the `username` (part of auth) in the url into `${username}:`. In other words, if you set the `username` in a URL to `user`. and create a WHATWG-URL, the submitted authorization request header will be `Basic dXNlcjo=` which decodes as `user:`. It appends a colon. Without reading the spec, looking at `curl` and Node, this is correct in the sense that `curl` does this and Node clearly does this with WHATWG-URL too. So when using `url-to-options` the outgoing request will be the exact same as when you'd use Node with WHATWG-URL. However, currently, this is not the case. When getting the full URL through `href` the WHATWG-URL serializes to `http://user@httpbin.org/get`. We currently parse that, which results in the old Node behavior which does not append the colon, and thus results in a different header.

I imagine there may be other edge-cases too. I'm guessing @stevenvachon, this one slipped by and you are not aware of any others? Would be good to know otherwise 😅.
Although it is a breaking change, it's also a bugfix. Not sure how to handle this.

stevenvachon:
@AlexTes These URLs are the same:
```
http://user@host/
http://user:@host/
```

AlexTes:
@stevenvachon well like I said. According to the spec, yes, but in practice no. One doesn't have a colon and encodes to a different base64 string with Node even if it clearly shouldn't. ~~I would bet Node can't fix that due to backward compatibility.~~ Just looked, can't find any such issue. I'll open one. RFC 2617 is pretty clear about the client having to add that ':'.

Regardless, we need to make a small choice. This is a breaking change. Previously WHATWG-URLs would wrongly get Node to send without the colon. With this PR they will again. So I'm just saying if we're being strict this means new major. But I also imagine practically no one is relying on WHATWG-URLs getting basic authentication right with Got 7. So maybe minor or even patch. I'm inexperienced is this regard.

stevenvachon:
We're using `url-to-options` which was copied from Node.js source. I think that all of this should be fixed upstream.

AlexTes:
I see.
[nodejs/node #13571](https://github.com/nodejs/node/issues/13571).

stevenvachon:
We should likely remove `url.parse()` completely from this library, as I'd originally requested in #228

stevenvachon:
Anyway, @sindresorhus please review.

PR #333, opened 2017-06-13
Add 'br' encoding

silverwind:
Noticed that the greatest encoding out there was missing from this list.

### DIFF
--- a/index.js
+++ b/index.js
@@ -89,7 +89,8 @@ function requestAsEventEmitter(opts) {
 					typeof decompressResponse === 'function' &&
 					req.method !== 'HEAD' ? decompressResponse(res) : res;
 
-				if (!opts.decompress && ['gzip', 'deflate'].indexOf(res.headers['content-encoding']) !== -1) {
+				const encodings = ['br', 'deflate', 'gzip'];
+				if (!opts.decompress && encodings.indexOf(res.headers['content-encoding']) !== -1) {
 					opts.encoding = null;
 				}
 


silverwind:
Also, according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding, there's a possibilty for `compress` encoding, but I don't think this is worth adding here.

kevva:
Is it worth adding a test for this maybe?

silverwind:
That would add a devDep on https://github.com/mayhemydg/iltorb, which requires a native module. I don't think it's worth that as gzip is already tested.

kevva:
How do we decompress the `br` content? Seems like `zlib` [doesn't](https://nodejs.org/api/zlib.html#zlib_class_zlib_unzip).

silverwind:
> How do we decompress the br content?

iltorb can both compress and decompress.

kevva:
Yeah, but I mean when using `got` and you try to fetch `br` encoded content, it wouldn't be decompressed if `decompress` is `true`?

silverwind:
Oh, right. Looks like https://github.com/sindresorhus/decompress-response is lacking support for `br`. I guess it should be added there first.

kevva:
The module I linked further up should handle it without native deps I think, https://github.com/devongovett/brotli.js.

silverwind:
Haven't seen that one yet, but iltorb is well maintained and I haven't had any issues with it yet, even under Windows. Brotli compression is very CPU intensive, so I'd definitely compare the performance before choosing a module.

kevva:
Yup, I'll work on a PR, but I'm willing to sacrifice performance over native deps tbh depending on the impact.

silverwind:
Opened https://github.com/sindresorhus/decompress-response/issues/12 which blocks this issue.

sindresorhus:
https://github.com/sindresorhus/decompress-response/pull/13#issuecomment-331626197

sindresorhus:
Node.js issue about adding it natively: https://github.com/nodejs/node/issues/18964

PR #335, opened 2017-06-15
Support Array with the `json` option

### LINKED ISSUE #334: Support posting Arrays with json option
Eventhough it was mentioned in #174:

> I'm ok with also supporting an array

Arrays are not supported with the `json` option (because of `isPlainObject`) so this fails:
```js
got.post('example.com', {
  json: true,
  body: [1, 2, 3]
})
```
Could you use something like [lodash's `isObjectLike`](https://lodash.com/docs#isObjectLike) instead of `isPlainObject` or test both `isPlainObject || isArray`?

DerTieran:
Allow to post Arrays with the `json` option.
This was discussed in [#174](https://github.com/sindresorhus/got/issues/174#issuecomment-299534498) but not implemented.

> I'm ok with also supporting an array

Fixes #334

### DIFF
--- a/index.js
+++ b/index.js
@@ -315,17 +315,18 @@ function normalizeArguments(url, opts) {
 			throw new TypeError('options.body must be a ReadableStream, string, Buffer or plain Object');
 		}
 
-		if ((opts.form || opts.json) && !(isPlainObj(body) || Array.isArray(body))) {
+		const canBodyBeStringified = isPlainObj(body) || Array.isArray(body);
+		if ((opts.form || opts.json) && !canBodyBeStringified) {
 			throw new TypeError('options.body must be a plain Object or Array when options.form or options.json is used');
 		}
 
 		if (isStream(body) && typeof body.getBoundary === 'function') {
 			// Special case for https://github.com/form-data/form-data
 			headers['content-type'] = headers['content-type'] || `multipart/form-data; boundary=${body.getBoundary()}`;
-		} else if (opts.form && (isPlainObj(body) || Array.isArray(body))) {
+		} else if (opts.form && canBodyBeStringified) {
 			headers['content-type'] = headers['content-type'] || 'application/x-www-form-urlencoded';
 			opts.body = querystring.stringify(body);
-		} else if (opts.json && (isPlainObj(body) || Array.isArray(body))) {
+		} else if (opts.json && canBodyBeStringified) {
 			headers['content-type'] = headers['content-type'] || 'application/json';
 			opts.body = JSON.stringify(body);
 		}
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -315,8 +315,10 @@ function normalizeArguments(url, opts) {
 			throw new TypeError('options.body must be a ReadableStream, string, Buffer or plain Object');
 		}
 
-		if ((opts.form || opts.json) && !isPlainObj(body)) {
-			throw new TypeError('options.body must be a plain Object when options.form or options.json is used');
+		if (opts.form && !isPlainObj(body)) {
+			throw new TypeError('options.body must be a plain Object when options.form is used');
+		} else if (opts.json && !(isPlainObj(body) || Array.isArray(body))) {
```
kevva: Don't need a `else if` here, just do a new `if` block.

DerTieran: True, changed it.
*(Hope you guys are okay with amending the last commit)*

index.js
```diff
@@ -315,8 +315,12 @@ function normalizeArguments(url, opts) {
 			throw new TypeError('options.body must be a ReadableStream, string, Buffer or plain Object');
 		}
 
-		if ((opts.form || opts.json) && !isPlainObj(body)) {
```
AlexTes: I'd prefer if we kept this joint, it's simpler. Let's discuss it over in the issue.

test/post.js
```diff
@@ -121,7 +129,7 @@ test('content-type header is not overriden when object in options.body', async t
 	t.is(headers['content-type'], 'doge');
 });
 
-test('throws when json body is not a plain object', async t => {
+test('throws when json body is not a plain object or array', async t => {
```
AlexTes: If we do choose to split the argument validation for `json` from `form` we should also have two tests to make sure both work the way we expect.

index.js
```diff
@@ -315,17 +315,17 @@ function normalizeArguments(url, opts) {
...
 
 		if (isStream(body) && typeof body.getBoundary === 'function') {
 			// Special case for https://github.com/form-data/form-data
 			headers['content-type'] = headers['content-type'] || `multipart/form-data; boundary=${body.getBoundary()}`;
-		} else if (opts.form && isPlainObj(body)) {
+		} else if (opts.form && (isPlainObj(body) || Array.isArray(body))) {
 			headers['content-type'] = headers['content-type'] || 'application/x-www-form-urlencoded';
 			opts.body = querystring.stringify(body);
-		} else if (opts.json && isPlainObj(body)) {
+		} else if (opts.json && (isPlainObj(body) || Array.isArray(body))) {
```
sindresorhus: You're repeating `(isPlainObj(body) || Array.isArray(body))` 3 times. That should be DRY'd up.

sindresorhus:
The docs needs to be updated to reflect this change.

index.js
DerTieran: Sure something like `isPlainObjOrArray`?

sindresorhus: `isValidBody` or something. Better to describe the intent than the contents.

PR #336, opened 2017-06-16
Add all helper functions in the same loop

lukechilds:
Just noticed the HTTP method helper functions are added to `got` in a forEach loop and then again to `got.stream` in a for in loop a few lines down. They might as well be added in the same loop.

### DIFF
--- a/index.js
+++ b/index.js
@@ -388,7 +388,9 @@ function got(url, opts) {
 	}
 }
 
-const helpers = [
+got.stream = (url, opts) => asStream(normalizeArguments(url, opts));
+
+const methods = [
 	'get',
 	'post',
 	'put',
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -388,25 +388,20 @@ function got(url, opts) {
...
 	'post',
 	'put',
 	'patch',
 	'head',
 	'delete'
-];
-
-helpers.forEach(el => {
-	got[el] = (url, opts) => got(url, Object.assign({}, opts, {method: el}));
+].forEach(method => {
```
kevva: Use `for...of` instead.

lukechilds: @kevva Done 👍

Why is for of preferable over forEach?

sindresorhus: Personal preference, consistency, and it's faster.

PR #339, opened 2017-06-24
Fix unix address parser

Dreamacro:
When request
```
unix:/var/run/docker.sock:/images/alpine:latest/json
```
It will return socketPath `unix:/var/run/docker.sock:/images/alpine`

so fix regex to `(.+?):(.+)`

### DIFF
--- a/index.js
+++ b/index.js
@@ -342,7 +342,7 @@ function normalizeArguments(url, opts) {
 	}
 
 	if (opts.hostname === 'unix') {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

Dreamacro:
Sorry, create additional test cases now.

sindresorhus:
Thanks :)

PR #347, opened 2017-07-21
Error should be called `err` for consistency

lukechilds:
Just spotted this.


lukechilds:
@sindresorhus Any idea why this didn't get picked up by XO in `unicorn/catch-error-name`?

sindresorhus:
@lukechilds https://github.com/sindresorhus/eslint-plugin-unicorn/issues/107

PR #348, opened 2017-07-25
Use conventional spelling of "cancellation"

josephfrazier:
This makes it a little easier to read without doing a double-take.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -18,7 +18,7 @@ Created because [`request`](https://github.com/request/request) is bloated *(sev
 ## Highlights
 
 - [Promise & stream API](#api)
-- [Request cancelation](#aborting-the-request)
+- [Request cancellation](#aborting-the-request)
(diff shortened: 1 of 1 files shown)


sindresorhus:
https://github.com/sindresorhus/p-cancelable#cancelable-vs-cancellable

josephfrazier:
Thanks for the link. FWIW, the site it links to also mentions:

> The spelling distinction ... does not extend to *cancellation*, which everywhere is spelled with two *l*’s.

Language is weird, but I understand if you still want to keep it consistent with canceled/canceling/cancelable :man_shrugging:

sindresorhus:
https://en.m.wiktionary.org/wiki/cancelation

sindresorhus:
In the end I just prefer consistency.

josephfrazier:
That makes sense :) Thanks for all your work!

Issue #266, opened 2017-01-22
const pkg = require('./package');


alanzhaonys:
v6.7.1

const pkg = require('./package');

this line is causing issue. "package" is not defined in package.json

ERROR in ./~/is-online/~/is-reachable/~/got/index.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./package in /Users/alan/Desktop/Electron/app/node_modules/is-online/node_modules/is-reachable/node_modules/got
resolve file
  /Users/alan/Desktop/Electron/app/node_modules/is-online/node_modules/is-reachable/node_modules/got/package doesn't exist
  /Users/alan/Desktop/Electron/app/node_modules/is-online/node_modules/is-reachable/node_modules/got/package.js doesn't exist
  /Users/alan/Desktop/Electron/app/node_modules/is-online/node_modules/is-reachable/node_modules/got/package.jsx doesn't exist
resolve directory

sindresorhus:
`package` is `package.json`. Extensions are optional.

hugohil:
*I leave that here because it can help people facing similar issue.*

I was having an issue with the `got` module (through `vue-resource`) like this:
```
ERROR in ./~/got/index.js
Module not found: Error: Can't resolve './package' in '/path/to/project/node_modules/got'
```

I found that the issue was caused by webpack not resolving the extension automatically, so I added this to my `webpack.config.js` file:

```
resolve: {
    extensions: ['.js', '.json']
}
```

so now I can omit the `.json` extension when I `require()` a module.

GabrielDuarteM:
@hugohil Sir, you helped me a lot with this, i was having this './package' error, and your fix solved it!

spiritOfSasidhar:
> _I leave that here because it can help people facing similar issue._
> 
> I was having an issue with the `got` module (through `vue-resource`) like this:
> 
> ```
> ERROR in ./~/got/index.js
> Module not found: Error: Can't resolve './package' in '/path/to/project/node_modules/got'
> ```
> 
> I found that the issue was caused by webpack not resolving the extension automatically, so I added this to my `webpack.config.js` file:
> 
> ```
> resolve: {
>     extensions: ['.js', '.json']
> }
> ```
> 
> so now I can omit the `.json` extension when I `require()` a module.
> 
> Hope this helps someone :)

i am facing same issue, but i didnt have webpack file in my project, so please give appropriate solution

PR #349, opened 2017-07-26
add `.json` to require('./package')

janpio:
This fixes issues for people using `got` with webpack, which by default has problems with the missing extension. 

I know that [extensions are optional](https://github.com/sindresorhus/got/issues/266#issuecomment-274361530), but this simple change shouldn't cause any problems but would resolve the problem of all wbepack users trying to use `got` (or libraries using `got`).

I agree that normally libraries shouldn't have to adapt to the ecosystem, but in this case the code becomes even more readable and clearer (I at least didn't know that you could just leave off the file extension.)

Why not just use the workaround? Well, it seems to [not work in all cases](https://github.com/ionic-team/ionic-app-scripts/issues/1148) - and I found one of them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ const isPlainObj = require('is-plain-obj');
 const PCancelable = require('p-cancelable');
 const pTimeout = require('p-timeout');
 const pify = require('pify');
-const pkg = require('./package');
+const pkg = require('./package.json');
 
 const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);


kevva:
There are plenty of closed issues about this already https://github.com/sindresorhus/got/search?q=package+json+webpack&type=Issues&utf8=%E2%9C%93. If the suggested solution doesn't work, I'd say it's something wrong with your configuration, or in `ionic-app-scripts`. This seems like it could cause it https://github.com/ionic-team/ionic-app-scripts/blob/master/config/optimization.config.js#L13-L15.

janpio:
Fair enough.

But let me ask, what is so important with using the shortcut version without the filename? Is there a advantage?

janpio:
Thanks for the pointer to the `optimization.config.js` - I wasn't aware of that and now created a PR to fix it.

kevva:
> But let me ask, what is so important with using the shortcut version without the filename? Is there a advantage?

There's no real reason to include the extension because of Webpack not being compatible with Node.js.

janpio:
Yeah but why not just add these 5 tiny characters that don't change a thing for you **and** you get rid of all these issues  about using got with webpack? Also the code is more readable for people that are not aware that extensions are optional...

PR #350, opened 2017-07-28
Fix package

rodrigooler:
Improving package organization

- [x] Fix package.json


sindresorhus:
No thanks, the current package ordering is exactly how we want it.

PR #351, opened 2017-07-28
Add prettier

rodrigooler:
I believe the prettier can improve the project code so I put it in the development dependencies.

- [x] Add husky;
- [x] Add lint-staged;
- [x] Add prettier;
- [x] Run prettier in index.js;
- [ ] Fix errors;


sindresorhus:
Thanks for the suggestion, but we're already using [XO](https://github.com/sindresorhus/xo) for linting and code style. I might [integrate Prettier into XO](https://github.com/sindresorhus/xo/issues/176) itself at some point.

It's usually best to open an issue before doing changes like these as code style is very subjective.

rodrigooler:
@sindresorhus I understand, in the next suggestions I will open. tks

PR #360, opened 2017-08-19
Fix cancel when timeout exists - fixes #344

### LINKED ISSUE #344: Cannot cancel when timeout option is set
When the options.timeout, or options.timeout.request, is set, the `pCancelable` promise is wrapped in a `pTimeout` promise. This prevents the possibility of calling cancel.

See [index.js#L130-L132](https://github.com/sindresorhus/got/blob/master/index.js#L130-L132).

Is this intentional?

jotto:
Fixes https://github.com/sindresorhus/got/issues/344

### DIFF
--- a/index.js
+++ b/index.js
@@ -270,7 +270,7 @@ function asPromise(opts) {
 
 	const proxy = new EventEmitter();
 
-	const promise = timeoutFn(new PCancelable((onCancel, resolve, reject) => {
+	const cancelable = new PCancelable((onCancel, resolve, reject) => {
 		const ee = requestAsEventEmitter(opts);
 		let cancelOnRequest = false;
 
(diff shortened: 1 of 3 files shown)


jotto:
I really just wanted cancel fixed so I could do this:

```javascript
const got = require("got");
const req = got("http://localhost:3000", { retries: 0, timeout: 100 });
req.catch(err => {
  if (err.name === "RequestError" && err.code === "ETIMEDOUT") {
    console.log("request timed out, canceling");
    return req.cancel();
  } else {
    return Promise.reject(err);
  }
});
```

so I think I agree from the comments in #344 that a timeout should automatically call cancel, I just didn't want to make any behavior changes in the interest of getting this through sooner than later.

sindresorhus:
I'm ok with this workaround for now. Thanks :)

Issue #371, opened 2017-09-11
Why are you throwing on status 503?


MySidesTheyAreGone:
`got` gives me an error and no response body when a server answers with a status code 503. This is the case whenever a page is protected by Cloudflare. `request` gives me a 503 status as well, but it also correctly downloads the response body and I can then go on to the real page. With `got` I can't do anything.

A page you can use to test this would be https://torrents.me - the first one I found that always serves Cloudflare's page on the first visit (clear your cookies if you want to see it again). You can try to `got` it - it will throw - and then `request` it - it won't and it will give you the body.

I need `got` to go on and download the response body. As long as it doesn't I can't switch to it (I would like very much to do that). I also believe `got` should be as capable as the browser for all things that are about getting webpages (whereas I think it's OK to leave the handling of cookies and other extra stuff to other modules).

Thank you for reading!

kevva:
The body is still available on `err.response.body`:

```
got('http://service.unavailable').catch(err => {
    console.log(err.response.body);    
});
```

sindresorhus:
Got rejects the promise for anything that is not a success (unlike `request`). We think it's better to handle unsuccessful requests in an alternative code-path.

MySidesTheyAreGone:
Oh sorry guys, I see it's in the docs after all; I missed it. Thanks!

MySidesTheyAreGone:
Ahem, do you do that in stream mode too? Because I'm not seeing it. Here's all the keys on the error object:

```
[ 'name',
  'host',
  'hostname',
  'method',
  'path',
  'protocol',
  'url',
  'statusCode',
  'statusMessage',
  'headers' ]
```
I use streaming so I can abort if the download size is too large.

PR #372, opened 2017-09-13
Set the response body on a HTTPError

rhodgkins:
This allows the end-user to parse any response body that might be describing the HTTP error.


kevva:
https://github.com/sindresorhus/got/issues/371#issuecomment-328633307

rhodgkins:
Ah, I missed that - cheers! Did think it was odd it wasn't there already!

kevva:
No worries :). It's easy to overlook sometimes, but it's in the [usage](https://github.com/sindresorhus/got#usage) section.

PR #376, opened 2017-09-20
Document how to test for cancelation

dinoboff:
Add example showing how the catch cancellation error.

Note that it requires the catch handler to have access to the cancellable promise. Exposing the CancelError to check the Error instance type would be useful.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -280,20 +280,20 @@ When given an unsupported protocol.
 
 ## Aborting the request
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks :)

sindresorhus:
> Exposing the CancelError to check the Error instance type would be useful.

PR welcome

PR #386, opened 2017-10-06
Support an object in the `agent` option

### LINKED ISSUE #365: Custom agent + redirects
I'm seeing the following error:
```
Uncaught error: Error: Protocol "https:" not supported. Expected "http:" 
         at get (/var/app/current/node_modules/got/index.js:33:18) 
         at ClientRequest.fn.request.res (/var/app/current/node_modules/got/index.js:51:5)
```

I think this is caused by passing in a custom agent (from `agentkeepalive`).
1. HTTP and HTTPS require different agent instances
2. If the initial request is HTTP we pass in the HTTP agent
3. If this request then redirects to an HTTPS url, got will still use the HTTP custom agent

Is there a way that we could pass in both an HTTP and HTTPS agent and let got handle the selection of the proper agent? Or alternatively to intercept the redirect and change the agent that will be used?

pietermees:
Fixes #365

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -410,7 +410,7 @@ const got = require('got');
 const HttpAgent = require('agentkeepalive');
 const HttpsAgent = HttpAgent.HttpsAgent;
 
-got('todomvc.com', {
+got('sindresorhus.com', {
 	agent: {
 		http: new HttpAgent(),
 		https: new HttpsAgent()
--- a/readme.md
+++ b/readme.md
@@ -403,6 +403,21 @@ got('todomvc.com', {
 });
 ```
 
+If you require different agents for different protocols, you can pass in all of them through the `agents` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
+
+```js
+const got = require('got');
+const HttpAgent = require('agentkeepalive')
+const HttpsAgent = HttpAgent.HttpsAgent
+
+got('todomvc.com', {
+	agents: {
+		'http:': new HttpAgent(),
+		'https:': new HttpsAgent()
+	}
+});
+```
+
 
 ## Cookies
 
--- a/readme.md
+++ b/readme.md
@@ -403,17 +403,17 @@ got('todomvc.com', {
 });
 ```
 
-If you require different agents for different protocols, you can pass in all of them through the `agents` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
+If you require different agents for different protocols, you can pass a map of agents to the `agent` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
 
 ```js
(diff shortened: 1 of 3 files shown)


pietermees:
Not sure why the cancel tests fail on Node 6, but that seems to be a general problem that randomly occurred before?

readme.md
```diff
@@ -334,6 +334,21 @@ got('todomvc.com', {
...
+
+```js
+const got = require('got');
+const HttpAgent = require('agentkeepalive')
+const HttpsAgent = HttpAgent.HttpsAgent
+
+got('todomvc.com', {
+	agents: {
+		'http:': new HttpAgent(),
+		'https:': new HttpsAgent()
```
sindresorhus: I think the option names should be without `:` and unquoted.

readme.md
```diff
@@ -334,6 +334,21 @@ got('todomvc.com', {
...
 
+If you require different agents for different protocols, you can pass in all of them through the `agents` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
+
+```js
+const got = require('got');
+const HttpAgent = require('agentkeepalive')
+const HttpsAgent = HttpAgent.HttpsAgent
+
+got('todomvc.com', {
+	agents: {
```
sindresorhus: Better to just overload the `agent` option. It's confusing to have two so similarly named options.

readme.md
```diff
@@ -334,6 +334,21 @@ got('todomvc.com', {
 });
 ```
 
+If you require different agents for different protocols, you can pass in all of them through the `agents` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
+
+```js
+const got = require('got');
+const HttpAgent = require('agentkeepalive')
+const HttpsAgent = HttpAgent.HttpsAgent
```
sindresorhus: Semi-colon

pietermees:
@sindresorhus updated in response to your review

readme.md
```diff
@@ -334,6 +334,21 @@ got('todomvc.com', {
...
 
+If you require different agents for different protocols, you can pass a map of agents to the `agent` option. This is necessary because a request to one protocol might redirect to another. In such a scenario, `got` will switch over to the right protocol agent for you.
+
+```js
+const got = require('got');
+const HttpAgent = require('agentkeepalive');
+const HttpsAgent = HttpAgent.HttpsAgent;
+
+got('todomvc.com', {
+	agents: {
```
sindresorhus: `agents` => `agent`

timdp:
@sindresorhus Can you retry the Node 6 test? Seems unrelated to this PR. Thanks!

pietermees:
Rebased & tests seem to be passing now

pietermees:
@sindresorhus added test case, I think this might actually be the first test coverage for the `agent` option

sindresorhus:
Excellent. Thank you @pietermees :)

PR #391, opened 2017-10-15
Add `request` and `response` events to the Promise API

jgdev:
Using got as Promise, I can't get the request object, eg:

```
got('https://jgdev.info/api/events', {
  method: "GET",
  json: true,
  timeout: 15000,
}).on('request', req => console.log(req));
```

This code isn't working, because the proxy event "request" is not being called.


sindresorhus:
What exactly is this fixing? Can you add a test?

jgdev:
I just update the description, I will be updating the PR with the test as soon possible.

sindresorhus:
They're not really missing, as they're not documented to be part of the Promise interface. The readme should also be updated then. We document these already, so I guess we could just move them out of the `Streams` section and into a section for `Promise & Stream API events` or something.

Out of curiosity, what's your use-case for needing these events on the Promise API?

jgdev:
Just for debugging purposes.
Why isn't part of Promise interface?

kevva:
> Why isn't part of Promise interface?

Because events are mostly used when working with streams.

jgdev:
OK, but I don't have any way to get the request using `json:true`?

sindresorhus:
> Why isn't part of Promise interface?

Mostly because we never saw any use-case for it.

I'm ok with adding this, however.

sindresorhus:
@jgdev Interested in finishing this up? :)

jgdev:
Yeah, I'll finish the test today.

sindresorhus:
Ping :)

jgdev:
I don't know exactly what is the best way to test these events, I just test if the events are beign called.

Another way maybe is matching the class constructor name of Request and Response objects?

index.js
```diff
@@ -338,6 +340,7 @@ function asPromise(opts) {
 						throw new got.HTTPError(statusCode, res.statusMessage, res.headers, opts);
 					}
 
+					proxy.emit('response', res);
```
sindresorhus: Shouldn't this be called right (on line 319) away, like the stream version, instead of waiting for the body to be fetched?

sindresorhus:
This still needs documentation, as mentioned earlier.

index.js
jgdev: Yeah, you right let me remove it.

sindresorhus:
@jgdev Still interested in finishing this up? :)

sindresorhus:
Closing for lack of response. @jgdev Happy to reopen if you ever want to finish it up :)

PR #393, opened 2017-10-16
Set AVA concurrency to 4

lukechilds:
Otherwise `os.cpus().length` will return the CPU count of the host machine not the Docker container in Travis.

[Builds consistently timeout with no concurrency set](https://travis-ci.org/lukechilds/got/builds/288678445) and consistently pass with concurrency set as 4.

You can see in [here](https://travis-ci.org/lukechilds/got/jobs/288682177) `os.cpus().length` outputs 32 on Travis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,9 @@
 		"universal-url": "1.0.0-alpha",
 		"xo": "^0.18.0"
 	},
+	"ava": {
+		"concurrency": 4
+	},
 	"browser": {
 		"decompress-response": false,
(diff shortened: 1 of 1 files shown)


lukechilds:
In fact it might be better to get this resolved directly in AVA as this will affect a lot of other projects running tests in containers.

Submitted a PR: https://github.com/avajs/ava/pull/1551

sindresorhus:
Yes, we should fix this in AVA.

PR #395, opened 2017-10-25
Add package.json extension

tmos:
When used with rollup.js, the absence of the file extension makes a warning and create the need to add a rollup plugin.
It may not be perceived well to modify Got to fit with a tierce lib, but I don't think it is a very big change :)

Regards,
Tom


sindresorhus:
Duplicate of https://github.com/sindresorhus/got/pull/349

PR #398, opened 2017-10-30
Use statusMessage returned by endpoint

sleewoo:
for cases when endpoints returns a descriptive error message

### DIFF
--- a/index.js
+++ b/index.js
@@ -629,7 +629,7 @@ got.ParseError = class extends StdError {
 got.HTTPError = class extends StdError {
 	constructor(statusCode, statusMessage, headers, opts) {
 		if (statusMessage) {
-			statusMessage = statusMessage.replace(/\n|\r/gm, ' ');
+			statusMessage = statusMessage.replace(/\r?\n/g, ' ').trim();
 		} else {
 			statusMessage = http.STATUS_CODES[statusCode];
 		}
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you include a test?

index.js
```diff
@@ -627,8 +627,12 @@ got.ParseError = class extends StdError {
 };
 
 got.HTTPError = class extends StdError {
-	constructor(statusCode, headers, opts) {
-		const statusMessage = http.STATUS_CODES[statusCode];
+	constructor(statusCode, statusMessage, headers, opts) {
+		if (statusMessage) {
+			statusMessage = statusMessage.replace(/\n|\r/gm, ' ');
```
sindresorhus: `/\n|\r/gm` => `/\r?\n/g`

sleewoo: yep, much better

sleewoo:
cranking a test

lukechilds:
I support this change but just wanna point out potential breakage.

Any existing user code like:

```js
if (err.statusMessage === 'Too Many Requests') {
  handleAPILimit();
}
```

will fail after this change if the API returns it's own error message.

sindresorhus:
@lukechilds I'll make sure to point that out in the release notes.

sindresorhus:
Thanks @sleewoo :)

PR #402, opened 2017-10-31
[documentation] correct User-Agent capitalisation

revelt:
Currently we have `user-agent`. It should be `User-Agent`, as per [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -529,7 +529,7 @@ request(`https://${config.host}/production/`, {
 
 ### User Agent
 
-It's a good idea to set the `'user-agent'` header so the provider can more easily see how their resource is used. By default, it's the URL to this repo.
+It's a good idea to set the `'User-Agent'` header so the provider can more easily see how their resource is used. By default, it's the URL to this repo.
 
 ```js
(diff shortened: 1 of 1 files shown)


sindresorhus:
If you actually read the spec (not just the MDN page), you'll see:

>  The type, subtype, and parameter name tokens are case-insensitive.

The keys are case-insensitive, so lowercase is totally valid.

revelt:
aha! thanks anyway

revelt:
I got `request` working and can't get `got` working to fetch simple data from GitHub. The GitHub API returns `null` so I'm checking what is wrong.

Currently, simplest thing copied from documentation is not working:

```
function getUserInfo(username) {
  return got(`https://api.github.com/users/${username}`, {
    headers: {
      'user-agent': 'Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0',
    },
  })
```

I tried all user agents (like legit-ones above), no luck.

Issue #399, opened 2017-10-30
Socket recycling causes memory leaks


timdp:
If sockets are recycled, in our case because we're using [`agentkeepalive`](https://www.npmjs.com/package/agentkeepalive) as the agent, `got`'s listeners don't always get removed from the socket after it's done its thing. We validated this by instrumenting [the `connect` listener](https://github.com/sindresorhus/got/blob/1b8c1d6cc2981dffd12babf48ae8a2431bfbf619/index.js#L219) to store a unique ID on every new request ...

```js
// At the top
let count = 0

// In requestAsEventEmitter()
req.connection.on('connect', () => {
  req.connection.__id = req.connection.__id || ++count
  console.log('connect', req.connection.__id)
```

... and then firing a few requests in sequence using the same socket:

```js
const got = require('got')
const Agent = require('agentkeepalive')

const agent = new Agent({timeout: 100, maxSockets: 1})

;(async () => {
  for (let i = 0; i < 3; i++) {
    await got('http://localhost:31337', {agent})
  }
})()
```

For the first request, the socket connection listener is hit once; for the second, twice; for the third, thrice; etc. So clearly, something isn't getting cleaned up.

Of course, creating lots of internal state that isn't garbage-collected creates quite a big issue. So unless we're missing something, perhaps it should be made clear that recycling sockets isn't supported for the time being. In the longer term, `got` should probably clean up after itself even if the socket isn't disposed.

This wasn't the case with version 6, so presumably, since version 7, one of the closures in `requestAsEventEmitter()` is leaky. In general, personally, I'm a fan of using `.once()` where possible, and calling `.removeEventListener()` explicitly whenever I can. I feel like that would also be the more prudent approach here.

This was originally reported by @pietermees as part of his investigation for #353. The `EventEmitter` leak warning there actually wasn't a red herring at all.

Thanks!

PR #406, opened 2017-11-06
Fix socket `connect` listener memory leak

pietermees:
This PR improves test coverage for the `agent` option and fixes an issue where the `connect` listener on the socket was not cleaned up, causing a memory leak when sockets are being reused across multiple requests (as is the case when you're using agents)

This fix relates to https://github.com/sindresorhus/got/issues/399

### DIFF
--- a/index.js
+++ b/index.js
@@ -221,7 +221,7 @@ function requestAsEventEmitter(opts) {
 					total: uploadBodySize
 				});
 
-				req.connection.on('connect', () => {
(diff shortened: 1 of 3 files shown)


pietermees:
@sindresorhus I think this is ready for review, as usual the tests seem to fail randomly on the cancelation though

sindresorhus:
Thanks for fixing this and the comprehensive test addition :)

PR #413, opened 2017-11-14
Exclude passed headers with values `null` or `undefined`

### LINKED ISSUE #410: Don't include a given request header when its value is null or undefined
If you do `got(url, { headers: { cookie: null } })`, the HTTP request has a header of `cookie: null` rather than not appearing at all. If you replace `null` with `undefined` then you get an error.

If a header is `null` or `undefined` the it shouldn't be added to the request headers.

brandon93s:
**Notes:**
- Extract call to `lowercaseKeys` to leverage the object-safety it provides
- Delete headers with a value of `null` or `undefined`
- Considered exclusion of `''` as well, but this seems more common / legitimate from my research

Closes #410

### DIFF
--- a/index.js
+++ b/index.js
@@ -458,10 +458,17 @@ function normalizeArguments(url, opts) {
 		opts
 	);
 
+	const headers = lowercaseKeys(opts.headers);
+	for (const key of Object.keys(headers)) {
+		if (headers[key] === null || headers[key] === undefined) {
+			delete headers[key];
+		}
+	}
+
 	opts.headers = Object.assign({
 		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`,
 		'accept-encoding': 'gzip,deflate'
-	}, lowercaseKeys(opts.headers));
+	}, headers);
 
 	const query = opts.query;
 
--- a/test/headers.js
+++ b/test/headers.js
@@ -89,6 +89,24 @@ test('form-data automatic content-type', async t => {
 	t.is(headers['content-type'], `multipart/form-data; boundary=${form.getBoundary()}`);
 });
 
+test('remove null value headers', async t => {
+	const headers = (await got(s.url, {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -458,10 +458,13 @@ function normalizeArguments(url, opts) {
 		opts
 	);
 
+	const headers = lowercaseKeys(opts.headers);
+	Object.keys(headers).forEach(key => (headers[key] === null || headers[key] === undefined) && delete headers[key]);
```
lukechilds: There's a lot going on in this line and it's inconsistent with the rest of the codebase.

```

Comment (lukechilds) [inline index.js]: There's a lot going on in this line and it's inconsistent with the rest of the codebase.

e.g `Array#forEach` instead of `for...of`, logical operator instead of `if`.

```
for (const header of Object.keys(headers)) {
  if (headers[header] === null || headers[header] === undefined) {
    delete headers[header];
  }
};
```

Comment (lukechilds) [inline index.js]: There's a lot going on in this line and it's inconsistent with the rest of the codebase.

e.g `Array#forEach` instead of `for...of`, logical operator instead of `if`.

I would suggest the following for better readability and consistency.

```js
for (const header of Object.keys(headers)) {
  if (headers[header] === null || headers[header] === undefined) {
    delete headers[header];
  }
};
```

lukechilds: There's a lot going on in this line and it's inconsistent with the rest of the codebase.

e.g `forEach` instead of `for...of`, `&&` instead of `if`.

I would suggest the following for better readability and consistency.

```js
for (const key of Object.keys(headers)) {
  if (headers[key] === null || headers[key] === undefined) {
    delete headers[key];
  }
}
```

Functionally all looks good though 👍

brandon93s:
👍 replaced `forEach` with `for..of`

sindresorhus:
Thanks Brandon :)

Issue #392, opened 2017-10-16
Failures from cache PR


sindresorhus:
Some issues after merging the cache PR:

- [x] https://github.com/sindresorhus/got/commit/2bc2b90f9e1517a5c79129f782607d68f57cc92f
- [ ] https://travis-ci.org/sindresorhus/got/jobs/288149045

// @lukechilds

lukechilds:
Oh no :/

Strange that the tests weren't failing on Travis in the PR builds 😕 

I finally got myself some paid work 🎉  but it's on a really tight deadline and I've lost a few days from travel (I'm back in Thailand). I need to catch up on some work but I'll get this fixed ASAP. Probably either tonight or tomorrow night.

lukechilds:
Regarding 2bc2b90, it seems to me this was originally a bug in Got. The fact that `http.request` could throw was never accounted for so in Promise mode the thrown http error caused the Promise to reject. Before the cache PR, in stream mode that error wouldn't of been caught, so definitely a bug IMO.

So this bug has now just moved to `cacheable-request`. It also existed in `request`: https://github.com/request/request/issues/2120

We should definitely fix this so thrown requests (as opposed to errored) can be handled properly, however I don't think implementing the previous behaviour and returning the raw http error is a good idea. This isn't documented anywhere in Got and appears to just be side affect from previous code. I would suggest either throwing this as a `got.RequestError` or creating a new error class to differentiate between requests that error and requests that immediately throw.

Thoughts?

sindresorhus:
> The fact that http.request could throw was never accounted for so in Promise mode the thrown http error caused the Promise to reject.

How is the Promise rejecting incorrect? It is currently accounted for in Promise mode, as https://github.com/sindresorhus/got/commit/92ed73ad7f2a9ab3bb582a208f9f8eb7623092f5 proves. No idea about Stream mode though.

sindresorhus:
> Potential solution in lukechilds/cacheable-request#14 that would align neatly with Got's functionality.

👍

lukechilds:
> How is the Promise rejecting incorrect? It is currently accounted for in Promise mode, as 92ed73a proves. No idea about Stream mode though.

Ahh, my bad. If it's intentional then IMO it would be nicer if it could be returned as one of the proper Got error types, like `got.RequestError`.

I'll make that proposed change to `cacheable-request` and then we can throw a `got.RequestError` if `http.request()` throws.

lukechilds:
[...]

PR #414, opened 2017-11-15
Ensure cancel tests wait for a connection before cancelling

lukechilds:
Resolves https://github.com/sindresorhus/got/issues/392

This is just a WIP to explain my findings, don't merge.

**Update:** Ready to merge

Intermittent test failures were caused by cancel tests relying on `setTimeout` to wait for the connection to be established. Resolved by implementing a `connection` event which is still reliable in slow CI environments.


lukechilds:
@sindresorhus Ok, I think I might know what's causing the cancel test to intermittently fail.

I think it's because on Travis (with AVA concurrency at 4 but only one thread available) any extra async code introduces significant overhead. Because the thread is tied up with four processes, simply calling `Promise.resolve(value).then(fn)` means `fn` has to wait until the thread is free again which could be a while. In most production environments this would take milliseconds but in our specific CI environment this could potentially be in the hundreds of milliseconds due to the thread being overworked.

I think possibly the extra overhead of the Promises introduced in the cache option code is enough to (sometimes) break the cancel tests because they [rely on](https://github.com/sindresorhus/got/blob/883e40f55fed24490c705ff1086ae38c1ce74d2d/test/cancel.js#L40-L44) a `setTimeout(fn, 100)` for the stream to be established before cancelling it.

This is supported by the fact that increasing to `setTimeout(fn, 500)` means that the tests consistently pass. I've re-ran the [Travis build](https://travis-ci.org/lukechilds/got/builds/302318728) about 30 times and tests passed every time.

With `setTimeout(fn, 100)` roughly every 1 in 5 Travis builds failed.

I'm not 100% sure this is the reason why but it seems to make sense. Hopefully the commit messages help you understand how I reached this conclusion.

What are your thoughts?

lukechilds:
Also, for whatever reason, the Travis builds don't seem to be showing up for every commit in this PR. You can view the identical PR on my repo for all the commit builds: https://github.com/lukechilds/got/pull/1

lukechilds:
> Sorry for writing a flaky test 🙈.

Haha, it's a pretty odd edge case, I'm still not 100% sure that's the reason.

> You could consider returning a connection event emitter from the abort server too if it turns flaky again.

I like this solution, if that works we can be pretty sure the problem won't come back again. I'll give it a bash now.

lukechilds:
Ok, so ensuring we reliably wait for the connection via triggering an event as opposed to setTimeout definitely seems to work.

 Just ran the [Travis Build](https://travis-ci.org/lukechilds/got/builds/302372583) another 30 times without a any test failures.



I think this can be merged.

sindresorhus:
Great!

AlexTes:
👏

PR #415, opened 2017-11-16
Fix EventEmitter memory leak

### LINKED ISSUE #399: Socket recycling causes memory leaks

sindresorhus:
Fixes #399 // @timdp

Currently, when you run the Got tests it outputs warnings about EventEmitter memory leaks.

Tip: Run the tests with `NODE_OPTIONS=--trace-warnings ava` to see the stack trace of the warnings

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,7 +198,7 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
-		cacheReq.once('error', err => {
+		cacheReq.on('error', err => {
 			if (err instanceof CacheableRequest.RequestError) {
 				ee.emit('error', new got.RequestError(err, opts));
 			} else {
@@ -427,7 +427,7 @@ function asStream(opts) {
 		proxy.emit('response', res);
 	});
 
-	ee.once('error', proxy.emit.bind(proxy, 'error'));
+	ee.on('error', proxy.emit.bind(proxy, 'error'));
 	ee.on('redirect', proxy.emit.bind(proxy, 'redirect'));
 	ee.on('uploadProgress', proxy.emit.bind(proxy, 'uploadProgress'));
 	ee.on('downloadProgress', proxy.emit.bind(proxy, 'downloadProgress'));
--- a/index.js
+++ b/index.js
@@ -198,7 +198,7 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
-		cacheReq.on('error', err => {
+		cacheReq.once('error', err => {
 			if (err instanceof CacheableRequest.RequestError) {
 				ee.emit('error', new got.RequestError(err, opts));
 			} else {
@@ -206,7 +206,7 @@ function requestAsEventEmitter(opts) {
 			}
 		});
 
-		cacheReq.on('request', req => {
+		cacheReq.once('request', req => {
 			req.once('error', err => {
 				clearInterval(progressInterval);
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -220,7 +220,7 @@ function requestAsEventEmitter(opts) {
 				ee.emit('error', new got.RequestError(err, opts));
 			});
 
-			ee.on('request', req => {
+			ee.once('request', req => {
```
sindresorhus: This is the offending line. The other ones are just safeguarding for future memory leaks.

index.js
```diff
@@ -198,15 +198,15 @@ function requestAsEventEmitter(opts) {
 			});
 		});
 
-		cacheReq.on('error', err => {
+		cacheReq.once('error', err => {
```
lukechilds: `cacheable-request` could potentially emit more than one error. These should be passed through.

index.js
```diff
@@ -347,10 +347,10 @@ function asPromise(opts) {
 				});
 		});
 
-		ee.on('error', reject);
+		ee.once('error', reject);
```
lukechilds: This is fine, we can only reject once, so we should do it on the first error.

index.js
```diff
@@ -427,8 +427,8 @@ function asStream(opts) {
 		proxy.emit('response', res);
 	});
 
+	ee.once('error', proxy.emit.bind(proxy, 'error'));
```
lukechilds: `cacheable-request` could potentially emit more than one error. These should be passed through.

index.js
lukechilds: Yep, all good.

timdp:
This is awesome. Thanks, guys!

Issue #417, opened 2017-11-16
Retry on some non-2xx HTTP status codes


rhodgkins:
Hi,

Just wanted to check why got doesn't do this...!

I thought the `retries` option would also be called HTTP errors but this doesn't appear to be the case ([Promise](https://github.com/sindresorhus/got/blob/v7.1.0/index.js#L182) / [as stream](https://github.com/sindresorhus/got/blob/v7.1.0/index.js#L250)) - is this by design?

I'd like got to retry certain 5xx error codes.

Cheers,

Rich

petermetz:
https://github.com/sindresorhus/got/pull/421

sindresorhus:
I think we should do this.

I propose is that we retry on the following methods and status codes by default:

Methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
Status codes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`500`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
*(Sourced from experience and other request clients)*

We also need to respect the [`Retry-After` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After).


Related issue: https://github.com/sindresorhus/got/issues/379

@lukechilds @reconbot @floatdrop @brandon93s Thoughts?

reconbot:
Retry after support is awesome

lukechilds:
One potential side effect to be aware of, especially if we're thinking of enabling this by default:

Consider you reach the request limit quota for the API you're using and they return `429 Too Many Requests` with a `Retry-After` header set for when your quota resets next month.

This would mean, if we by default retry `429`s and respect `Retry-After`, that making a request with Got to your API once you've hit your quota would return a Promise that may not resolve for up to 30 days in the future.

That doesn't seem very intuitive and could cause bugs that would be a nightmare to track down.

This seems like a great feature but might be better as opt in or not whitelisting `429`.

[...]

PR #421, opened 2017-11-19
Add failing test for retrying on status code 503

petermetz:
…ng retried (test is currently failing which is the point)

### DIFF
--- a/test/retry.js
+++ b/test/retry.js
@@ -6,6 +6,7 @@ let s;
 let trys = 0;
 let knocks = 0;
 let fifth = 0;
+let fifthPOST = 0;
 
 test.before('setup', async () => {
 	s = await createServer();
@@ -28,11 +29,21 @@ test.before('setup', async () => {
 		}
 	});
 
-	await s.listen(s.port);
-});
+	s.on('request', (req, res) => {
+		if (req.method === 'POST' && req.url.includes('fifthPOST')) {
+			console.log(`Sending back HTTP 503 Service Unavailable ...`);
+			fifthPOST++;
+			if (fifthPOST === 5) {
+				res.end(`All is well now, calm down.`);
+			} else {
+				res.statusCode = 503;
+				res.statusMessage = 'Service Unavailable';
+				res.end('You are cool, just make sure you retry this request....');
+			}
+		}
+	});
 
-test('works on timeout error', async t => {
-	t.is((await got(`${s.url}/knock-twice`, {timeout: {connect: 100, socket: 100}})).body, 'who`s there?');
+	await s.listen(s.port);
 });
 
 test('can be disabled with option', async t => {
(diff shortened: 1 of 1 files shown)


test/retry.js
```diff
@@ -28,11 +29,21 @@ test.before('setup', async () => {
...
-});
+	s.on('request', (req, res) => {
+		if (req.method === 'POST' && req.url.includes('fifthPOST')) {
+			console.log(`Sending back HTTP 503 Service Unavailable ...`);
+			fifthPOST++;
+			if (fifthPOST === 5) {
+				res.end(`All is well now, calm down.`);
+			} else {
+				res.statusCode = 503;
+				res.statusMessage = 'Service Unavailable';
```
sindresorhus: This is not needed. It will be set for us by Node.

test/retry.js
```diff
@@ -28,11 +29,21 @@ test.before('setup', async () => {
 		}
 	});
 
-	await s.listen(s.port);
-});
+	s.on('request', (req, res) => {
+		if (req.method === 'POST' && req.url.includes('fifthPOST')) {
+			console.log(`Sending back HTTP 503 Service Unavailable ...`);
```
sindresorhus: Remove this

test/retry.js
```diff
@@ -28,11 +29,21 @@ test.before('setup', async () => {
...
+			fifthPOST++;
+			if (fifthPOST === 5) {
+				res.end(`All is well now, calm down.`);
+			} else {
+				res.statusCode = 503;
+				res.statusMessage = 'Service Unavailable';
+				res.end('You are cool, just make sure you retry this request....');
```
sindresorhus: Use something shorter, like `Error`.

test/retry.js
```diff
@@ -28,11 +29,21 @@ test.before('setup', async () => {
 		}
 	});
 
-	await s.listen(s.port);
-});
+	s.on('request', (req, res) => {
```
sindresorhus: Use `s.on('/retry-503', (req, res) => {`

petermetz:
> Can you change this to use GET? Retrying on POST by default is not safe, so even when we support retrying on HTTP status codes, we won't do POST by default.

This should've been more clear in my PR explanation: I'm not proposing to change the default behavior. I do propose that the documented feature of the `iter` callback to be operational for `POST` verbs as it is for get. This in my opinion doesn't make `POST` retries the default, but it makes it possible for people to opt-in.
My use-case for this was about writing integration tests and verifying some error handling logic on the server side (can't remember the exact details to be honest).

So with all that said my answer is that if you don't approve of the opt-in `POST` retries idea, then the whole PR is pointless anyway and I will rest my case by simply closing the PR.

sindresorhus:
I don't like this way of opting in. It should be even more explicit. I have some ideas on how the retry functionality can be significantly improved: https://github.com/sindresorhus/got/issues/417#issuecomment-366602879

Issue #418, opened 2017-11-16
Got v9


sindresorhus:
Probably for version 9 at some point in the future. **(This is not happening soon)**

- Target Node.js 8
- Use async/await, destructuring, spread
- Use `require('url').URL` (WHATWG URL) over `require('url')` and `require('url').URLSearchParams` over `require('querystring')`
- Remove `safe-buffer` dependency
- Reduce dependencies

Ideas welcome for breaking changes. New features and bugs should be opened as individual issues.

brandon93s:
- Use `util.promisify` over `pify` 
- Replace `is-plain-obj` dependency with `is` and leverage throughout for additional readability and code clarity: ([1](https://github.com/sindresorhus/got/blob/master/index.js#L33), [2](https://github.com/sindresorhus/got/blob/master/index.js#L46), [3](https://github.com/sindresorhus/got/blob/master/index.js#L58), [4](https://github.com/sindresorhus/got/blob/master/index.js#L71), [5](https://github.com/sindresorhus/got/blob/master/index.js#L182), [6](https://github.com/sindresorhus/got/blob/master/index.js#L439-L441), [7](https://github.com/sindresorhus/got/blob/master/index.js#L481), [8](https://github.com/sindresorhus/got/blob/master/index.js#L496), [9](https://github.com/sindresorhus/got/blob/master/index.js#L517), [10](https://github.com/sindresorhus/got/blob/master/index.js#L543), [11](https://github.com/sindresorhus/got/blob/master/index.js#L562), [12](https://github.com/sindresorhus/got/blob/master/index.js#L543), etc) 
- Use `require('url').URL` (WHATWG URL) as a replacement for `isurl` and `url-to-options` dependencies
- Extract errors into separate file and use an assignment syntax similar to http methods (e.g. `for (error of errors)`)

sindresorhus:
@brandon93s Agreed :)

sindresorhus:
> Replace is-plain-obj dependency with is and leverage throughout for additional readability and code clarity: (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, etc)

> Extract errors into separate file and use an assignment syntax similar to http methods (e.g. for (error of errors))

We can do these today. No need to wait. PR welcome.

brandon93s:
I'll tackle the `is` implementation once https://github.com/sindresorhus/is/pull/34 lands!

brandon93s:
[...]

PR #422, opened 2017-11-19
Use `@sindresorhus/is` for type checking

brandon93s:
- Replace all type checks with `is.{method}`
- Remove `is-plain-obj` dependency
- Remove `is-stream` dependency 

---

#418


kevva:
> This also slightly improves the code readability.

Amazingly much.

PR #429, opened 2017-11-27
Check req.connection exists before attaching event listeners

### LINKED ISSUE #427: Uncaught TypeError: Cannot read property 'once' of undefined     at EventEmitter.ee.once.req
I'm attempting to use `got` from the browser with `browserify`. Looking through the issues I see others have had success, but I'm currently getting an error when using it. My request completes with the correct respsonse but an error come with it.

```
Uncaught TypeError: Cannot read property 'once' of undefined
    at EventEmitter.ee.once.req (bundle.js:10705)
    at EventEmitter.g (bundle.js:7590)
    at EventEmitter.emit (bundle.js:7521)
    at setImmediate (bundle.js:10740)
    at onNextTick (bundle.js:9657)
    at Item.run (bundle.js:8030)
    at drainQueue (bundle.js:8000)
```

The code is a fairly simple example.
```index.js
const got = require('got')
global.setImmediate = require('timers').setImmediate

const options = {
        method: 'GET'
}

got('https://api.github.com', options)
        .then(x => console.log(x.body))
        .catch(console.log)
```
I'm running this command to make my `bundle.js` file
```
browserify index.js > bundle.js
```

Have I done something incorrectly?

lukechilds:
Fixes #427

Obviously progress events won't work in the browser, do we need to specify this in the readme?

### DIFF
--- a/index.js
+++ b/index.js
@@ -226,33 +226,35 @@ function requestAsEventEmitter(opts) {
 					total: uploadBodySize
 				});
 
-				req.connection.once('connect', () => {
-					const uploadEventFrequency = 150;
+				if (req.connection) {
+					req.connection.once('connect', () => {
+						const uploadEventFrequency = 150;
 
-					progressInterval = setInterval(() => {
-						const lastUploaded = uploaded;
-						const headersSize = Buffer.byteLength(req._header);
-						uploaded = req.connection.bytesWritten - headersSize;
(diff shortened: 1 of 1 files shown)


lukechilds:
`req.connection` also sometimes appears to be missing in Node.js when Got is used with a proxy agent.

PR #431, opened 2017-11-30
Update p-timeout to version 2.0.1

lpinca:
Currently the request is not aborted if the connection cannot be established and the timeout expires. This prevents the process from exiting.

Test case:

```js
const got = require('.');

// 192.0.2.1 is non-routable.
got('http://192.0.2.1', { timeout: 1000, retries: 0 }).then(null, (err) => {
  console.error(err);
});
```


lukechilds:
Thanks!

PR #436, opened 2017-12-09
Extract errors into errors.js

brandon93s:
- Extract errors into their own file
- Export each error individually 
- Apply each error to `got`, maintaining functionality  

----

Related to #418

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,6 +25,7 @@ const pTimeout = require('p-timeout');
 const pify = require('pify');
 const Buffer = require('safe-buffer').Buffer;
 const pkg = require('./package.json');
+const errors = require('./errors');
 
 const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -597,89 +598,10 @@ for (const method of methods) {
...
-			method: opts.method,
-			path: opts.path,
-			protocol: opts.protocol,
-			url: opts.href
-		});
+for (const error in errors) {
+	if (Object.prototype.hasOwnProperty.call(errors, error)) {
+		got[error] = errors[error];
 	}
 }
```
```js
for (const error of Object.keys(errors)) {
	got[error] = errors[error];
}
```

sindresorhus:
You need to add `errors.js` to the `files` property in package.json

index.js
sindresorhus: Actually, can just use Object.assign:

```js
Object.assign(got, errors);
```

panva:
How would you feel about exposing `StdError` as well? When working with `got` promise rejections a developer might find checking for the following useful.

```js
((async () => {
  try {
    await got.get(url);
  } catch (err) {
    if (err instanceof StdError) {
      //
    } else {
      //
    }
  }
})();

maybe considering a different name, such as GotError too?

sindresorhus:
Yeah, I think we should expose it too and rename it to `GotError`.

index.js
brandon93s: > Actually, can just use Object.assign:

Ah, duh!

brandon93s:
- Renamed to `GotError` and exposed
- Simplify with `Object.assign`
- Include `errors.js` in `files`

sindresorhus:
Looks good :)

PR #438, opened 2017-12-13
enable sending plain string over rest 

paulknulst:


sindresorhus:
Why?

paulknulst:
Hi,
sorry that does not really solve a problem. I just want to fork your project. pressed the wrong buttons i guess.

what i want to do:
During communication between a REST-API and my project i need to send a POST containing only a string and no json object. additionally i have to set the headers accept/content-type to application/json.

i know that this is not the standard use case for post calls but the api im working with is kinda stupid. because i want to use your very good implemented package i change this line of code so i can use it with the REST-API im working with.

PR #439, opened 2017-12-13
Document cache behaviour

### LINKED ISSUE #423: Clarify cache behaviour
Thanks @lukechilds for adding caching to the library. Would you mind if I ask for a couple of clarifications?

I'd like to use the npmjs registry as my example. They return headers like so:
```
Cache-Control: max-age=300
ETag: W/"5a14f9ea-234b93"
Last-Modified: Wed, 22 Nov 2017 04:15:38 GMT
```

Consider a long-running app (hours/days) that queries npmjs regularly.

1. If queries are made for the exact same package (URL) twice within 5 minutes, will the second receive an answer directly from cache, or only after a conditional query with 304 response?
2. Will cached values ever "expire", or will they remain in cache indefinitely for use in `If-Modified-Since` and `If-None-Match` headers in future?

My current approach in got v7 is:
- check `keyv` cache for matching url
- if not found, `got` the request
- filter json result for data I need
- save filtered json using `keyv` with expiry 5 minutes

So I'm wondering how replacing this with `got` v8's implicit caching would impact things. (a) it will obviously store the entire (huge) npm responses as opposed to my filtered ones, but (b) it would result in a lot more 304 responses assuming `got` v8 keeps the "old" values longer than 300s.

lukechilds:
Resolves #423

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -339,6 +339,8 @@ The promise returned by Got has a [`.cancel()`](https://github.com/sindresorhus/
 <a name="cache-adapters"></a>
 ## Cache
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Perfect :)

PR #440, opened 2017-12-14
Add new related project to the readme

khaosdoctor:
I've created a library to wrap around Got and GraphQL but using JSON as base, not strings. Just including it here 😄


sindresorhus:
Cool idea :)

PR #441, opened 2017-12-18
Validate that `url` is UTF-8 encoded

### LINKED ISSUE #420: URI malformed when trying to get URL with URL-encoded chars
Hello,

Catched URI malformed error in new "got", when I trying to send request to URL with URL-encoded chars.
URL examples:
https://www.kinopoisk.ru/community/city/%D2%E0%EB%EB%E8%ED/
https://www.kinopoisk.ru/news/keyword/%C7%E2%E5%E7%E4%ED%FB%E5+%E2%EE%E9%ED%FB/

nodejs: 9.2.0
got: 8.0.0

Failed code:
```js
const got = require('got');

(async () => {
	try {
		const response = await got('https://www.kinopoisk.ru/community/city/%D2%E0%EB%EB%E8%ED/');
		console.log(response);
	} catch (error) {
		console.log(error);
	}
})();
```
```
URIError: URI malformed
    at decodeURI (<anonymous>)
    at module.exports (/Users/kirill-m/git/test/node_modules/normalize-url/index.js:87:21)
    at /Users/kirill-m/git/test/node_modules/cacheable-request/src/index.js:43:16
    at get (/Users/kirill-m/git/test/node_modules/got/index.js:98:20)
    at Promise.resolve.then.size (/Users/kirill-m/git/test/node_modules/got/index.js:274:5)
    at <anonymous>
```

Broken at this commit: https://github.com/sindresorhus/got/commit/3c7920507fae88a5f53d0640b5116fa34a5ed829

Any ideas?

brandon93s:
Opted to decode without assignment back to `url` so that requests and other error messages continue to display the user provided `url` value. 

---
Closes #420

### DIFF
--- a/index.js
+++ b/index.js
@@ -444,6 +444,13 @@ function normalizeArguments(url, opts) {
 		throw new TypeError(`Parameter \`url\` must be a string or object, not ${is(url)}`);
 	} else if (is.string(url)) {
 		url = url.replace(/^unix:/, 'http://$&');
+
+		try {
+			decodeURI(url);
+		} catch (err) {
+			throw new Error('Parameter `url` must contain valid UTF-8 character sequences');
+		}
+
 		url = urlParseLax(url);
 		if (url.auth) {
 			throw new Error('Basic authentication must be done with the `auth` option');
--- a/test/arguments.js
+++ b/test/arguments.js
@@ -29,6 +29,11 @@ test('url is required', async t => {
 	t.regex(err.message, /Parameter `url` must be a string or object, not undefined/);
 });
 
+test('url should be utf-8 encoded', async t => {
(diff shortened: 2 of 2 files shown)


lukechilds:
Will `decodeURI()` only ever throw if the URI contains non UTF-8 chars?

Just to clarify if there are any other circumstances that could cause it to throw then it might be worth either purely checking for non UTF-8 chars or changing the error message to "valid URI".

However it looks like we're ok:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI

> Throws an URIError ("malformed URI sequence") exception when encodedURI contains invalid character sequences.

sindresorhus:
Looks good. Thanks @brandon93s.

PR #451, opened 2018-02-01
Add `throwHttpErrors` option

### LINKED ISSUE #430: Add a `reject` option that allow to returns error rather than throwing them
It's sometime useful to make an http and only verify the `statusCode`. For example to check is a resource is available (check `statusCode === 404`) or authorized (check `statusCode === 401`).

Current that requires to handle the error with a `try / catch`:
```js
async function isAuthorized(url) {
  try {
    const response = await got(url);
    return response.statusCode !== 401;
  } catch (err) {
    if (err.statusCode) {
      return err.statusCode !== 401;
    }
    throw err;
  }
}
```
The `reject` option would work similarly to [execa#reject](https://github.com/sindresorhus/execa#reject). In case of a 4xx error the error would be returned instead of being thrown. Other error cases (5xx, network, timeout etc...) would throw the error. 

The code above would look like this:
```js
async function isAuthorized(url) {
  return (await got(url, {reject: false})).statusCode !== 401;
}
```

brandon93s:
Implemented the `throwHttpErrors` option discussed in #430.

- [X] Documented
- [X] Default to true
- [X] Implemented for HTTP
- [X] Implemented for Streams (for consistentency)
- [X] Tested
---
Closes #430

### DIFF
--- a/index.js
+++ b/index.js
@@ -337,7 +337,7 @@ function asPromise(opts) {
 						}
 					}
 
-					if (statusCode !== 304 && (statusCode < 200 || statusCode > limitStatusCode)) {
+					if (opts.throwHttpErrors && statusCode !== 304 && (statusCode < 200 || statusCode > limitStatusCode)) {
 						throw new got.HTTPError(statusCode, res.statusMessage, res.headers, opts);
 					}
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
Thanks @brandon93s :)

PR #452, opened 2018-02-04
Allow calling of `got.stream` via options

### LINKED ISSUE #362: Calling got.stream via options flag with got()
Looking through the source, I see `got.stream` calls `asStream`, with got calling `asPromise`, with the various methods added one level down.

I'd love to be able to do something like: 

```
got(url, {
  stream: true
})
``` 

which under the hood would call `asStream`. Something similar is fine too.

The use case is to have got wrapped so each request has the same configurable retries and timeout specs.

sotojuan:
Fixes #362.

This commit allows the user to pass a `stream` flag to `opts`. When this flag is `true`, `got` returns a stream just as if it was called with `got.stream`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -585,7 +585,13 @@ function normalizeArguments(url, opts) {
 
 function got(url, opts) {
 	try {
-		return asPromise(normalizeArguments(url, opts));
+		const normalizedArgs = normalizeArguments(url, opts);
+
+		if (normalizedArgs.stream) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you also document it?

sotojuan:
👍

sindresorhus:
Great. Thank you Juan :)

PR #387, opened 2017-10-08
[WIP] Use create-test-server in tests

lukechilds:
Totally forgot about this.

Requested on Gitter:

> @lukechilds We should use https://github.com/lukechilds/create-test-server in the Got tests.

https://gitter.im/got-dev/Lobby?at=59319ef6cf9c13503c4931eb

WIP, not ready to be merged yet.

PR #455, opened 2018-02-12
Add documentation on how to test requests

### LINKED ISSUE #454: How to mock "got"?
I'm wondering if there is an easy way to mock "got" dependency?
I managed to deal with it using `proxyquire` but looking for easier solution.

kevva:
Fixes #454.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -588,6 +588,24 @@ nock('https://sindresorhus.com')
 })();
 ```
 
+If you need real integration tests you can use [`create-test-server`](https://github.com/lukechilds/create-test-server):
+
+```js
+const got = require('got');
+const createTestServer = require('create-test-server');
+
+(async () => {
+	const server = await createTestServer();
+	server.get('/', 'Hello world!');
+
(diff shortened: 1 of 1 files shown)


lukechilds:
Do you think it would be worth mentioning [`create-test-server`](https://github.com/lukechilds/create-test-server) here?

The official Got tests will actually be using this once #387 is merged?

kevva:
> Do you think it would be worth mentioning create-test-server here?

Might be. I think their use cases are a little different. E.g. I would use `nock` for mocking real API requests while I would use `create-test-server` in other cases where I'd just need a server.

sindresorhus:
I think it's worth mentioning it. Nock should be good enough for most REST cases, but sometimes you need real integration tests, and that's when `create-test-server` becomes useful.

sindresorhus:
@lukechilds LGTY?

lukechilds:
Yeah, perfect 👌

sindresorhus:
Thanks Kevin 👌✨

PR #456, opened 2018-02-13
Don't set `accept-encoding` if decompress is false

### LINKED ISSUE #419: Problem with default `Accept-Encoding` header
Why Got adds a default `Accept-Encoding` header ([this line](https://github.com/sindresorhus/got/blob/master/index.js#L475)) even with disabled `decompress` option?

**My scenario:** In my current project I wanted to forward some client request headers from one server to another and header values must be equal in both servers, but when `Accept-Encoding` header is not set, Got adds this header with a default value and that is a problem in this scenario.

I can't find a way to send a request with no `accept-encoding` header,  I had to set an empty string in `accept-encoding` to get around this:

```js
app.use(ctx => {
  const headers = ctx.headers; // client request headers

  got(url, {
    headers: {
      ...headers,
      // Prevent got from adding default accept-encoding header
      'accept-encoding': headers['accept-encoding'] || ''
    }
  })
})
```
`null` and `undefined` not working because in v8.0.0 it will be deleted and then default header will be added.

**My suggestion:** Only set default `accept-encoding` when `decompress` option is enabled, then the following code will work for my scenario:

```js
got(url, {
  decompress: false,
  headers
})
```

I can send a PR, but first I wanted to ask that is there any specific reason to set default `accept-encoding` other than `decompress` option?

sotojuan:
Fixes #419.

Let me know if you want to move this somewhere else or use a different style 👍 . I originally wanted to put this inline under `user-agent` but then if disabled the headers object would still have an `accept-encoding` instead of none at all.

### DIFF
--- a/test/headers.js
+++ b/test/headers.js
@@ -29,7 +29,7 @@ test('accept-encoding', async t => {
 
 test('do not set accept-encoding header when decompress options is false', async t => {
 	const headers = (await got(s.url, {json: true, decompress: false})).body;
-	t.false(Reflect.has(headers, 'accept-encoding'));
+	t.false(Object.prototype.hasOwnProperty.call(headers, 'accept-encoding'));
 });
 
 test('accept header with json option', async t => {
--- a/test/headers.js
+++ b/test/headers.js
@@ -29,7 +29,7 @@ test('accept-encoding', async t => {
 
 test('do not set accept-encoding header when decompress options is false', async t => {
 	const headers = (await got(s.url, {json: true, decompress: false})).body;
-	t.false(Object.prototype.hasOwnProperty.call(headers, 'accept-encoding'));
+	t.false(Reflect.has(headers, 'accept-encoding'));
 });
 
 test('accept header with json option', async t => {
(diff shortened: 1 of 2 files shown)


test/headers.js
```diff
@@ -27,6 +27,11 @@ test('accept-encoding', async t => {
 	t.is(headers['accept-encoding'], 'gzip,deflate');
 });
 
+test('do not set accept-encoding if decompress not enabled', async t => {
+	const headers = (await got(s.url, {json: true, decompress: false})).body;
+	t.false(Object.prototype.hasOwnProperty.call(headers, 'accept-encoding'));
```
sindresorhus: Can you use `Reflect.has()` instead?

test/headers.js
```diff
@@ -27,6 +27,11 @@ test('accept-encoding', async t => {
 	t.is(headers['accept-encoding'], 'gzip,deflate');
 });
 
+test('do not set accept-encoding if decompress not enabled', async t => {
```
sindresorhus: > do not set `accept-encoding` header when `decompress` options is false

sotojuan:
@sindresorhus Seems like Node 4 doesn't support `Reflect`.

sindresorhus:
Ugh, I dread having to support older Node.js versions. Revert that change then...

sotojuan:
👍

lukechilds:
Hang in there Node.js 4 is EOL in April 🙏

sindresorhus:
Thanks Juan :)

Issue #4, opened 2014-08-16
Support for binary files


julien-f:
IMHO, it is too bad that this module cannot be used for binary content due to the `setEncoding('utf-8')`.

I think it should instead return a buffer and let the user call `toString()` to obtain a string if necessary.

See https://gist.github.com/julien-f/9cc9976ce768890eb274#file-get-js for inspiration, maybe :)

shaunc:
Is there any way to get `response.body` as a buffer if you want it that way? Seems even with `{encoding: 'binary'}` response body is string. Or do you have to use stream?

sindresorhus:
@shaunc From the docs:

> Encoding to be used on setEncoding of the response data. **If null, the body is returned as a Buffer.**

`binary` is a string encoding, not actual binary nor buffer.

PR #457, opened 2018-02-20
Improve docs

dcharbonnier:
It does not seams so easy to find this option even if it's obvious when you think about it.
This will allow for a search on the keywork "binary" on github and avoid to jump into this issue https://github.com/sindresorhus/got/issues/4

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -129,7 +129,7 @@ If `content-length` or `transfer-encoding` is not set in `options.headers` and `
 Type: `string` `null`<br>
 Default: `'utf8'`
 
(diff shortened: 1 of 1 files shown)


kevva:
Doesn't this describe exactly what the previous sentence does?

dcharbonnier:
It is but the think is that people don't read a full doc, people search for what they need and what I need is to get "binary" data so I search "binary" and not "Buffer". I don't think it hurt to add it and it may save some time for other people but you can close if you disagree.

sindresorhus:
I think people should know what buffers are in Node, but I don't mind improving the docs either.

PR #459, opened 2018-02-23
Prevent uncaught exception in some cases

wwwouter:
If something goes wrong in the code contained by setImmediate, an uncaughtException is thrown.

The code in this pull requests wraps the code in a try/catch block and emits an error if an error is thrown.

### DIFF
--- a/index.js
+++ b/index.js
@@ -196,10 +196,10 @@ function requestAsEventEmitter(opts) {
 					});
 
 					res.pipe(progressStream);
-				});
-			} catch (e) {
-				ee.emit('error', e);
-			}
+				} catch (e) {
+					ee.emit('error', e);
+				}
+			});
 		});
 
 		cacheReq.on('error', err => {
--- a/index.js
+++ b/index.js
@@ -142,60 +142,9 @@ function requestAsEventEmitter(opts) {
 				return;
 			}
 
-			const downloadBodySize = Number(res.headers['content-length']) || null;
-			let downloaded = 0;
-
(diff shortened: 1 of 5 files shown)


kevva:
Could you add a test for this?

index.js
```diff
@@ -146,55 +146,59 @@ function requestAsEventEmitter(opts) {
 			let downloaded = 0;
 
 			setImmediate(() => {
-				const progressStream = new Transform({
-					transform(chunk, encoding, callback) {
-						downloaded += chunk.length;
+				try {
```
sindresorhus: Can you move the contents of the `try` into a function and call it from the `try`, so we don't add more nesting.

wwwouter:
I've put the code in an `getResponse` function and added a test.
This test will fail with the old code, but succeeds with the changes.

I've also added some files to work on VS Code on Windows. I can remove these if you want.

wwwouter:
Also: Travis is timing out, but this shows it's ok for at least node 4 and 8: https://travis-ci.org/sindresorhus/got/builds/346221646

kevva:
> I've also added some files to work on VS Code on Windows. I can remove these if you want.

Yes :).

wwwouter:
Ok, removed the .vscode folder!

sindresorhus:
Thanks :)

wwwouter:
You're welcome!
Could you take a quick peek at #461? 
Now at least I can catch the error, but it would be even better if there wasn't an error to catch :)

PR #465, opened 2018-03-01
Fix socket 'connect' memory leak

Nodge:
After turning on keep-alive agent with got@8.2.0 i found memory leak. The `connect` listener on a socket is not executed because socket already connected.

Related issue in `request` module: https://github.com/request/request/issues/2438

### DIFF
--- a/index.js
+++ b/index.js
@@ -227,14 +227,18 @@ function requestAsEventEmitter(opts) {
 					total: uploadBodySize
 				});
 
-				if (req.connection) {
-					req.connection.once('connect', () => {
+				const socket = req.connection;
+				if (socket) {
+					// `._connecting` was the old property which was made public in node v6.1.0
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -227,14 +227,18 @@ function requestAsEventEmitter(opts) {
 					total: uploadBodySize
 				});
 
-				if (req.connection) {
-					req.connection.once('connect', () => {
+				const socket = req.connection;
+				if (socket) {
+					// `._connecting` was the old property which was made public in node v6.1.0
+					const isConnecting = socket._connecting || socket.connecting;
```
floatdrop: You should check for `undefined` here explicitly.

Nodge: @floatdrop fixed

floatdrop:
Lgtm. @sindresorhus can we issue a patch with this fix?

PR #466, opened 2018-03-07
Add missing `Content-Length` header for `FormData` bodies

gasi:
S3 requires form uploads to specify `Content-Length` header, otherwise `411 Length Required` error is returned:
- https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html#mpUploadUploadPart-requests-request-headers
- https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

#  Sample Code

The following code can be used to reproduce the issue:

### Implementation

`debuglogs.js`
```javascript
const FormData = require('form-data');
const got = require('got');


const BASE_URL = 'https://debuglogs.org';

// Workaround: Submitting `FormData` using native `FormData::submit` procedure
// as integration with `got` results in S3 error saying we haven’t set the
// `Content-Length` header:
const submitFormData = (form, url) =>
  new Promise((resolve, reject) => {
    form.submit(url, (error) => {
      if (error) {
        return reject(error);
      }

      return resolve();
    });
  });

//      upload :: String -> Promise URL
exports.upload = async (content) => {
  const signedForm = await got.get(BASE_URL, { json: true });
  const { fields, url } = signedForm.body;

  const form = new FormData();
  // The API expects `key` to be the first field:
  form.append('key', fields.key);
  Object.entries(fields)
    .filter(([key]) => key !== 'key')
    .forEach(([key, value]) => {
      form.append(key, value);
    });

  const contentBuffer = Buffer.from(content, 'utf8');
  const contentType = 'text/plain';
  form.append('Content-Type', contentType);
  form.append('file', contentBuffer, {
    contentType,
    filename: 'got-missing-content-length-form-data.txt',
  });

  // Works:
  // await submitFormData(form, url);

  // Does not work without fix:
  await got.post(url, { body: form });

  return `${BASE_URL}/${fields.key}`;
};

```

### Test

`debuglogs_test.js`
```javascript
const { assert } = require('chai');
const got = require('got');

const debuglogs = require('./debuglogs');


describe('debuglogs', () => {
  describe('upload', () => {
    it('should upload log content', async () => {
      const nonce = Math.random().toString().slice(2);
      const url = await debuglogs.upload(nonce);

      const { body } = await got.get(url);
      assert.equal(nonce, body);
    });
  });
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -240,6 +240,9 @@ function requestAsEventEmitter(opts) {
 		Promise.resolve(getBodySize(opts))
 			.then(size => {
 				uploadBodySize = size;
+				if (isFormData(opts.body)) {
+					opts.headers['content-length'] = size;
+				}
 				get(opts);
 			})
 			.catch(err => {


sindresorhus:
We already check for this here: https://github.com/gasi/got/blob/c3daa9e2e9c7f9070dfa405a7747f75a3a058b04/index.js#L50-L52 So there might be an issue with that check.

gasi:
> We already check for this here: https://github.com/gasi/got/blob/c3daa9e2e9c7f9070dfa405a7747f75a3a058b04/index.js#L50-L52 So there might be an issue with that check.

I saw that but I am fairly sure it didn’t kick in my case. It’s been a while but the code above should help you repro the issue.

gasi:
Thanks, @sindresorhus 😄

PR #469, opened 2018-03-13
Clear the progress interval if the socket has been destroyed

AsaAyers:
At work recently we ran into a problem where we used `got` in a script and the `progressInterval` is preventing node from being able to shut down.

If you make a request to a local server that is not running, [`req.once('error', ...`](https://github.com/sindresorhus/got/blob/7d1aa01e69229ad50ce355589681ff1d1a765627/index.js#L168) fires before [`req.once('request', ...`](https://github.com/sindresorhus/got/blob/7d1aa01e69229ad50ce355589681ff1d1a765627/index.js#L185). This causes the `progressInterval` to be cleared before it is set. Then `setInterval` gets called and your process will never exit. This is reliable using both `127.0.0.1` and my computer's local IP `192.168.1.165`. This does not trigger if you use the hostname `localhost`.

Because the result of this is that node just never exits, I don't know how to add this to your existing tests. I have this example that I just put in the root of the project to verify the bug and fix.

```js
// example.js
const got = require('./index.js')

got('http://127.0.0.1:55555', { retries: 0 })
.then(function (response) {
  console.log('response', response)
})
.catch(function (error) {
  console.log('error', error)
})
```

With the following local modifications for debugging

<details>
<summary>diff</summary>

```diff
diff --git a/index.js b/index.js
index 3cd9d09..ea2bc6a 100644
--- a/index.js
+++ b/index.js
@@ -166,6 +166,7 @@ function requestAsEventEmitter(opts) {
                        });
 
                        req.once('error', err => {
+                               console.log(`clearInterval(${progressInterval})`)
                                clearInterval(progressInterval);
 
                                if (aborted) {
@@ -197,6 +198,7 @@ function requestAsEventEmitter(opts) {
                                        const onSocketConnect = () => {
                                                const uploadEventFrequency = 150;
 
+                                               console.log('onSocketConnect')
                                                progressInterval = setInterval(() => {
                                                        if (socket.destroyed) {
                                                                clearInterval(progressInterval)

```

</details>

I get this output and the process hangs forever.

```
asa:~/repos/got$ node example.js 
clearInterval(undefined)
error { Error: connect ECONNREFUSED 127.0.0.1:55555
    at ClientRequest.req.once.err (/home/asa/repos/got/index.js:183:22)
    at Object.onceWrapper (events.js:316:30)
    at emitOne (events.js:115:13)
    at ClientRequest.emit (events.js:210:7)
    at Socket.socketErrorListener (_http_client.js:401:9)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  name: 'RequestError',
  code: 'ECONNREFUSED',
  host: '127.0.0.1:55555',
  hostname: '127.0.0.1',
  method: 'GET',
  path: '/',
  protocol: 'http:',
  url: 'http://127.0.0.1:55555/' }
onSocketConnect
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,6 +200,7 @@ function requestAsEventEmitter(opts) {
 						progressInterval = setInterval(() => {
 							if (socket.destroyed) {
 								clearInterval(progressInterval);
+								return;
 							}
 							const lastUploaded = uploaded;
 							const headersSize = Buffer.byteLength(req._header);
--- a/index.js
+++ b/index.js
@@ -198,6 +198,9 @@ function requestAsEventEmitter(opts) {
 						const uploadEventFrequency = 150;
 
 						progressInterval = setInterval(() => {
+							if (socket.destroyed) {
+								clearInterval(progressInterval);
+							}
 							const lastUploaded = uploaded;
 							const headersSize = Buffer.byteLength(req._header);
 							uploaded = socket.bytesWritten - headersSize;
--- a/index.js
+++ b/index.js
@@ -202,6 +202,7 @@ function requestAsEventEmitter(opts) {
 								clearInterval(progressInterval);
 								return;
 							}
+
 							const lastUploaded = uploaded;
 							const headersSize = Buffer.byteLength(req._header);
 							uploaded = socket.bytesWritten - headersSize;
--- a/test/socket-destroyed.js
+++ b/test/socket-destroyed.js
@@ -1,0 +1,15 @@
+import test from 'ava';
+import got from '..';
+
+test.serial('Clear the progressInterval if the socket has been destroyed', async t => {
+	// There are 2 handles at this point
+	// const handleCount = process._getActiveHandles().length;
+
+	const err = await t.throws(got(`http://127.0.0.1:55555/`, {retry: 0}));
+	// Without the code from #469 there are 5 handles here. With #469's changes
+	// there are 4 handles. I can't figure out where the other two handles are
+	// getting creacted, so the best I can do is hard-code this 4 and hope someone
+	// else can figure out a better solution.
+	t.is(process._getActiveHandles().length, 4);
+	t.is(err.code, 'ECONNREFUSED');
+});
--- a/test/socket-destroyed.js
+++ b/test/socket-destroyed.js
@@ -1,15 +1,9 @@
 import test from 'ava';
 import got from '..';
 
-test.serial('Clear the progressInterval if the socket has been destroyed', async t => {
-	// There are 2 handles at this point
-	// const handleCount = process._getActiveHandles().length;
-
-	const err = await t.throws(got(`http://127.0.0.1:55555/`, {retry: 0}));
-	// Without the code from #469 there are 5 handles here. With #469's changes
-	// there are 4 handles. I can't figure out where the other two handles are
-	// getting creacted, so the best I can do is hard-code this 4 and hope someone
-	// else can figure out a better solution.
-	t.is(process._getActiveHandles().length, 4);
+test.serial('clear the progressInterval if the socket has been destroyed', async t => {
+	const handlesComingFromAVA = 2;
+	const err = await t.throws(got(`http://127.0.0.1:55555`, {retry: 0}));
+	t.is(process._getActiveHandles().length - handlesComingFromAVA, 2);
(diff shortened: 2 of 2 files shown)


brandon93s:
You may be able to test this with `process._getActiveHandles` with an assertion that the count is the same before and after the tested code. A failing test could be written demonstrating the lingering handle.

AsaAyers:
Thanks. It doesn't seem like a simple count works, I think ti's because ava runs tests in parallel. 

> Note that this only applies to tests within a particular test file. AVA will still run multiple tests files at the same time unless you pass the --serial CLI flag.

```
test.serial('Clear the progressInterval if the socket has been destroyed', async t => {
	const handleCount = process._getActiveHandles().length;
	const err = await t.throws(got(`http://127.0.0.1:55555/`, { retry: 0 }));
	t.is(err.code, 'ECONNREFUSED');
	t.is(process._getActiveHandles().length, handleCount)
})
```

I think it might work to see how many of the handles are timers. Using logging `Object.getPrototypeOf(handle)` I can see something that says its a `Timer`, but I don't know how else I can get a reference to that prototype to filter the list with. `Object.getPrototypeOf(setInterval(...` seems to give me a different thing named `Timer`.

szmarczak:
@AsaAyers Hmm, I've checked your issue on Windows 10 (latest got version), I got such error, but there's no `onSocketConnect`. The process exits normally. Take a look:
```
>node blah
clearInterval(undefined)
error { RequestError: connect ECONNREFUSED 127.0.0.1:55555
    at ClientRequest.req.once.err (C:\Users\Szymon Marczak\Desktop\project\node_modules\got\index.js:183:22)
    at Object.onceWrapper (events.js:254:19)
    at ClientRequest.emit (events.js:159:13)
    at Socket.socketErrorListener (_http_client.js:389:9)
    at Socket.emit (events.js:159:13)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  name: 'RequestError',
  code: 'ECONNREFUSED',
  host: '127.0.0.1:55555',
  hostname: '127.0.0.1',
  method: 'GET',
  path: '/',
  protocol: 'http:',
  url: 'http://127.0.0.1:55555/' }
```

AsaAyers:
The problem exists on Mac and Linux, I didn't know if it would trigger on windows.

index.js
```diff
@@ -198,6 +198,9 @@ function requestAsEventEmitter(opts) {
 						const uploadEventFrequency = 150;
 
 						progressInterval = setInterval(() => {
+							if (socket.destroyed) {
+								clearInterval(progressInterval);
```
sindresorhus: Shouldn't it return here? I don't think you want to run the below code if the socket is destroyed.

sindresorhus:
@AsaAyers Try moving the test to its own file. That way other tests will not affect it (Test files are run in separate processes)

mnmkng:
I just wanted to note, that this PR ( with the `return` suggested by @sindresorhus ) fixes the issue in my comment in: https://github.com/sindresorhus/got/issues/476

AsaAyers:
I added the early return and moved the test into its own file. There are 2 more handles getting created that I can't account for.

https://github.com/sindresorhus/got/blob/9e0d054f726e31f2ef50473d68b1fe74fc9dc2a3/test/socket-destroyed.js#L4-L15

sindresorhus:
@AsaAyers The other handles are probably ones used by AVA, the test runner. I think hard-coding is the best we can do.

szmarczak:
Sorry, but after half a year I've finally found the solution (I've just looked at the code for some improvements today). There's a bug:

https://github.com/sindresorhus/got/blob/21bef3c54106a28e9a2198eb01cb249d2d555e87/source/progress.js#L62-L68

If the socket closed before it connected (important thing: line 62 must return false), `onSocketConnect` is called: `socket.connecting` is false but it isn't writable (so it's destroyed). Using [`defer-to-connect`](https://github.com/szmarczak/defer-to-connect) (v1.0.1) would fix that.

PR #471, opened 2018-03-21
Use query-string instead of querystring

Hoishin:
Since Node's native `querystring` is incompatible with RFC3986, possible conflict can happen with the letters `!*'()`.

Added query-string for more strict encoding instead of `querystring`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ const PassThrough = require('stream').PassThrough;
 const Transform = require('stream').Transform;
 const urlLib = require('url');
 const fs = require('fs');
-const querystring = require('querystring');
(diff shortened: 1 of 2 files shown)


roblav96:
Can't argue that one :P

sindresorhus:
I like the idea in general, but sounds risky as there might be other subtle differences between them. In a month, Node.js 4 will be deprecated and I plan to target Node.js 8 then, which has `URLSearchParams`. So I'd rather wait on that, than risk breaking existing users now.

Hoishin:
@sindresorhus I see. That looks way better and safer solution. I'd be willing to work on it and do a PR for it.

PR #472, opened 2018-03-28
Do not override `Accept-Encoding` header if already provided

pietermees:
It's currently not possible to set the `Accept-Encoding` header to `gzip` only. We would like to do so because we've noticed some servers sending malformed `deflate` responses that Node is not able to inflate.

### DIFF
--- a/index.js
+++ b/index.js
@@ -527,8 +527,8 @@ function normalizeArguments(url, opts) {
 		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`
 	}, headers);
 
-	if (opts.decompress) {
-		opts.headers['accept-encoding'] = 'gzip,deflate';
+	if (opts.decompress && opts.headers['accept-encoding'] == null) {
+		opts.headers['accept-encoding'] = 'gzip, deflate';
 	}
 
 	const query = opts.query;
--- a/index.js
+++ b/index.js
@@ -527,7 +527,7 @@ function normalizeArguments(url, opts) {
 		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`
 	}, headers);
 
-	if (opts.decompress && opts.headers['accept-encoding'] == null) {
+	if (opts.decompress && !is.undefined(opts.headers['accept-encoding'])) {
 		opts.headers['accept-encoding'] = 'gzip, deflate';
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -527,8 +527,8 @@ function normalizeArguments(url, opts) {
 		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`
 	}, headers);
 
-	if (opts.decompress) {
-		opts.headers['accept-encoding'] = 'gzip,deflate';
+	if (opts.decompress && opts.headers['accept-encoding'] == null) {
```
brandon93s: This breaks existing functionality. `opts.headers['accept-encoding']` will always evaluate to false due to L522. This means the default headers will never be added.

You'd either need to fix the check, e.g. `if (opts.decompress && !is.undefined(opts.headers['accept-encoding'])` or update the assignment to keep existing values: `opts.headers['accept-encoding'] = opts.headers['accept-encoding'] || 'gzip, deflate';`.

timdp: Aren't you basically saying that `== null` isn't allowed and Pieter should instead use `is.undefined()`? I don't see how it would break existing functionality.

brandon93s: Line 531 is unreachable. The change needs the respect an existing header value, but it's preventing the default from being applied entirely.

pietermees: @brandon93s I think the confusion is the use of `== null`, which actually returns `true` if the value is `undefined` as well (not the `==` rather than `===`). That should result in the line being reachable and executing as expected.

In our internal code style policies this is the idiomatic way of doing such checks, but if you prefer another code style using `is.undefined`, `hasOwnProperty` or the `||` approach I'm happy to do either one :)

pietermees: And I see the test error is actually a code style error where it's instructing not to use `== null`, so I guess I just answered my own question ;)

brandon93s:
Whoops! My comment was wrong - shouldn't be negated.

pietermees:
@brandon93s this PR also changes the default value from `gzip,deflate` to `gzip, deflate`. At first I thought the space was required, but it seems that's only specified as such in the Fetch API (see https://fetch.spec.whatwg.org/#concept-header-value-combined) while the HTTP spec says that a simple `,` is fine. I thought it might make sense to align closer to fetch moving forward?

PR #473, opened 2018-04-02
use destructuring and spread operator

Hoishin:
Part of #418 

No logic changes


sindresorhus:
Thanks for the PR, but I had other changes I had to get in too, so was easier to do it all myself.

PR #479, opened 2018-04-16
Expose response body when return statusCode > 400

jas0ncn:
Need access response body when statusCode > 400, such as 500 Internal Server Error


jas0ncn:
oh, thanks a lot. It's my negligence...I'll close this pr.

PR #484, opened 2018-05-05
Using 'qs' library to parse and stringify request body

kushubham9:
This brings support for indices when parsing the request body which contains an array. 

```
qs.stringify({ a: ['b', 'c', 'd'] });
// 'a[0]=b&a[1]=c&a[2]=d'
```
Indices are required when sending a form request containing array on a server (Specially with PHP as a backend). Without indicies, the $_POST variable does not identify array in the request but takes the last value associated.

eg,
```
querystring.stringify({ a: ['b', 'c', 'd'] })
// a=b&a=c&a=d'

On the backend:
$_POST['a'] will fetch 'd'
```


sindresorhus:
No thanks. You can just use `qs` directly instead of the built-in query string handling if you need this.

PR #495, opened 2018-06-29
Support discrete client configuration via `got.create([options])`

knksmith57:
Add first-class support for discrete client configuration via new public `got.create([options])` method.

Feature is 100% backwards compatible and internal defaults have been refactored ever so slightly to dogfood the interface.

I did my best to adhere to existing code standards-- linted w/ `xo`, added tests to maintain coverage (statements + lines are slightly up, branches are slightly down), documented the method in `readme.md`, and tailored the commit message style to match the predominant pattern.

My first time contributing here so please let me know if there's anything else I can do to help! If y'all prefer to keep this out of core I'll understand and can publish elsewhere-- no hard feelings either way.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ const {PassThrough, Transform} = require('stream');
 const urlLib = require('url');
 const fs = require('fs');
 const querystring = require('querystring');
+const extend = require('extend');
 const CacheableRequest = require('cacheable-request');
 const duplexer3 = require('duplexer3');
 const toReadableStream = require('to-readable-stream');
@@ -27,6 +28,19 @@ const errors = require('./errors');
 
 const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);
+
+const defaults = {
+	retries: 2,
+	cache: false,
+	decompress: true,
+	useElectronNet: false,
+	throwHttpErrors: true,
+	headers: {
+		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`
+	}
+};
+
+const got = create(defaults);
 
 const isFormData = body => is.nodeStream(body) && is.function(body.getBoundary);
 
@@ -511,17 +525,8 @@ function normalizeArguments(url, opts) {
 		url = urlToOptions(url);
 	}
 
-	const defaults = {
-		path: '',
-		retries: 2,
-		cache: false,
-		decompress: true,
-		useElectronNet: false,
-		throwHttpErrors: true
-	};
-
 	opts = {
-		...defaults,
+		path: '',
 		...url,
 		protocol: url.protocol || 'http:', // Override both null/undefined with default protocol
 		...opts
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -641,36 +643,46 @@ function normalizeArguments(url, opts) {
 	return opts;
 }
 
-function got(url, opts) {
-	try {
-		const normalizedArgs = normalizeArguments(url, opts);
+function create(defaults = {}) {
+	function got(url, opts) {
+		try {
+			opts = extend(true, {}, defaults, opts);
```
jstewmon: It looks like the global defaults are not inherited? Was that intentional? My expectation would be that when I create a client, the global defaults are inherited.

I think you can easily address by renaming the function argument to `clientDefaults` and making this line:

```js
opts = extend(true, {}, defaults, clientDefaults, opts);
```

knksmith57: the global defaults are inherited as the top-level `got` export dogfoods this interface and all public `create()` calls operate off of that instance:
```js
const got = create(defaults);  // L43
```

knksmith57: heads up that, unless I'm missing something, this should be covered by the _preserve global defaults_ test case in `test/create.js` line 18

ewolfe:
I just pulled this down and it works great :) 

I found myself wanting to create an instance with a `hostname` option and using only the path as the argument to my instance. ie:

```js
const client = got.create({
  hostname: 'https://example.com',
  path: '/api',
  headers: {
    Authorization: `Bearer ${TOKEN}`,
  },
});

const response = await client('/hello-world'); // GET https://example.com/api/hello-world
```

What are your thoughts on adding that functionality? I don't think it's _too_ opinionated and is in the spirit of the tagline "Simplified HTTP requests"

jstewmon:
@ewolfe that feature is usually implemented with a discrete option `baseUrl` to disambiguate the caller's intent. I'd say that's a complementary feature to being able to create an instance, but deserving of its own PR, since it would add a config param.

knksmith57:
@ewolfe thanks for taking a look! Super glad you find the feature useful.

re simplified interface: I would love to get there with `got` and have a few ideas on how to support that sort of functionality without polluting the public API-- @jstewmon's suggestion to support something like `baseUrl` is definitely one way to do it.

For what it's worth, everything from your example snippet _except_ the URI prefixing is made possible by this feature (protocol + hostname + default headers).

If there's interest, I'd be happy to put together a proposal PR. Thanks again for taking a look!

knksmith57:
how do y'all feel about this @sindresorhus / @lukechilds / @brandon93s?

lukechilds:
Apologies, I'm very busy on another project right now.

I haven't reviewed the code in depth but I really like the idea. Much nicer than having to manually create a wrapper function and should make writing API clients with Got a simpler too.

sindresorhus:
This is a nice feature. I'll take a closer look this weekend and merge.

sindresorhus:
This looks great 🙌

It will probably be at least a week until I can publish this as I'm currently preparing for a big v9 release.

sindresorhus:
> I found myself wanting to create an instance with a hostname option and using only the path as the argument to my instance. ie:

@ewolfe That would be very handy. Can you open a new issue about it?

PR #496, opened 2018-07-03
Make it possible to unset the default `user-agent` header

### LINKED ISSUE #485: Opting out of default user-agent
Hi,

The `user-agent` header is optional in RFC 2616, it shouldn't be enforced imo. Currently there is no way to send a request without the `user-agent` header.

Happy to submit a PR to make this happen.

szmarczak:
Fixes #485. To disable `user-agent` header set it to `null` or `undefined`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -528,16 +528,17 @@ function normalizeArguments(url, opts) {
 	};
 
 	const headers = lowercaseKeys(opts.headers);
+	if (!headers.hasOwnProperty('user-agent')) { // eslint-disable-line no-prototype-builtins
+		headers['user-agent'] = `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`;
+	}
+
 	for (const [key, value] of Object.entries(headers)) {
 		if (is.nullOrUndefined(value)) {
 			delete headers[key];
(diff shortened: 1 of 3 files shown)


sindresorhus:
Great! Can you document the behavior here https://github.com/sindresorhus/got#user-agent and here https://github.com/sindresorhus/got#api

szmarczak:
Done. Please see if it's good :unicorn:

sindresorhus:
@szmarczak Your change seems to fail on master. Could you take a look?

szmarczak:
I just wanna note it has passed all tests before commit https://github.com/sindresorhus/got/commit/8103bc57b9a6e3418bff5cc8808d43c7ea04e944. I'll look at it now.

szmarczak:
Weird. Suddenly it fails. I'll send a PR.

szmarczak:
Gotcha! :unicorn:
`require('extend')(true, {}, { x: undefined })` -> `{}`

Not my fault. I had sent this PR before `got` has used `extend` dependency.
058452b011ae2285697ef1b408ab5e8a4a9be0ae broke that.

PR #499, opened 2018-07-05
Fix merging defaults and options

szmarczak:
Fixes commit 058452b.


sindresorhus:
Sorry, I moved the source files into a `source` directory earlier... Could you update the PR?

source/index.js
```diff
@@ -654,10 +645,25 @@ function normalizeArguments(url, options) {
 	return options;
 }
 
+function assignOptions(defaults, options = {}) {
```
sindresorhus: The `= {}` is no longer needed here.

szmarczak: Nope. Tests have failed :laughing:

sindresorhus: Ugh, yeah I didn't notice `options` was used outside the `extend` call...

PR #500, opened 2018-07-06
Split `index.js` into separate files

brandon93s:
Splits the got codebase into its sub-components.


test/error.js
```diff
@@ -90,10 +90,13 @@ test.serial('http.request error', async t => {
 });
 
 test.serial('catch error in mimicResponse', async t => {
+	const mimicResponse = () => {
+		throw new Error('Error in mimic-response');
+	};
+	mimicResponse['@global'] = true;
```
brandon93s: Not best practice to use `@global` with `proxyquire` but it does not _currently_ result in any side-effects in the tests. An alternative would be to introduce a `mimic-response-wrapper.js` that exports a named function we can mock with `Sinon` and then use that in got.

sindresorhus: Nah, I think what you have here is fine.

sindresorhus:
Really good! 🙌

alextes:
@brandon93s some part of me had been wanting to do this PR for soooo long haha.
I'm a tiny bit jealous and mostly very thankful you made a long time wish of mine come true 👏 😄 !

PR #308, opened 2017-05-17
Request timeout

djmadeira:
Fixes #257.

Changes timeout option so that if a number is supplied, that is the timeout for the entire request to be completed; if an object is supplied, the keys `connect`, `socket`, and `request` set connection, socket, and entire request timeouts, respectively.

Open questions:
- Is "request" the right name for the full request timeout option? Should it be "response" instead?

PR #501, opened 2018-07-06
Fix timeouts

### LINKED ISSUE #344: Cannot cancel when timeout option is set

### LINKED ISSUE #478: Retries not working
Hi,
I have added the following request options:
```
{
      timeout: 3000, // request timeout in milliseconds
      json: true,
      followRedirect: false,
      throwHttpErrors: true,
      retries: (retry, error) => {
        if (retry > 3) {
          // got promise will be rejected with `error`
          return 0
        }
        console.log('Retry Attempt')
        return 1000 * Math.pow(2, retry) + Math.random() * 100
      }
}
```

I was expecting a retry in case of REQUEST TIMEOUT, or any event of failure. However, retries are not working. 
There is nothing getting logged on the console for me. 

Am I missing something?

szmarczak:
TODO: need to check how it works with streams. Probably broken.

Fixes #478.
Fixes #344.

I just have improved [timed-out](https://github.com/floatdrop/timed-out).

### DIFF
--- a/source/timed-out.js
+++ b/source/timed-out.js
@@ -1,0 +1,76 @@
+'use strict';
+
+module.exports = function (req, time) {
+	if (req.timeoutTimer) {
+		return req;
+	}
+
+	const delays = isNaN(time) ? time : {socket: time, connect: time};
+	const host = req._headers ? (' to ' + req._headers.host) : '';
+
+	function throwESOCKETTIMEDOUT() {
+		req.abort();
+		const e = new Error('Socket timed out on request' + host);
+		e.code = 'ESOCKETTIMEDOUT';
+		req.emit('error', e);
+	}
+
+	function throwETIMEDOUT() {
+		req.abort();
+		const e = new Error('Connection timed out on request' + host);
+		e.code = 'ETIMEDOUT';
+		req.emit('error', e);
+	}
+
+	if (delays.connect !== undefined) {
+		req.timeoutTimer = setTimeout(throwETIMEDOUT, delays.connect);
+	}
+
+	if (delays.request !== undefined) {
+		req.requestTimeoutTimer = setTimeout(() => {
+			clear();
+
+			if (req.connection.connecting) {
+				throwESOCKETTIMEDOUT();
+			} else {
+				throwETIMEDOUT();
+			}
+		}, delays.request);
+	}
+
+	// Clear the connection timeout timer once a socket is assigned to the
+	// request and is connected.
+	req.on('socket', socket => {
+		// Socket may come from Agent pool and may be already connected.
+		if (!(socket.connecting || socket._connecting)) {
+			connect();
+			return;
+		}
+
+		socket.once('connect', connect);
+	});
+
+	function clear() {
+		if (req.timeoutTimer) {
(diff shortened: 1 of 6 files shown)


sindresorhus:
Can you explain the change?

szmarczak:
> Can you explain the change?

What it does: it wraps `p-timeout`. I've changed that so it doesn't.

Please see the issues. The timeout has began since `asPromise()` was called. Until we make a request (real request, `http(s).request`) to something it may pass a few ms. Also, it doesn't retry on ETIMEDOUT because `p-timeout` throws that and rejects the promise. Merged request timeout into `timed-out`. :tada: 

https://github.com/sindresorhus/got/blob/ab5e8e1509d95b01756d40546eea0868849a1d35/source/as-promise.js#L10-L13

Forgot to add `t.is(tried, true);` to tests :smile: So this is why it has passed them without this change :boom:

sindresorhus:
Looks good. Nice work @szmarczak :)

source/timed-out.js
```diff
@@ -0,0 +1,78 @@
...
+	function connect() {
+		clear();
+
+		if (delays.socket !== undefined) {
+			// Abort the request if there is no activity on the socket for more
+			// than `delays.socket` milliseconds.
+			req.setTimeout(delays.socket, throwESOCKETTIMEDOUT);
+		}
+
+		req.on('end', () => {
```
jstewmon: I don't think this is right. I think the `request` timeout was introduced in https://github.com/sindresorhus/got/pull/308, which describes it as the time for the entire response to be received.

`'end'` is not an event for `http.ClientRequest`, is it? It looks like tests around this are contrived to pass - I don't see a test that shows the timeout is not raised if the request (and response) completes before the timeout.

I think this should be something like:

```js
req.on('response', res => {
  res.on('end', () => {
    clearTimeout(req.requestTimeoutTimeout)
  })
})

szmarczak: Fixin' right now. You have really sharp eyes!

szmarczak: > It looks like tests around this are contrived to pass

Make another test and increase the timeout so the request finishes. Why hadn't you checked that? It passes. It doesn't throw, because the promise has been resolved already.

szmarczak: > `await delay(reqDelay * 3);`

If you used a Promise and attached some listeners, it would throw with that too.

The point is no one has added a delay because everybody assumed if the request finishes without any errors, it won't throw then. That's just an oversight. If you tell that I just did this intentionally, then you're wrong.

sindresorhus: @jstewmon Could you submit the above test?

jstewmon: Tests that have setup are always contrived to pass - I should have just said that tests are missing for cases where the timeout is not breached. I didn't mean that you deliberately introduced a bug. :-)

Issue #203, opened 2016-06-08
Endpoint prefix / default config


jamestalmage:
I think it would be cool if there was an easy way to do something like:

``` js
var github = got.endpoint('https://api.github.com', {agent: ...});

// github has the same API as got, except you don't need the base path
// and certain default options (i.e. `agent`) are already defined
github('/foo', {});
```

This main advantage would be for testing:

``` js
function MyGitHubApi() {
  this._gh = got.endpoint('https://api.github.com', {agent: ...});
}

MyGitHubApi.getFoo = function () {
  return this._gh('/foo');
}

// later in tests:

var api = new MyGitHubApi();
api._gh.override('http://localhost:3000');
```

floatdrop:
We have a lot files, that wraps `got` inside and give it some default values. I would love to see a package `got-api` with some sweet stuff for generating cool API wrappers.

szmarczak:
Ability to add new options would be cool too.

szmarczak:
Here it comes: #503

PR #503, opened 2018-07-06
Simplify wrapping `got`

szmarczak:
Fixes https://github.com/sindresorhus/got/issues/203.

Real example:

https://gist.github.com/szmarczak/6255d5fa9e2f0f0e15ef2264be867478


sindresorhus:
I'm a little bit worried that we're exposing too many internals now, especially `got.normalizeArguments`.

sindresorhus:
What's the difference between `.fork()` and `.create()`?

sindresorhus:
I really like how much it simplifies `gh-got` though. I'm sure it will enable more useful niche versions of Got.

szmarczak:
> What's the difference between .fork() and .create()?

Fork - forks the options from parent instance (squashes them into a new piece, performs `assignOptions` again), replaces `methods` and `handler` if provided.

Create - imagine there are no options, no methods, no handler. You set it up from scratch. Like another got, not inherited from any instance.

sindresorhus:
We should list `fork()` first then, as that's what most end-users would want. They just want to tweak some of the defaults.

sindresorhus:
A common use-case is just overriding a few defaults when creating a new instance. I don't like how you now need to specify a nested object to set a few options.

I think `got.fork()` should be optimized for the common-case and only allow overriding options, not handler/methods. If users need more advanced control, they can use `got.create()` and pass in the exposed `got.defaults`.

So `got.fork()`:

```js
const client = got.fork({headers: {'x-foo': 'bar'});
```

Also, is `fork` the best word for this? Some alternatives would be `extend()` or `defaults()`. Not saying it's not. Just putting it out there.

source/as-stream.js
```diff
@@ -4,47 +4,49 @@ const duplexer3 = require('duplexer3');
 const is = require('@sindresorhus/is');
 const requestAsEventEmitter = require('./request-as-event-emitter');
 const {HTTPError, ReadError, RequestError} = require('./errors');
+const normalizeArguments = require('./normalize-arguments');
 
-module.exports = options => {
-	options.stream = true;
+module.exports = (url, options) => {
+	const normalizedArgs = normalizeArguments(url, options);
+	normalizedArgs.stream = true;
```
sindresorhus: `normalizedArgs` => `normalizedOptions` would be more accurate, but that's quite verbose.

What's wrong with just:

```js
options = normalizeArguments(url, options);
```

Do we need the original options somewhere?

szmarczak: Fixed that.

source/index.js
```diff
@@ -3,13 +3,26 @@ const pkg = require('../package.json');
...
 const defaults = {
-	retries: 2,
-	cache: false,
-	decompress: true,
-	useElectronNet: false,
-	throwHttpErrors: true,
-	headers: {
-		'user-agent': `${pkg.name}/${pkg.version} (https://github.com/sindresorhus/got)`
+	handler: (url, options, next) => {
+		return next(url, options);
```
szmarczak: I think `handler` could use `next` directly. Like `handler: next`

source/create.js
```diff
@@ -1,64 +1,56 @@
...
 
-	if (Reflect.has(options, 'headers')) {
-		for (const [key, value] of Object.entries(options.headers)) {
-			if (is.nullOrUndefined(value)) {
-				delete opts.headers[key];
-				continue;
-			}
-		}
+	if (options.endpoint) {
+		url = /^https?/.test(path) ? path : options.endpoint + path;
```
szmarczak: This should be:

source/create.js
```diff
@@ -1,64 +1,56 @@
...
-			return asPromise(normalizedArgs);
+			options = assignOptions(defaults.options, options);
+			return defaults.handler(url, options, next);
 		} catch (error) {
 			return Promise.reject(error);
 		}
 	}
 
-	got.create = (options = {}) => create(assignOptions(defaults, options));
+	got.create = newDefaults => create(newDefaults);
```
szmarczak: This should be:

readme.md
```diff
@@ -99,6 +99,14 @@ Properties from `options` will override properties in the parsed `url`.
 
 If no protocol is specified, it will default to `https`.
 
+##### endpoint
+
+Type: `string` `Object`
+
+When specified, `url` will be preceded by `endpoint`.
+
```
sindresorhus: Can you also mention that this one is especially useful with `got.extend()` to create niche specific Got instances?

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
+Configure a new `got` instance with provided settings:
+
+##### [options](#options)
+
+To inherit from parent, set it as `got.defaults.options` or use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).
+
+##### methods
+
+Type: `object`
```
sindresorhus: `Object`

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
+
+Array of supported request methods.
+
+To inherit from parent, set it as `got.defaults.methods`.
+
+##### handler
+
+Type: `function`<br>
```
sindresorhus: `Function`

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
+Default: `undefined`
+
+Function making additional changes to the request.
+
+To inherit from parent, set it as `got.defaults.handler`.<br>
+To use the default handler, set it as `null` or `undefined`.
```
sindresorhus: Couldn't this just be: `To use the default handler, just omit specifying this.`?

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
 
+#### got.create(settings)
+
+Example: [gh-got](https://github.com/sindresorhus/gh-got/blob/master/index.js)
+
+Configure a new `got` instance with provided settings:
+
+##### [options](#options)
+
+To inherit from parent, set it as `got.defaults.options` or use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).
```
sindresorhus: > use [destructuring assignment]

Did you mean to say "object spread"?

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
+Function making additional changes to the request.
+
+To inherit from parent, set it as `got.defaults.handler`.<br>
+To use the default handler, set it as `null` or `undefined`.
+
+###### [url](#url)
+
+###### [options](#options)
+
+###### next
```
sindresorhus: `next` => `next()`

sindresorhus:
Can you add a small tip to the Tips section (https://github.com/sindresorhus/got#tips) on how to use the `endpoint` option with `got.extend()` to quickly get an endpoint specific Got instance? I think that's a common need, so would be good to have a clear example for that specific thing.

source/create.js
```diff
@@ -1,64 +1,60 @@
...
 
-	if (Reflect.has(options, 'headers')) {
-		for (const [key, value] of Object.entries(options.headers)) {
-			if (is.nullOrUndefined(value)) {
-				delete opts.headers[key];
-				continue;
-			}
-		}
+	if (options.endpoint && !/^https?/.test(path)) {
+		url = options.endpoint + path;
```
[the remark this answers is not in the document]
szmarczak: Yeah, I was using that in my app but forgot about it :D

readme.md
```diff
@@ -302,23 +312,134 @@ Sets `options.method` to the method name and makes a request.
...
 	//=> headers2['x-foo'] === 'bar'
 	//=> headers2['x-baz'] === 'qux'
 })();
 ```
 
+#### got.create(settings)
+
+Example: [gh-got](https://github.com/sindresorhus/gh-got/blob/master/index.js)
+
+Configure a new `got` instance with provided settings:
```
sindresorhus: We should be more explicit here about `got.create()` having no good defaults included by default and how it's different from `got.extend()`.

readme.md
szmarczak: Yep, naming is mixed up in my head sometimes

brandon93s:
I'm a bit concerned that a user is going to be confused as to whether they should be using `get.extend()` or `got.create()` for their use case. Most users are just going to want to provide new defaults, but `extend` has a small blurb in the docs compared to the in-depth `create` section which could mislead them. `create` seems to cater primarily to library / module creation (e.g. gh-got) and not everyday use. It may be time to consider breaking out some advanced sections of the docs into separate files and linking to them to avoid information overload:

> Need more control over the behavior of got? Check out the [advanced creation](#) options.

Looking elsewhere, axios has `axios.create()` and request has `request.defaults()` which are functionally equivalent to `got.extend()`. Users coming from axios may reach for `create` and have a very different experience.

readme.md
```diff
@@ -99,6 +99,15 @@ Properties from `options` will override properties in the parsed `url`.
 
 If no protocol is specified, it will default to `https`.
 
+##### baseUrl
+
+Type: `string` `Object`
+
+When specified, `url` will be preceded by `baseUrl`.<br>
```
brandon93s: "will be _prepended_ with `baseUrl`" is a bit more clear.

Should we note that this only applies to relative URLs? If you specify an absolute URL it will skip the `baseUrl`.

brandon93s: The placement in the docs has this as a separate parameter to the `got` api. I think it's just part of the options, however. If that's the case it should be moved into the options section below.

readme.md
```diff
@@ -302,23 +313,135 @@ Sets `options.method` to the method name and makes a request.
 
 #### Instances
 
-#### got.create([options])
+#### got.extend([options])
 
-Configure a new `got` instance with default `options`:
+Configure a new `got` instance with default `options` and custom `baseUrl` (optional):
```
brandon93s: See other notes, `baseUrl` _is_ an option. This feels redundant.

readme.md
```diff
@@ -302,23 +313,135 @@ Sets `options.method` to the method name and makes a request.
...
 
 ```js
 (async () => {
```
brandon93s: This example is a bit unclear if you're not familiar with `httpbin.org` and knowing that this endpoint will reflect headers back in the body. Maybe something like this would help illustrate: 

```js
const client = got.extend({
	baseUrl: 'https://example.com',
	headers: {
		'x-unicorn': 'rainbow'
	}
});

client.get('/demo')

/* HTTP Request => 
 * GET /demo HTTP/1.1
 * Host: example.com
 * x-unicorn: rainbow
 */
```

source/create.js
```diff
@@ -1,64 +1,61 @@
...
-	];
-
-	for (const method of methods) {
+	for (const method of defaults.methods) {
 		got[method] = (url, options) => got(url, {...options, method});
 		got.stream[method] = (url, options) => got.stream(url, {...options, method});
 	}
 
 	Object.assign(got, errors);
+	Object.assign(got, {defaults});
```
brandon93s: Exposing this and having it used in the closure could lead to some unexpected behavior if defaults are modified: 

```js
test('tampering with defaults', async t => {
	const instance = got.create({
		handler: got.defaults.handler,
		methods: got.defaults.methods,
		options: { ...got.defaults.options,
			baseUrl: 'example'
		}
	});

	const instance2 = instance.create({
		handler: instance.defaults.handler,
		methods: instance.defaults.methods,
		options: instance.defaults.options
	});

	// Tamper Time
	instance.defaults.options.baseUrl = 'http://google.com'

	t.is(instance.defaults.options.baseUrl, instance2.defaults.options.baseUrl);
});
```

I think I'd expect two instances created with `.create` to be independent and not impact each other. Similar tampering could be performed on the global `got.defaults` that the module exports.

Thoughts?

readme.md
szmarczak: Yes, it's an option. Done in next commit.

source/create.js
szmarczak: You're right. We should deep freeze those or clone.

readme.md
szmarczak: Oh wait, you mean users may not be familiar with `httpbin.org`? Then yes too. I'll replace that.

sindresorhus:
@brandon93s Good point. Let's move it out into a separate file like you've described.

source/create.js
sindresorhus: We can deep clone it with the `extend` dependency, I think.

szmarczak: Deep cloning may lead to false information because it's still writable. Used deep freeze.

szmarczak:
@brandon93s I would leave the names unchanged. Previously `got.extend()` was named `got.fork()`. I would just note in `readme.md`/`xxx.md` the difference between `got.create()` and `got.extend()`. I can rename it anytime if it's needed.

sindresorhus:
@szmarczak I don't think @brandon93s was suggesting we rename it, but rather just pointing out the benefit of moving `got.create()` out of the readme.

szmarczak:
@sindresorhus So I'll do that tomorrow then :unicorn:

source/create.js
```diff
@@ -6,6 +6,18 @@ const asStream = require('./as-stream');
 const asPromise = require('./as-promise');
 const normalizeArguments = require('./normalize-arguments');
 
+const deepFreeze = obj => {
+	const keys = Object.keys(obj);
+
+	for (let i = 0; i < keys.length; i++) {
```
sindresorhus: Use a `for-of` loop and `Object.entries()`.

source/create.js
```diff
@@ -6,6 +6,18 @@ const asStream = require('./as-stream');
...
+		if (typeof obj[keys[i]] === 'object' && obj[keys[i]] !== null) {
```
sindresorhus: Use the `is` module here: `is.object`

source/create.js
szmarczak: Wouldn't it be better: `for-of` and `Object.keys`? I don't see much difference.

sindresorhus: `Object.entries()` is better as it gives you the value too right away, instead of having to do `options.headers[key]`.

advanced-creation.md
```diff
@@ -0,0 +1,163 @@
+# Advanced creation
+
+> Make calling REST APIs easier by creating niche-specific `got` instances.
+
+#### got.extend([options])
```
sindresorhus: `got.extend()` should stay in the readme, it was only `got.create()` that supposed to be moved out.

szmarczak: Oh, right.

sindresorhus:
This is perfect now. Amazing work @szmarczak 🙌

leo-celebrate

PR #508, opened 2018-07-09
Support retrying on some HTTP status codes and generally improve the retry functionality

### LINKED ISSUE #379: [proposal] Don't retry if no error.code
I was thinking about my previous issue with testing #187 reading the outcome of #222 and I was looking at the source of [`is-retry-allowed`](https://github.com/floatdrop/is-retry-allowed/blob/master/index.js#L59) and the [`aws-sdk`](https://github.com/aws/aws-sdk-js/blob/1edbf0fcc910d7e5f834c75ae317a16130cfa814/lib/util.js#L844) and I realized that we retry on unknown errors, while AWS and possibly others, do not.

I'd like to propose to only retry if we have a known retryable error code.

(_note_: I considered moving this upstream to `is-retry-allowed` but I think it's a `got` concern first and foremost. It's weird we have a default true and a whitelist in that module.)

(_related_: https://github.com/node-nock/nock/pull/898 I've been pushing for this for [`nock`](https://github.com/node-nock/nock/) because it's really annoying when developing, but I'm starting to think got should solve this. I'm not 100% yet)

### LINKED ISSUE #417: Retry on some non-2xx HTTP status codes

szmarczak:
Fixes #417 
Fixes #379


readme.md
```diff
@@ -192,13 +192,22 @@ This also accepts an object with separate `connect`, `socket`, and `request` fie
...
+	methods: ['GET', 'PUT', 'HEAD', 'DELETE', 'OPTIONS', 'TRACE'],
+	statusCodes: [408, 413, 429, 502, 503, 504]
+}
+```
 
 Number of request retries when network errors happens. Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
 
 Option accepts `function` with `retry` and `error` arguments. Function must return delay in milliseconds (`0` return value cancels retry).
 
+It accepts `object` with optional `retry`, `method`, `statusCodes` fields for time until retry, allowed methods, and allowed status codes.
```
sindresorhus: `method` => `methods`

readme.md
```diff
@@ -192,13 +192,22 @@ This also accepts an object with separate `connect`, `socket`, and `request` fie
...
 
-Type: `number` `Function`<br>
-Default: `2`
+Type: `number` `Function` `Object`<br>
+Default:
+```js
+{
+	retry: 2,
+	methods: ['GET', 'PUT', 'HEAD', 'DELETE', 'OPTIONS', 'TRACE'],
+	statusCodes: [408, 413, 429, 502, 503, 504]
```
sindresorhus: I think we should link up the status codes (like I did in the issue), so people can quickly check what they mean. Would save them some time.

source/normalize-arguments.js
```diff
@@ -108,14 +108,31 @@ module.exports = (url, options) => {
...
 
-		options.retries = (iter, error) => {
-			if (iter > retries || !isRetryAllowed(error)) {
+		options.gotRetries.retry = (iter, error) => {
+			if (iter > retry || (!isRetryAllowed(error) && !options.gotRetries.methods.includes(error.method) && !options.gotRetries.statusCodes.includes(error.statusCode))) {
 				return 0;
 			}
 
+			if (error.statusCode === 413 && Reflect.has(error.headers, 'retry-after')) {
+				const after = Number(error.headers['retry-after']);
```
sindresorhus: Keep in mind this comment: https://github.com/sindresorhus/got/issues/417#issuecomment-364760887 We need to set some sort of upper limit (and document it). What kind of limit do you think makes sense?

szmarczak: 1. There's limit of redirects set to 10.
2. We can set a timeout, if it exceeds the timeout then throw.

> We need to set some sort of upper limit (and document it). What kind of limit do you think makes sense?

Indeed. It should be the `retries` option.

readme.md
```diff
@@ -192,13 +192,18 @@ This also accepts an object with separate `connect`, `socket`, and `request` fie
 
 ###### retries
 
-Type: `number` `Function`<br>
-Default: `2`
+Type: `number` `Function` `Object`<br>
```
sindresorhus: I think we should remove support for `retries: Function` and just require `retries: { retry: Function} }`.

readme.md
```diff
@@ -192,13 +192,18 @@ This also accepts an object with separate `connect`, `socket`, and `request` fie
...
+Default:
+- retry: `2`
+- methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
+- statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
 
 Number of request retries when network errors happens. Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
```
sindresorhus: It's no longer just network errors.

source/as-stream.js
```diff
@@ -10,6 +10,8 @@ module.exports = options => {
 	const output = new PassThrough();
 	const proxy = duplexer3(input, output);
 
+	options.gotRetries.retry = () => 0;
```
sindresorhus: Why is this not just `options.gotRetries = 0;`?

source/normalize-arguments.js
```diff
@@ -108,14 +108,31 @@ module.exports = (url, options) => {
...
+				if (is.number(after)) {
+					return after * 1000;
+				}
+
+				return Math.max((new Date(error.headers['retry-after'])).getTime() - Date.now(), 0);
```
sindresorhus: You can use `Date.parse(error.headers['retry-after'])` instead of `(new Date(error.headers['retry-after'])).getTime()`

test/retry.js
```diff
@@ -6,6 +6,7 @@ let s;
 let trys = 0;
 let knocks = 0;
 let fifth = 0;
+let lastTried = Date.now();
```
sindresorhus: This needs to be named more specific. `lastTried` => `lastTried413access`

test/retry.js
```diff
@@ -71,6 +86,38 @@ test('falsy value prevents retries #2', async t => {
 	t.truthy(err);
 });
 
+test('custom retries', async t => {
+	let tried = false;
+	const err = await t.throws(got(`${s.url}/500`, {
+		throwHttpErrors: true,
+		retries: {
+			retry: iter => {
```
sindresorhus: `iter` - Use the full word.

test/retry.js
```diff
@@ -71,6 +86,38 @@ test('falsy value prevents retries #2', async t => {
...
+				}
+
+				return 0;
+			}, methods: [
+				'GET'
+			], statusCodes: [
+				500
+			]
+		}
+	}));
```
sindresorhus: Use the new `t.throws` API

test/timeout.js
```diff
@@ -132,6 +132,11 @@ test('retries on timeout (ETIMEDOUT)', async t => {
 });
 
 test('timeout with streams', async t => {
-	const stream = got.stream(s.url, {timeout: 0, retries: 0});
+	const stream = got.stream(s.url, {
+		timeout: 0,
+		retries: () => {
+			t.fail('Retries on streams');
+		}
```
sindresorhus: Don't reuse tests for different purposes. This should be a new test.

test/retry.js
```diff
@@ -71,6 +86,38 @@ test('falsy value prevents retries #2', async t => {
...
+	t.true(tried);
+});
+
+test('respect 413 Retry-After', async t => {
+	const err = await got(`${s.url}/413`, {
+		throwHttpErrors: false,
+		retries: 1
+	});
+	t.is(err.statusCode, 413);
+	t.true(Number(err.body) >= 2000);
```
sindresorhus: Not immediately clear from reading the code what `2000` is. Make it a constant like `TOTAL_REQUEST_TIME` or something.

sindresorhus:
This needs more tests as there are quite a few moving pieces here.

- Needs a test to see what happens when you specify an empty array for `methods`. Same with `statusCodes`.
- Some more variations of the options.

sindresorhus:
This is looking very good. I'm excited to see this land.

source/normalize-arguments.js
sindresorhus: I was referring to a time limit, not redirect limit.

sindresorhus: > We can set a timeout, if it exceeds the timeout then throw.

sindresorhus: > We can set a timeout, if it exceeds the timeout then throw.

What should the timeout be? Should it be configurable?

How does retry interact with the set timeout?

readme.md
szmarczak: That's not needed IMO.

source/as-stream.js
szmarczak: The tests would fail then, because `normalizeArguments` gets called before and makes a function. Calling `options.gotRetries.retry(++retryCount, error)` would throw because `options.gotRetries.retry` is not a `function`. So it must be done like that.

sindresorhus: 👍

readme.md
sindresorhus: I don't think to define a custom method is that common and it would be nice to simplify the options. Right now it's quite confusing, with IMHO too many variations. Then it would just be either a number (common case) or object (full power).

sindresorhus: @brandon93s Thoughts?

source/normalize-arguments.js
szmarczak: > What should the timeout be? Should it be configurable?

Yes, we should specify `retries: { ..., timeout: number }`. That would override the `timeout` option. By default it should be set to `15000` or `30000` I think.

> How does retry interact with the set timeout?

It just makes a new request and uses the timeout option provided by user.
If got 413 status code, then it overrides the timeout according to the `Retry-After` header (if exists). I think I should note that in the docs.

test/retry.js
szmarczak: custom retries

  /home/szymon/Desktop/retries-patch/test/retry.js:93

   92:   let tried = false;
   93:   await t.throws(got(`${s.url}/500`, {
   94:     throwHttpErrors: true,

  The second argument to `t.throws()` contains unexpected properties

  Called with:

  {
    statusCode: 500,
  }

source/normalize-arguments.js
sindresorhus: > Yes, we should specify retries: { ..., timeout: number }. That would override the timeout option. By default it should be set to 15000 or 30000 I think.

So why wouldn't the user just change the `timeout` option themselves instead then? If `options.retries.timeout` just overrides `options.timeout`.

> I think I should note that in the docs.

Yes

szmarczak: > So why wouldn't the user just change the timeout option themselves instead then? If options.retries.timeout just overrides options.timeout.

test/retry.js
sindresorhus: Oh right... Never mind then.

readme.md
```diff
@@ -190,15 +190,20 @@ Milliseconds to wait for the server to end the response before aborting request
...
+Type: `Object`<br>
+Default:
+- retries: `2`
+- methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
+- statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
 
-Number of request retries when network errors happens. Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
+Number of request retries when network errors happens or receives specified status code and method. Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
 
 Option accepts `function` with `retry` and `error` arguments. Function must return delay in milliseconds (`0` return value cancels retry).
```
sindresorhus: This needs to be updated

source/is-retry-allowed.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const WHITELIST = [
```
sindresorhus: Use a `Set` here

source/is-retry-allowed.js
```diff
@@ -0,0 +1,18 @@
...
+	'ETIMEDOUT',
+	'ECONNRESET',
+	'EADDRINUSE',
+	'ESOCKETTIMEDOUT',
+	'ECONNREFUSED',
+	'EPIPE'
```
sindresorhus: These need to be documented in the readme.

source/normalize-arguments.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 const URLSearchParamsGlobal = typeof URLSearchParams === 'undefined' ? require('url').URLSearchParams : URLSearchParams; // TODO: Use the `URL` global when targeting Node.js 10
 const is = require('@sindresorhus/is');
-const isRetryAllowed = require('is-retry-allowed');
 const toReadableStream = require('to-readable-stream');
 const urlParseLax = require('url-parse-lax');
+const isRetryAllowed = require('./is-retry-allowed');
```
sindresorhus: `isRetryAllowed` => `isRetryOnNetworkErrorAllowed` would be more correct.

source/normalize-arguments.js
brandon93s: > What should the timeout be? Should it be configurable?

It could be the already configured timeout. If `retry-after` > `options.gotTimeout.request`, then don't retry. Not sure I'd add a new timeout until it's needed.

source/normalize-arguments.js
```diff
@@ -108,14 +108,31 @@ module.exports = (url, options) => {
...
+		options.gotRetries = {...options.gotRetries, ...options.retries};
+	}
+	delete options.retries;
+
+	if (!is.function(options.gotRetries.retry)) {
+		const {retry} = options.gotRetries;
 
-		options.retries = (iter, error) => {
-			if (iter > retries || !isRetryAllowed(error)) {
+		options.gotRetries.retry = (iter, error) => {
```
brandon93s: `iter` here, too. Full word?

source/normalize-arguments.js
szmarczak: > Be aware Number(null) is 0, which would cause an immediate retry. Maybe not worth coding for, though.

test/retry.js
```diff
@@ -54,7 +54,7 @@ test.before('setup', async () => {
 });
 
 test('works on timeout error', async t => {
-	t.is((await got(`${s.url}/knock-twice`, {timeout: {connect: 100, socket: 100}})).body, 'who`s there?');
+	t.is((await got(`${s.url}/knock-twice`, {timeout: {connect: 500, socket: 100}})).body, 'who`s there?');
```
sindresorhus: `500` and `100` should be constants at the top (with descriptive names) and used here and elsewhere. Makes it much clearer and easier to change.

szmarczak: Indeed.

readme.md
```diff
@@ -190,16 +190,31 @@ Milliseconds to wait for the server to end the response before aborting request
...
+If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
-Number of request retries when network errors happens. Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
+Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 0).
 
-Option accepts `function` with `retry` and `error` arguments. Function must return delay in milliseconds (`0` return value cancels retry).
+Option `retries` can be a `number`, but also accepts a `function` with `retry` and `error` arguments. Function must return delay in milliseconds (`0` return value cancels retry).
 
-**Note:** if `retries` is `number`, `ENOTFOUND` and `ENETUNREACH` error will not be retried (see full list in [`is-retry-allowed`](https://github.com/floatdrop/is-retry-allowed/blob/master/index.js#L12) module).
+**Note:** It retries only on specified methods & status codes **or** on these network errors:
```
sindresorhus: `methods` => `HTTP methods`

sindresorhus: And shouldn't `**or**` be `**and**`?

source/normalize-arguments.js
```diff
@@ -1,12 +1,14 @@
...
 const URLSearchParamsGlobal = typeof URLSearchParams === 'undefined' ? require('url').URLSearchParams : URLSearchParams; // TODO: Use the `URL` global when targeting Node.js 10
 const is = require('@sindresorhus/is');
-const isRetryAllowed = require('is-retry-allowed');
 const toReadableStream = require('to-readable-stream');
 const urlParseLax = require('url-parse-lax');
+const isRetryOnNetworkErrorAllowed = require('./is-retry-allowed');
 const urlToOptions = require('./url-to-options');
 const isFormData = require('./is-form-data');
 
+const RetryAfterStatusCodes = new Set([413, 429, 503]);
```
sindresorhus: Should either be `retryAfterStatusCodes` or `RETRY_AFTER_STATUS_CODES`

source/normalize-arguments.js
```diff
@@ -1,12 +1,14 @@
 'use strict';
 const URLSearchParamsGlobal = typeof URLSearchParams === 'undefined' ? require('url').URLSearchParams : URLSearchParams; // TODO: Use the `URL` global when targeting Node.js 10
 const is = require('@sindresorhus/is');
-const isRetryAllowed = require('is-retry-allowed');
 const toReadableStream = require('to-readable-stream');
 const urlParseLax = require('url-parse-lax');
+const isRetryOnNetworkErrorAllowed = require('./is-retry-allowed');
```
sindresorhus: The filename needs to be renamed too.

readme.md
```diff
@@ -190,16 +190,31 @@ Milliseconds to wait for the server to end the response before aborting request
 
 This also accepts an object with separate `connect`, `socket`, and `request` fields for connection, socket, and entire request timeouts.
 
-###### retries
+###### retry
 
-Type: `number` `Function`<br>
-Default: `2`
+Type: `Object`<br>
```
sindresorhus: It should still have `number` here.

readme.md
```diff
@@ -190,16 +190,31 @@ Milliseconds to wait for the server to end the response before aborting request
...
+Default:
+- retries: `2`
+- methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
+- statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
+- maxRetryAfter: `undefined`
+
+Object representing `retry`, `methods`, `statusCodes` and `maxRetryAfter` fields for time until retry, allowed methods, allowed status codes and maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time.
```
sindresorhus: representing `retry` => representing `retries`

sindresorhus:
Don't forget that `413` should *only* be retried if the `Retry-After` header is valid, while `429` and `503` can be retried even without the `Retry-After` header, but should adhere to it if it exists. I would also like to see a test for this; One that proves `413` is not retried without a `Retry-After` header and one that proves `429` and `503` are retried even without a `Retry-After` header).

readme.md
sindresorhus: The way it's written makes it sound like if you specify methods it will not retry on network errors.

We can improve it to:

> It retries only on the specified methods, status codes, and on these network errors:

sindresorhus:
This is super awesome! Huge 🙌 to @szmarczak for implementing this very complicated logic.

sindresorhus:
@brandon93s Please do give this a look and comment here if you see anything I missed ;)

PR #509, opened 2018-07-09
Fix the behavior of timeouts

szmarczak:
Did you know that connecting to a local socket can take >40ms? :fearful:

### DIFF
--- a/source/timed-out.js
+++ b/source/timed-out.js
@@ -7,7 +7,7 @@ module.exports = function (req, time) {
 		return req;
 	}
 
-	const delays = isNaN(time) ? time : {socket: time, connect: time};
+	const delays = isNaN(time) ? time : {request: time};
 	const host = req._headers ? (' to ' + req._headers.host) : '';
 
 	function throwESOCKETTIMEDOUT() {
@@ -32,10 +32,10 @@ module.exports = function (req, time) {
 		req.requestTimeoutTimer = setTimeout(() => {
 			clear();
 
-			if (req.connection.connecting) {
-				throwESOCKETTIMEDOUT();
-			} else {
+			if (req.connection.connecting || req.connection._connecting) {
 				throwETIMEDOUT();
+			} else {
+				throwESOCKETTIMEDOUT();
 			}
 		}, delays.request);
 	}
--- a/source/timed-out.js
+++ b/source/timed-out.js
@@ -32,7 +32,7 @@ module.exports = function (req, time) {
 		req.requestTimeoutTimer = setTimeout(() => {
 			clear();
 
-			if (req.connection.connecting || req.connection._connecting) {
+			if (req.connection.connecting) {
 				throwETIMEDOUT();
 			} else {
 				throwESOCKETTIMEDOUT();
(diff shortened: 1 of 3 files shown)


source/timed-out.js
```diff
@@ -32,10 +32,10 @@ module.exports = function (req, time) {
 		req.requestTimeoutTimer = setTimeout(() => {
 			clear();
 
-			if (req.connection.connecting) {
-				throwESOCKETTIMEDOUT();
-			} else {
+			if (req.connection.connecting || req.connection._connecting) {
```
sindresorhus: We don't need to check `._connecting` here. See: https://github.com/sindresorhus/got/blob/bd3315b6c61d20a68944831d8b3a05046d5554ad/index.js#L194

test/timeout.js
```diff
@@ -10,26 +10,44 @@ test.before('setup', async () => {
...
-test('timeout option', async t => {
-	const err = await t.throws(got(`${s.url}/`, {
-		timeout: 1,
+test('timeout option (ETIMEDOUT)', async t => {
+	const err = await t.throws(got(`${s.url}`, {
+		timeout: 0,
+		retries: 0
+	}));
+
+	t.is(err.code, 'ETIMEDOUT');
```
sindresorhus: For new assertions, we should use the improve `t.throws` API: https://github.com/avajs/ava#throwsthrower-expected-message

In this case, it would be:

```js
await t.throws(
	got(s.url, {
		timeout: 0,
		retries: 0
	}),
	{
		code: 'ETIMEDOUT'
	}
);
```

test/timeout.js
```diff
@@ -10,26 +10,44 @@ test.before('setup', async () => {
...
 	});
 
 	await s.listen(s.port);
 });
 
-test('timeout option', async t => {
-	const err = await t.throws(got(`${s.url}/`, {
-		timeout: 1,
+test('timeout option (ETIMEDOUT)', async t => {
+	const err = await t.throws(got(`${s.url}`, {
```
sindresorhus: No point in using a template literal when there's no other content than the variable.

test/timeout.js
```diff
@@ -38,27 +56,29 @@ test('timeout option as object', async t => {
...
 	t.is(err.code, 'ESOCKETTIMEDOUT');
 });
 
-test('connection, request timeout', async t => {
+test.todo('connection timeout');
+
+test('request timeout', async t => {
 	const err = await t.throws(got(`${s.url}`, {
-		timeout: {socket: 50, request: 1},
+		timeout: {request: 40},
```
sindresorhus: Can you make `40` a constant at the top of the file and use that instead? Makes it easier if we ever need to change it.

test/timeout.js
```diff
@@ -5,60 +5,99 @@ import got from '../source';
 import {createServer} from './helpers/server';
 
 let s;
+const ms = 40;
```
sindresorhus: I meant a constant with a descriptive name.

sindresorhus:
👍

PR #510, opened 2018-07-09
Add ability to merge instances

szmarczak:
## WHY?

Imagine there are `got` plugins. You can limit download & upload, use GitHub API, send random User Agent and many more. The possibilities are infinite. What if you want to merge some of them into one single piece? Well, `got.merge()` comes with help.

```js
const singleInstance = got.merge([limitInstance, ghInstance, randomUAInstance, ...]);
```

## DONE
- [x] It's working!
- [x] More tests
- [x] Update `advanced-creation.md`
- [x] Stores only handlers of merged instances instead of whole instances
- [x] Split this PR into smaller pieces: #530 & #532
- [x] Provide some useful use-cases

## TODO
- [ ] Discuss possible alternatives

## DROPPED TASKS
Removing instances from a chain
Checking if a chain has chosen instance
Getting an array of used instances: `got.instances`.

#### Why these tasks were dropped?

That's because defaults are merged, so it's another **single** instance now. The instances you have provided aren't stored.


brandon93s:
What is the use case for this?

> E.g. one instance can limit download & upload and second instance may set some headers. Join them through the `got.forward` function! `instanceC = instanceA.forward(instanceB)`

Does this imply that an app would be using three instances of got? Seems that this could just be handled by constructing / merging numerous config objects together instead of mashing instances together.

test/create.js
```diff
@@ -99,7 +99,7 @@ test('custom endpoint with custom headers (extend)', async t => {
 		json: true
 	})).body;
 	t.is(headers.unicorn, 'rainbow');
-	t.is(headers['user-agent'] === undefined, false);
+	t.false(headers['user-agent'] === undefined);
```
sindresorhus:
```js
t.not(headers['user-agent'], undefined);
```

Same with the other test.

sindresorhus:
I'm also struggling to imagine a use-case for this. What actual (not just example) problem does this solve? I'm not saying it's not a good feature, I just don't see when I would use it.

advanced-creation.md
```diff
@@ -114,3 +112,31 @@ const unicorn = got.create(settings);
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+#### got.forward(instance)
```
sindresorhus: I think `forward` is a confusing name as it sounds like you're forwarding the request. If this were to be added, it would make more sense to just let `got.extend` and `got.create` accept a parent instance.

szmarczak: What if you use two different libs (instances of `got`)? Then you have no possibility to do that. Wait, you could do that...

szmarczak:
@brandon93s You're right, I should've documented that better. 

> Does this imply that an app would be using three instances of got?

No. But it says so, I need to change the docs then.

> Seems that this could just be handled by constructing / merging numerous config objects together instead of mashing instances together.

That's how it's done!

https://github.com/szmarczak/got/blob/074462ae98c8ea27108db17d8d59ee10199bd3c3/source/create.js#L46-L50

szmarczak:
I used to do `limit(gotInstance, bytes)` to limit download & upload or
`process(() => limit(gotInstance, bytes))` when I needed to make request many times + limit that.

Instead of `process(() => limit(gotInstance, bytes))` we can do something like

```js
const instance = process.forward(limit); // I can merge `.forward()` into `.extend()` so it'd look like `process.extend(limit)`

instance(url, { limit: bytes })
```

test/create.js
szmarczak: What other test?

advanced-creation.md
szmarczak: `got.extend(instance)` people would think it'll only merge options (not handlers), the same as `got.extend(instance.defaults.options)` but no the same as `got.forward(instance)`. I'll rename it to `got.merge`.

sindresorhus:
@szmarczak You got a comment on your gist: https://gist.github.com/szmarczak/86217216ad2fbc0c8a3e833e5f290460#gistcomment-2642454 (Mentioning since there are no notifications for gists).

sindresorhus:
@brandon93s Can you do your gist comment here instead? As it's easier to follow the discussion here.

szmarczak:
@brandon93s Yeah. We could do something like this:

```js
const merge = (a, b) => got.create({
	methods: a.defaults.methods,
	options: {...a.defaults.options, ...b.defaults.options},
    handler: (url, options, next) => a.defaults.handler(url, options, (url, options) => b.defaults.handler(url, options, next));
});

const ghLimit = merge(ghGot, limit);
```

I'll just close the PR.

advanced-creation.md
```diff
@@ -11,7 +11,8 @@ Configure a new `got` instance with the provided settings.<br>
 
 ##### [options](readme.md#options)
 
-To inherit from parent, set it as `got.defaults.options` or use [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals).
+To inherit from parent, set it as `got.defaults.options` or use `got.assignOptions(defaults.options, options)`.<br>
+Avoid using [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) as it may give unwanted result.
```
szmarczak: const a = {headers: {cat: 'meow'}};

szmarczak:
@jstewmon Can you review this? I'd be very grateful if you left some notes :smiley:

source/as-stream.js
```diff
@@ -14,6 +14,8 @@ module.exports = options => {
 
 	options.gotRetry.retries = () => 0;
 
+	options.gotRetry.retries = () => 0;
```
brandon93s: Isn't this the same as line 15?

advanced-creation.md
```diff
@@ -101,16 +99,59 @@ const unchangedGot = got.create(defaults);
...
 
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+### Merging instances
+
+#### got.merge(instances, [methods])
+
+Merges many instances into a single one.
```
brandon93s: Would suggest we document the merge strategy. Which options are merged, in what order, etc.

source/as-stream.js
szmarczak: Yes, I forgot to remove it when I was resolving a merge conflict.

advanced-creation.md
szmarczak: Done.

szmarczak:
> It adds complexity (maintenance risk) to the code base for a very advanced used case.

Maybe.

> Since extensive documentation is already needed for it, we could offer tips on how to merge instances in the readme as opposed to offering it in core.

:+1: 

> A separate module could also support this need.

What do you mean by `a separate module`? Another NPM package? There's no need I think.

Another idea:

```js
got.register('functionName', function() {});
got.functionName();
```

sindresorhus:
> Imagine there are got plugins. You can limit download & upload, use GitHub API, send random User Agent and many more.

We should consider what other ways there are to solve these problems.

szmarczak:
> There might be use-cases for this, but I'm a little bit reluctant as this locks us into support the ability to arbitrarily merge instances.

Indeed.

> There might be features we want to add in the future that makes this difficult. 

That's with every single feature, isn't it? :)

> And we have yet to figure out a way to support plugins, which might or might not overlap or conflict with this.
>
> So in short, I just feel like this is rushing into supporting a huge API surface without enough time and exploration to see what problems it solves and the alternatives.

Understood. Better to consider all the alternatives we have here. I'll leave this PR open. This feature is too complicated to release it now.

Tomorrow I'll send two PR with these changes:
- expose `assignOptions` to enable proper options merging (`got.create`)
- better `handler` function: `(normalizedOptions, next)` instead of `(options, next)`
  - move `baseUrl` option from main handler to `normalize-arguments.js`

szmarczak:
This PR will receive a new feature: `defaults.mergeable` - tells whether an instance is mergeable or not. Throws if trying to merge not-mergeable instance.

advanced-creation.md
```diff
@@ -111,3 +111,49 @@ const unicorn = got.create(settings);
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+### Merging instances
+
+#### got.merge(instances, [methods])
```
sindresorhus: `.merge()` is too ambiguous. Let's name it `.mergeInstances()`.

source/create.js
```diff
@@ -9,6 +10,38 @@ const deepFreeze = require('./deep-freeze');
 const next = options => options.stream ? asStream(options) : asPromise(options);
 const mergeOptions = (defaults, options = {}) => merge({}, defaults, options);
 
+const mergeInstances = (instances, methods = instances[0].defaults.methods) => {
```
sindresorhus: Can this be in a separate file?

advanced-creation.md
```diff
@@ -111,3 +111,49 @@ const unicorn = got.create(settings);
...
+##### [[methods]](#methods)
+
+Default: `instances[0].defaults.methods`
+
+```js
+const instanceA = got.extend({headers: {unicorn: 'rainbow'}});
+const instanceB = got.create({
+	options: {baseUrl: 'http://example.com'},
+	methods: [],
+	handler: (options, next) => { // These options are normalized, so assigning `baseUrl` here won't work.
```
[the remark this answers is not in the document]
szmarczak: > Is this enforced?

What do you mean by `enforced`?

> These options are normalized, so assigning baseUrl here won't work.

This should be: `These options are normalized, so assigning baseUrl here won't have any effect.`

source/create.js
```diff
@@ -31,6 +35,24 @@ const create = defaults => {
...
+	got.merge = (...args) => {
+		const lastArgument = args[args.length - 1];
+		let methods;
+		let instances;
+
+		if (is.array(lastArgument)) {
+			methods = lastArgument;
+			instances = args.slice(0, -1);
+		} else {
+			methods = args[0].defaults.methods; // eslint-disable-line prefer-destructuring
```
szmarczak: Is it even possible to do: `let x; {x} = {x: 123};`. Not yet 😕

source/create.js
```diff
@@ -31,6 +35,24 @@ const create = defaults => {
...
+			instances = args;
+		}
+
+		instances = instances.length === 1 ? [got, ...instances] : instances;
```
szmarczak: This hack enables merging two instances by `instanceA.merge(instanceB)`. You can always do `got.merge(instanceA, instanceB)`.

source/merge-instances.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+const merge = require('./merge');
+
+module.exports = (instances, methods) => {
+	for (const [index, instance] of instances.entries()) {
+		if (Reflect.has(instance.defaults, 'mergeable') && !instance.defaults.mergeable) {
+			throw new Error(`Instance ${index} is not mergeable.`);
```
szmarczak: Note: many errors should have their own class.

advanced-creation.md
```diff
@@ -111,3 +111,49 @@ const unicorn = got.create(settings);
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+### Merging instances
+
+#### got.merge(instances, [methods])
+
```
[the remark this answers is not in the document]
szmarczak: Agreed. Mine are:

- controlling redirects,
- limiting download & upload (e.g. you machine has 512MB RAM and you are downloading 1GB file [hue hue hue](https://www.youtube.com/watch?v=okclddpv_9k)),
- using custom API (e.g. `ghGot`),
- sending random `user-agent`.

Also you can use `beforeRequest` hook to sign requests. All these as a single instance.

Do you have any ideas?

advanced-creation.md
sindresorhus: Verbose is better than unclear. It's only a few extra characters and you don't write it that often.

advanced-creation.md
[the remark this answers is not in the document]
sindresorhus: I don't have any ideas, but those sound good. Can you add those as example use-cases to the docs?

advanced-creation.md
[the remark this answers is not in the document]
sindresorhus: By enforced, I mean that it should throw if users accidentally assign `baseUrl` there. Better to fail loud than silently.

source/create.js
sindresorhus: It is, but it's ugly...

```js
({methods} = args[0].defaults);
```

source/create.js
sindresorhus: That what I commented I don't think we should do in https://github.com/sindresorhus/got/pull/510#discussion_r208347443 Better to have one way to do something. Pick one of those, not both.

source/merge-instances.js
sindresorhus: I think the error should say *why* it's not mergeable. That's not clear.

source/create.js
szmarczak: I wonder why it's needed to enclose it in the brackets.

advanced-creation.md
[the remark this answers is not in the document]
szmarczak: Sure!

advanced-creation.md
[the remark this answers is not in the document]
szmarczak: Will do.

source/create.js
szmarczak: Ok.

source/merge-instances.js
szmarczak: It's not mergeable because `instance.defaults.mergeable` is set to `false`. This is the only way to make it unmergeable.

> I think the error should say why it's not mergeable.

`Instance {$index} refused to be merged`?

advanced-creation.md
szmarczak: Okay, I'll change it then.

source/create.js
sindresorhus: From what I can remember, it's something about JS parsing ambiguity.

source/merge-instances.js
sindresorhus: The `mergeable` option is not documented and it's not clear what use-case it serves? Why do you need to make an instance unmergable?

szmarczak: > The mergeable option is not documented (I we decide to keep it, it should be)

Yeah, it should be.

> Why do you need to make an instance unmergable?

I assume instances can be very complicated, e.g. override the current behaviour, or merging it with other instances may somehow result in incompatibility. I see no other reasons. If you feel it's not needed, I'll remove it.

szmarczak: I'll remove it. As for now I don't see any useful use-case. We can add this feature later if needed.

advanced-creation.md
szmarczak: done.

advanced-creation.md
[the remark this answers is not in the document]
szmarczak: done.

szmarczak:
TODO: define throwing setter for options.baseUrl [here](1d0f03b7418c902cb7ec7637a8cd8520R27)

source/normalize-arguments.js
```diff
@@ -51,6 +51,16 @@ module.exports = (url, options, defaults) => {
...
+	const {baseUrl} = options;
+	Object.defineProperty(options, 'baseUrl', {
+		set: _ => {
+			throw new Error('Failed to set baseUrl. Options are normalized already.');
+		},
+		get: () => {
+			return baseUrl;
+		}
+	});
+
```
szmarczak: The same result we can achieve using:

```js
Object.defineProperty(options, 'baseUrl', {
	writable: false
});
```

But the error will just say it's not writable, no other details.

source/normalize-arguments.js
```diff
@@ -51,6 +51,16 @@ module.exports = (url, options, defaults) => {
 		...options
 	};
 
+	const {baseUrl} = options;
+	Object.defineProperty(options, 'baseUrl', {
+		set: _ => {
```
sindresorhus: `_` => `()`

source/normalize-arguments.js
```diff
@@ -51,6 +51,16 @@ module.exports = (url, options, defaults) => {
...
+			throw new Error('Failed to set baseUrl. Options are normalized already.');
+		},
+		get: () => {
```
sindresorhus: `get: () => baseUrl`

source/create.js
```diff
@@ -1,10 +1,14 @@
 'use strict';
+module.exports = () => {};
```
sindresorhus: ?

readme.md
```diff
@@ -795,6 +795,7 @@ const custom = got.extend({
 	const list = await custom('/v1/users/list');
 })();
 ```
+*Need to merge some instances into a single one? Check out the [`got.merge()`](advanced-creation.md#gotmergeinstances-methods).*
```
sindresorhus: The naming needs to be updated here.

advanced-creation.md
```diff
@@ -111,3 +111,112 @@ const unicorn = got.create(settings);
...
+##### [[methods]](#methods)
+
+Default: `instances[0].defaults.methods`
+
+#### Usage
+
+```js
+const got = require('./got');
+
+// #1 You can deny redirects that lead to other sites than specified
```
sindresorhus: When I asked for example use-cases, I really just meant a text list, but this is much better. Can you move the different examples into separate code-blocks with the comment as a header instead.

advanced-creation.md
```diff
@@ -111,3 +111,112 @@ const unicorn = got.create(settings);
...
+
+Merges many instances into a single one:
+- options are merged using [`got.mergeOptions()`](readme.md#gotmergeoptionsparentoptions-newoptions) (+ hooks are merged too),
+- handlers are stored in an array.
+
+##### [instances](readme.md#instances)
+
+##### [[methods]](#methods)
+
+Default: `instances[0].defaults.methods`
```
sindresorhus: What's the use-case for this?

source/create.js
szmarczak: Oh, I was playing with cyclic modules. Removed it.

readme.md
szmarczak: Done.

advanced-creation.md
szmarczak: It's not needed at all (that was an old idea). Hang on, I'll fix it.

advanced-creation.md
szmarczak: Sure! :)

source/normalize-arguments.js
szmarczak: Then node will throw an error that setter is missing an argument, because how you can set something to somewhat without knowing the value?

szmarczak: Weird. Usually it threw an error that setter is missing an argument. I'll change it :)

advanced-creation.md
szmarczak: Removed.

advanced-creation.md
```diff
@@ -111,3 +112,121 @@ const unicorn = got.create(settings);
...
+
+#### got.mergeInstances(instanceA, instanceB, ...)
+
+Merges many instances into a single one:
+- options are merged using [`got.mergeOptions()`](readme.md#gotmergeoptionsparentoptions-newoptions) (+ hooks are merged too),
+- handlers are stored in an array.
+
+##### [instances](readme.md#instances)
+
+## Usage
```
szmarczak: Should the examples be moved to the top?

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+
+### Merging instances
+
+#### got.mergeInstances(instanceA, instanceB, ...)
+
+Merges many instances into a single one:
+- options are merged using [`got.mergeOptions()`](readme.md#gotmergeoptionsparentoptions-newoptions) (+ hooks are merged too),
+- handlers are stored in an array.
+
+## Usage
```
sindresorhus: `Examples`

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+### Merging instances
```
sindresorhus: Can you add an entry to the Highlights section in the readme called `Composable`, which links to this section?

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+
```
sindresorhus: I think it would be nice to have something like this as an intro. Feel free to improve it.

> Got supports composing multiple instances together. This is very powerful. You can create a client that limits download speed and then compose it with an instance that signs a request. It's like plugins without any of the plugin mess. You just create instances and then compose them together.

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+#### got.mergeInstances(instanceA, instanceB, ...)
+
+Merges many instances into a single one:
+- options are merged using [`got.mergeOptions()`](readme.md#gotmergeoptionsparentoptions-newoptions) (+ hooks are merged too),
+- handlers are stored in an array.
+
+## Usage
+
```
sindresorhus: > Some examples of what kind of instances you could compose together.

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+			const host = new URL(resp.url).host;
+			if (options.allowedHosts && !options.allowedHosts.includes(host)) {
+				promiseOrStream.cancel(`Redirection to ${host} is not allowed`);
+			}
+		});
+	}
+});
+```
+
+#### Limiting download & upload (in case your machine's got a little amount of RAM)
```
sindresorhus: `(in case your machine's got a little amount of RAM)` should be in text below the title.

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+				options.headers['sign'] = getMessageSignature(options.body || '', process.env.SECRET);
+			}
+		]
+	}
+});
+```
+
+If these ^^^ are different modules and you don't want to rewrite them, use `got.mergeInstances()`.
+
+**Note**: `noUserAgent` must be placed at the end of chain as the instances are merged serially. Other modules do have the `user-agent` header.
```
sindresorhus: ``noUserAgent`` => The `noUserAgent` instance

sindresorhus: `are merged serially` => `are merged in order`

sindresorhus: `Other modules` => `Other instances`

advanced-creation.md
```diff
@@ -111,3 +112,119 @@ const unicorn = got.create(settings);
...
+			options => {
+				options.headers['sign'] = getMessageSignature(options.body || '', process.env.SECRET);
+			}
+		]
+	}
+});
+```
+
+If these ^^^ are different modules and you don't want to rewrite them, use `got.mergeInstances()`.
+
```
sindresorhus: There should be a header here called `#### Putting it all together` or something.

test/create.js
```diff
@@ -102,7 +102,6 @@ test('extend merges URL instances', t => {
 test('create', async t => {
 	const instance = got.create({
 		options: {},
-		methods: ['get'],
```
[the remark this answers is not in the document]
szmarczak: These are just aliases. If `methods` are set to `['get']` you can't do `got.post()` but you can do `got(url, {method: 'POST'});`. I see no reason why we should let users to modify the aliases.

advanced-creation.md
szmarczak: Sounds good!

test/create.js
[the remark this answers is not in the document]
sindresorhus: I agree 👍

sindresorhus:
This is looking good now :)

PR #513, opened 2018-07-11
Add tests for `url-to-options.js`

alextes:
One moment you're looking at how beautiful got has grown. The next you're submitting a PR to increase test coverage.

Not sure we want to take on these tests. The code it tests is copied from Node and more code means more maintenance. That said, I feel if you copy it, it's yours. Good to make sure it keeps working going forward I think.

If you disagree, feel free to close it. This was more for me than the users, though I do hope our users can get something out of it too 😁 .

### DIFF
--- a/test/url-to-options.js
+++ b/test/url-to-options.js
@@ -1,0 +1,76 @@
+import url from 'url';
+import test from 'ava';
+import urlToOptions from '../source/url-to-options';
+
+const exampleURL = 'https://user:password@github.com:443/say?hello=world#bang';
+
+test('converts node legacy URL to options', t => {
+	const parsedURL = url.parse(exampleURL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: '#bang',
+		hostname: 'github.com',
+		href: exampleURL,
+		path: '/say?hello=world',
+		pathname: '/say',
+		port: 443,
+		protocol: 'https:',
+		search: '?hello=world'
+	};
+
+	t.deepEqual(options, expected);
+});
+
+test('converts URL to options', t => {
+	// TODO: Use the `URL` global when targeting Node.js 10
+	const parsedURL = new url.URL(exampleURL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		auth: 'user:password',
+		hash: '#bang',
+		hostname: 'github.com',
+		href: 'https://user:password@github.com/say?hello=world#bang',
+		path: '/say?hello=world',
+		pathname: '/say',
+		protocol: 'https:',
+		search: '?hello=world'
+	};
+
+	t.deepEqual(options, expected);
+});
+
+test('converts IPv6 URL to options', t => {
+	const IPv6URL = 'https://[2001:cdba::3257:9652]:443/';
+	// TODO: Use the `URL` global when targeting Node.js 10
+	const parsedURL = new url.URL(IPv6URL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: '',
+		hostname: '2001:cdba::3257:9652',
+		href: 'https://[2001:cdba::3257:9652]/',
+		path: '/',
+		pathname: '/',
+		protocol: 'https:',
+		search: ''
+	};
+
+	t.deepEqual(options, expected);
+});
+
+test('only adds port to options for URLs with ports', t => {
+	const noPortURL = 'https://github.com/';
+	const parsedURL = new url.URL(noPortURL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: '',
+		hostname: 'github.com',
+		href: 'https://github.com/',
+		path: '/',
+		pathname: '/',
+		protocol: 'https:',
+		search: ''
+	};
+
+	t.deepEqual(options, expected);
+	t.false(Reflect.has(options, 'port'));
+});


alextes:
From
```
-----------------------------|----------|----------|----------|----------|-------------------|
File                         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------------------|----------|----------|----------|----------|-------------------|
All files                    |    97.71 |     93.4 |    98.31 |    97.69 |                   |
 url-to-options.js           |    85.71 |     62.5 |      100 |    85.71 |                20 |
-----------------------------|----------|----------|----------|----------|-------------------|

```
To
```
-----------------------------|----------|----------|----------|----------|-------------------|
File                         |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-----------------------------|----------|----------|----------|----------|-------------------|
All files                    |    97.71 |     93.4 |    98.31 |    97.69 |                   |
 url-to-options.js           |      100 |      100 |      100 |      100 |                   |
-----------------------------|----------|----------|----------|----------|-------------------|
```
😁

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+
+const githubURL = 'https://alex:pass@github.com:443/';
+
+test('Converts node legacy URL to options', t => {
+	const parsedURL = url.parse(githubURL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: null,
+		hostname: 'github.com',
+		href: 'https://alex:pass@github.com:443/',
```
sindresorhus: `href: githubURL`,

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
+import url from 'url';
+import test from 'ava';
+import urlToOptions from '../source/url-to-options';
+
+const githubURL = 'https://alex:pass@github.com:443/';
```
sindresorhus: `const githubURL = 'https://user:password@github.com:443/';`

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+
+test('Converts node legacy URL to options', t => {
```
sindresorhus: `Converts` => `converts`

test/url-to-options.js
alextes: Why didn't you suggest this approach for the second test 🤔 ? Did you know it would break?

test/url-to-options.js
alextes: Sorry for letting my ego creep into the code 🙈 .

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+		port: 443,
+		protocol: 'https:',
+		search: null
+	};
+
+	t.deepEqual(options, expected);
+});
+
+test('converts URL to options', t => {
+	const parsedURL = new url.URL(githubURL);
```
szmarczak: Keep in mind that since Node.JS 10 URL is a global :)
I think we should duplicate [that TODO comment](https://github.com/sindresorhus/got/blob/master/source/normalize-arguments.js#L2) and paste it there:
```// TODO: Use the `URL` global when targeting Node.js 10```

File: test/url-to-options.js [hunk 5]
```diff
@@ -0,0 +1,56 @@
+import url from 'url';
+import test from 'ava';
+import urlToOptions from '../source/url-to-options';
+
+const githubURL = 'https://user:password@github.com:443/';
```
szmarczak: It should be more complicated. Like:

```js
const githubURL = 'https://user:password@github.com:443/say?hello=world#awesome';
```

Also I would rename `githubURL` to `exampleURL`.

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: '',
+		hostname: '2001:cdba::3257:9652',
+		href: 'https://[2001:cdba::3257:9652]/',
+		path: '/',
+		pathname: '/',
+		protocol: 'https:',
+		search: ''
+	};
```
szmarczak: Shouldn't be there `expected.port = 443`?

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+
+	t.deepEqual(options, expected);
+});
```
szmarczak: Needs another test with no port:
```js
const noPortURL = 'https://github.com/';
const parsedURL = new url.URL(noPortURL);
const options = urlToOptions(parsedURL);
const expected = {
    hash: '',
    hostname: 'github.com',
    href: 'https://github.com/',
    path: '/',
    pathname: '/',
    protocol: 'https:',
    search: ''
};
```

test/url-to-options.js
alextes: To be honest, I don't feel these are useful at all. I'd simply say that if we end up preferring global over imported `URL` and we're at a point where our lowest target is Node 10, anybody can feel free to change it.

Until then, even assuming it is a change we don't want to forget to make, I don't see the point in having the comment.

Regardless, I don't like being too opinionated on a project I've been away from especially on a trivial two line change 😊 , so fixed!

test/url-to-options.js
```diff
@@ -0,0 +1,56 @@
...
+const githubURL = 'https://user:password@github.com:443/';
+
+test('converts node legacy URL to options', t => {
+	const parsedURL = url.parse(githubURL);
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: null,
+		hostname: 'github.com',
+		href: githubURL,
+		path: '/null',
```
alextes: This doesn't seem right 🤔 .

test/url-to-options.js
alextes: What can I tell you 😅 ? Computer says no.

szmarczak: You could change the port to another, e.g. to `8080` or `80`. But I don't know if it's a good idea.

I would just add a comment: `// IPv6 + :443 = no port`

test/url-to-options.js
szmarczak: Smart!

`url.search` is null in this case. It should add that to `path`. Instead:

https://github.com/alextes/got/blob/dccde51438af10a8a5236288c52b643dd95eb969/source/url-to-options.js#L11

do:
```js
	const options = {
		protocol: url.protocol,
		hostname: url.hostname.startsWith('[') ? url.hostname.slice(1, -1) : url.hostname,
		hash: url.hash,
		search: url.search,
		pathname: url.pathname,
		path: `${url.pathname}`,
		href: url.href
	};
	if (url.search !== null) {
		options.path += url.search;
	}
```

test/url-to-options.js
alextes: Out of curiosity, why Reflect? Why not `hasOwnProperty`?

szmarczak: Because `AVA`. (throws an error IIRC)

test/url-to-options.js
szmarczak: The comment shouldn't be on a separate line I think :) Maybe (?):

```js
	const parsedURL = new url.URL(IPv6URL); // TODO: Use the `URL` global when targeting Node.js 10
```

alextes: Just to be a little more clear 😅 , if Node feels that the default HTTPS port should not be added if the protocol is already HTTPS, I guess that's fine. I wouldn't write code to go against what Node does.

szmarczak:
I would change:

```js
	const options = {
		protocol: url.protocol,
		hostname: url.hostname.startsWith('[') ? url.hostname.slice(1, -1) : url.hostname,
		hash: url.hash,
		search: url.search,
		pathname: url.pathname,
		path: `${url.pathname}${url.search}`,
		href: url.href
	};
```

to:
```js
	const options = {
		protocol: url.protocol,
		hostname: url.hostname.startsWith('[') ? url.hostname.slice(1, -1) : url.hostname,
		hash: url.hash,
		search: url.search,
		pathname: url.pathname,
		path: `${url.pathname}`,
		href: url.href
	};
	if (url.search !== null) {
		options.path += url.search;
	}
```

test/url-to-options.js
szmarczak: That's https? Then sorry, hadn't noticed that :D

alextes:
@szmarczak yea, already working on it, the odd thing is, I started with writing a failing test and can't quite seem to reproduce 🤔 .

One sec.

szmarczak:
Opened an issue at node repo: https://github.com/nodejs/node/issues/21762

test/url-to-options.js
sindresorhus: > Out of curiosity, why Reflect? Why not hasOwnProperty?

`.hasOwnProperty` won't work with `Object.create(null)` created objects.

> Because AVA. (throws an error IIRC)

You mean `XO` ;)

PR #515, opened 2018-07-11
Fix null search being appended to path when legacy parsing URL

alextes:
This still needs to be rebased after #513 is merged.

### DIFF
--- a/test/url-to-options.js
+++ b/test/url-to-options.js
@@ -1,0 +1,76 @@
+import url from 'url';
+import test from 'ava';
+import urlToOptions from '../source/url-to-options';
+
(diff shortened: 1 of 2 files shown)


test/url-to-options.js
```diff
@@ -0,0 +1,98 @@
...
+	t.is(parsedURL.search, null);
+
+	const options = urlToOptions(parsedURL);
+	const expected = {
+		hash: null,
+		hostname: 'github.com',
+		href: 'https://github.com/',
+		path: '/',
+		pathname: '/',
+		port: 0,
```
alextes: What 😅 .

alextes:
Ah and same problem with `hash` of course.

sindresorhus:
Good stuff, @alextes 👌

PR #517, opened 2018-07-11
Various minor code tweaks

jstewmon:
These three commits improve a few small nits I discovered while working on #516


jstewmon:
I noticed that I used inconsistent formatting on the EE proxy commit, so I pushed an update to make them look the same.

PR #518, opened 2018-07-11
Set headers on stream proxy

### LINKED ISSUE #401: Proxy headers in got.stream
I use `got@6.7.1`. I need to proxy an url from some upstream. So, I do
```
got.stream(url).pipe(res);
```
where `res` is an express response.

But it proxies only body, not headers. To proxy headers as well I have to add an extra handler
```
got.stream(url)
    .on('response', (msg) => {
        res.set(msg.headers);
    })
    .pipe(res);
```

Is it expected behaviour? For example `request` library proxies body and headers out of the box.

szmarczak:
Closes #401.

### DIFF
--- a/source/as-stream.js
+++ b/source/as-stream.js
@@ -62,7 +62,7 @@ module.exports = options => {
 		response.pipe(output);
 
 		for (const destination of piped) {
-			if (!destination.headersSent) {
+			if (destination.headersSent) {
 				continue;
 			}
 
--- a/source/as-stream.js
+++ b/source/as-stream.js
@@ -10,7 +10,7 @@ module.exports = options => {
 	const output = new PassThrough();
 	const proxy = duplexer3(input, output);
 	const piped = new Set();
-	let finished = false;
+	let isFinished = false;
 
 	if (options.json) {
 		throw new Error('Got can not be used as a stream when the `json` option is used');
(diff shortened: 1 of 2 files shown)


source/as-stream.js
```diff
@@ -50,13 +52,27 @@ module.exports = options => {
...
 			proxy.emit('error', new HTTPError(statusCode, response.statusMessage, response.headers, options), null, response);
 			return;
 		}
 
+		finished = true;
+
+		response.pipe(output);
+
+		for (const destination of piped) {
+			if (!destination.headersSent) {
```
sindresorhus: Usually better with early returns to reduce nesting.

```js
if (!destination.headersSent) {
  continue;
}
```

source/as-stream.js
```diff
@@ -50,13 +52,27 @@ module.exports = options => {
...
+				for (const [key, value] of Object.entries(response.headers)) {
+					if (key.toLowerCase() !== 'content-encoding') {
```
sindresorhus: Can you add a short code comment why we don't set this? Are there other headers we shouldn't overwrite?

source/as-stream.js
```diff
@@ -9,6 +9,8 @@ module.exports = options => {
 	const input = new PassThrough();
 	const output = new PassThrough();
 	const proxy = duplexer3(input, output);
+	const piped = new Set();
+	let finished = false;
```
sindresorhus: `isFinished`

source/as-stream.js
szmarczak: > Can you add a short code comment why we don't set this?

Yes.

> Are there other headers we shouldn't overwrite?

No.

source/as-stream.js
szmarczak: You misled me :P You shouldn't negate the condition :D

Issue #487, opened 2018-05-24
Cache doesn't work properly when using `query` option


bdougherty:
Noticed that two requests that share the same base url but have a different query string were coming from the cache of the first request when using the `query` option. I was using a `Map` for the cache and the only key in there did not include the query string, so not sure if this is an issue here or with [cacheable-request](https://github.com/lukechilds/cacheable-request).

Notice there is only one key and the assertion fails in this example:

```js
const got = require('got');

(async () => {
	const cache = new Map();

	const { body: body1 } = await got('https://nominatim.openstreetmap.org/search', {
		cache,
		query: {
			format: 'jsonv2',
			addressdetails: 1,
			q: 'philadelphia, pa'
		}
	});

	const { body: body2 } = await got('https://nominatim.openstreetmap.org/search', {
		cache,
		query: {
			format: 'jsonv2',
			addressdetails: 1,
			q: 'atlantic city, nj'
		}
	});

	console.assert(body1 !== body2);
	// Assertion failed
	console.log([...cache.keys()]);
	// [ 'cacheable-request:GET:https://nominatim.openstreetmap.org/search' ]
})();
```

But everything works fine when using a `URL`:

```js
const { URL } = require('url');
const got = require('got');

(async () => {
	const cache = new Map();

	const url1 = new URL('https://nominatim.openstreetmap.org/search');
	url1.searchParams.set('format', 'jsonv2');
	url1.searchParams.set('addressdetails', 1);
	url1.searchParams.set('q', 'philadelphia, pa');

	const { body: body1 } = await got(url1, {
		cache
	});

	const url2 = new URL('https://nominatim.openstreetmap.org/search');
	url2.searchParams.set('format', 'jsonv2');
	url2.searchParams.set('addressdetails', 1);
	url2.searchParams.set('q', 'atlantic city, nj');

	const { body: body2 } = await got(url2, {
		cache
	});

	console.assert(body1 !== body2);
	console.log([...cache.keys()]);
	// [ 'cacheable-request:GET:https://nominatim.openstreetmap.org/search?addressdetails=1&format=jsonv2&q=philadelphia, pa',
	// 'cacheable-request:GET:https://nominatim.openstreetmap.org/search?addressdetails=1&format=jsonv2&q=atlantic city, nj' ]
})();
```

lukechilds:
Thanks for reporting with details. Definitely looks like an issue with `cacheable-request`. Would you mind reporting this over there?

brandon93s:
[...]

PR #519, opened 2018-07-13
Fix: formatted query should be assigned to `options.search`

jstewmon:
cacheable-request uses url.format to construct the cache key.
url.format uses `pathname` and `search` when formatting the path.

fixes: #487

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
 	],
 	"dependencies": {
 		"@sindresorhus/is": "^0.9.0",
-		"cacheable-request": "^2.1.1",
+		"cacheable-request": "^4.0.1",
 		"decompress-response": "^3.3.0",
 		"duplexer3": "^0.1.4",
 		"extend": "^3.0.1",
--- a/test/arguments.js
+++ b/test/arguments.js
@@ -60,7 +60,21 @@ test('requestUrl with url.parse object as first argument', async t => {
 });
 
 test('overrides querystring from opts', async t => {
-	t.is((await got(`${s.url}/?test=doge`, {query: {test: 'wow'}})).body, '/?test=wow');
+	const response = await got(
+		`${s.url}/?drop=this`,
+		{
+			query: {test: 'wow'},
+			cache: {
+				get(key) {
+					t.is(key, `cacheable-request:GET:${s.url}/?test=wow`);
+				},
+				set(key) {
+					t.is(key, `cacheable-request:GET:${s.url}/?test=wow`);
+				}
+			}
+		}
+	);
+	t.is(response.body, '/?test=wow');
 });
 
 test('should throw with auth in url string', async t => {


sindresorhus:
Wouldn't it be better to fix it at the source: https://github.com/lukechilds/cacheable-request/issues/42 ?

jstewmon:
This is a complicated issue due to the inconsistencies of the node apis for urls and http requests...

The [http.request] options argument aligns more closely with the legacy [urlObject] structure when the argument is an options object (not an instance of URL) in that it expects `auth`, not `username` and `password`. http.request requires `path` to be `${pathname}${search}`.

I originally thought it would be better to use WHATWG URL because it only has `pathname` and `search`, which would mean that got would know to set `search`, not `path` in normalize-arguments, but what about the additional request options, which are not members of URL?

I'm not sure yet what `http.request` will do if the argument is an instance of URL with additional option properties that are not part of the URL interface (`localAddress`, etc).

I skimmed the cacheable-request docs and it doesn't actually document the contract it uses to compute the cache key, so that also needs clarification.

In summary, both url apis use `pathname` and `search` to format the href string while `http.request` expects a `path` option with the two combined.

cacheable-request stipulates the contract for determining the cache key, so my goal here was to ensure that got's normalized arguments conform to both the `http.request` api it tells cacheable-request to use when sending the request and to the api used to format the cache key.

@sindresorhus if you have a suggestion on how to create better alignment between the url apis, http.request, cacheable-request and got, I can try to work something out, but it appears arduous to me. I'd suggest taking this change to fix the current bug if a more elaborate plan is conceived.

[http.request]: https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_http_request
[urlObject]: https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_legacy_urlobject

jstewmon:
It looks like node will drop any additional properties if options is an instance of URL.

The [ClientRequest] constructor will reassign the `options` variable to the result of [urlToOptions], which plucks only the known URL class properties.

[ClientRequest]: https://github.com/nodejs/node/blob/v8.11.3/lib/_http_client.js#L90
[urlToOptions]: https://github.com/nodejs/node/blob/v8.11.3/lib/internal/url.js#L1298

jstewmon:
Here's an idea - if cacheable-request wants to maintain the same contract as `http.request`, then we can fix the issue there by formatting the key according to the options understood by `http.request`. I'll update the linked issue with that suggestion.

source/normalize-arguments.js
```diff
@@ -49,8 +49,8 @@ module.exports = (url, options, defaults) => {
 		if (!is.string(query)) {
 			options.query = (new URLSearchParamsGlobal(query)).toString();
 		}
-
-		options.path = `${options.path.split('?')[0]}?${options.query}`;
+		options.search = `?${options.query}`;
+		options.path = `${options.path.split('?')[0]}${options.search}`;
```
szmarczak: If `options.search` gets changed then you need to change `options.path` too.

sindresorhus:
> if cacheable-request wants to maintain the same contract as http.request

Yes, the point of it is to be a drop-in replacement of `http.request`, so it should maintain the same contract.

> then we can fix the issue there by formatting the key according to the options understood by http.request.

👍

source/normalize-arguments.js
szmarczak: I mean:

szmarczak: You can change `options.search` using `beforeRequest` too.

jstewmon:
I updated to `cacheable-request@4`, but the new version is causing something to hang. I don't have time to look into it right now.

cc @lukechilds

lukechilds:
@jstewmon what OS are you on?

Do `cacheable-request` tests pass for you?

jstewmon:
Yeah,  I suspect that the got test is hanging because the path being used in the request does not match a route the server is listening for. I think we just need to verify that the cacheable-request behavior is correct and that the test set up is incorrect .

jstewmon:
I found the problem. There's a bug in https://github.com/lukechilds/cacheable-request/pull/43. The `search` property doesn't have the leading `?`.

I'll file an issue and submit a patch.

szmarczak:
@jstewmon Is this issue fixed in https://github.com/lukechilds/cacheable-request/pull/44?

jstewmon:
Indeed, it is. I updated this PR to cacheable-request@4.0.1 and all tests are now passing.

PR #520, opened 2018-07-13
Standardize on camelcase

jstewmon:
addresses discussion from https://github.com/sindresorhus/got/pull/516#discussion_r202440819 about standardizing on camelCase to avoid having to decide when to use UPPER_CASE for const vars


sindresorhus:
I've opened an issue on ESLint so we can enforce this in the future: https://github.com/eslint/eslint/issues/10605

jstewmon:
Thanks for opening that issue - I reviewed the docs today when I made this change and was disappointed that they didn't have an option to disallow `_` when the alpha chars are uppercase.

PR #521, opened 2018-07-14
Loosen unnecessary restrictions on the `body` option

### LINKED ISSUE #381: Regression in 7.0.0 options processing
#217 loosened unnecessary restrictions on `body` option that were discussed in #212 and allowed it to be an object that is generally considered *plain* by a developer, i.e. not a *complex* object - a buffer or a stream, which are processed differently. 

#297 introduced a regression that imposed a restriction on `body` option again.

Excessive sanitization results from the use of `is-plain-obj` and restricts `body` object prototype chain to contain only  `Object.prototype` or `null`. This restriction results in counter-intuitive behaviour and requires a developer to use tricks like `Object.assign({}, obj)` instead of not doing them, even if an object is perceived as 'plain' (in my experience it was `querystring.parse()`) .

More importantly, this results in breaking changes that aren't reflected in change logs. The restrictions were dropped for a good reason, and resulting behaviour has been already relied on by package users. Now the regression just happened without an obvious reason.

szmarczak:
Fixes #381

### DIFF
--- a/test/error.js
+++ b/test/error.js
@@ -26,6 +26,15 @@ test.before('setup', async () => {
 		res.end('body');
 	});
 
+	s.on('/body', (req, res) => {
+		const body = [];
+		req.on('data', chunk => {
+			body.push(chunk);
+		}).on('end', () => {
+			res.end(Buffer.concat(body).toString());
+		});
+	});
+
 	await s.listen(s.port);
 });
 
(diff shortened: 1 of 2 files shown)


test/error.js
```diff
@@ -26,6 +26,15 @@ test.before('setup', async () => {
 		res.end('body');
 	});
 
+	s.on('/body', (req, res) => {
+		const body = [];
+		req.on('data', chunk => {
+			body.push(chunk);
+		}).on('end', () => {
+			res.end(Buffer.concat(body).toString());
+		});
```
sindresorhus: Use `get-stream` here for simplicity

test/error.js
```diff
@@ -54,17 +63,27 @@ test('dns message', async t => {
 
 test('options.body error message', async t => {
 	const err = await t.throws(got(s.url, {body: {}}));
-	t.regex(err.message, /The `body` option must be a stream\.Readable, string or Buffer/);
+	t.is(err.message, 'The `body` option must be a stream.Readable, string or Buffer');
```
sindresorhus: We should change all of the `t.throws` assertions to the new format. Should not be done in this PR though. Just a note to myself.

test/error.js
```diff
@@ -54,17 +63,27 @@ test('dns message', async t => {
...
-	t.regex(err.message, /The `body` option must be a plain Object when the `form` option is used/);
+	t.is(err.message, 'The `body` option must be an Object when the `form` option is used');
+});
+
+test('no plain object restriction on body', async t => {
+	function CustomObject() {
+		this.a = 123;
+	}
+
+	const {body} = (await got(`${s.url}/body`, {body: new CustomObject(), json: true}));
```
sindresorhus: You don't need parens around `await`

test/error.js
szmarczak: Ok, so after this gets merged I'll do another PR for that.

test/error.js
szmarczak: Yup. Forgot because before I did `const body = (...).body` then I used `const { body } = ...` :)

PR #523, opened 2018-07-15
Make hooks better 🦄

szmarczak:
- [x] `got.hooks = got.defaults.options.hooks` (all defaults are deep frozen except `hooks`)
- [x] By default hooks are always empty<br>(before: if you hadn't specified `hooks` in the defaults, `got.hooks` would be `undefined`)
- [x] Improved hook type checking
- [x] New hooks: `onSocketConnect`, `onAbort`


sindresorhus:
// @jstewmon

source/normalize-arguments.js
```diff
@@ -187,24 +186,14 @@ module.exports = (url, options, defaults) => {
...
-				options.hooks[hookEvent] = [];
-			} else if (is.array(hooks)) {
-				hooks.forEach(
-					(hook, index) => {
-						if (!is.function_(hook)) {
-							throw new TypeError(
-								`Parameter \`hooks.${hookEvent}[${index}]\` must be a function, not ${is(hook)}`
-							);
-						}
+		for (const [hookEvent, hooks] of Object.entries(options.hooks)) {
```
jstewmon: This is not equivalent to the current implementation. Consider these examples:

```js
const a = [1, 2];
a.foo = 'bar';
Object.entries(a); // [ [ '0', 1 ], [ '1', 2 ], [ 'foo', 1 ] ]
```

source/normalize-arguments.js
```diff
@@ -187,24 +186,14 @@ module.exports = (url, options, defaults) => {
...
+			if (is.array(hooks)) {
```
jstewmon: This fails to validate the known hooks if something other than an array is provided, which will result in an error downstream.

source/request-as-event-emitter.js
```diff
@@ -150,7 +157,9 @@ module.exports = (options = {}) => {
 
 				const socket = req.connection;
 				if (socket) {
-					const onSocketConnect = () => {
+					const onSocketConnect = async () => {
```
jstewmon: I'm not sure what the origin of this conditional block are, but I think the right way to handle this event is to remove the if block and make this:

```js
req.on('socket', () => { ... });
```

https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_event_socket

jstewmon:
I really like being able to configure event listeners on a per-instance basis - it's really useful for being able to add instrumentation to a client without having to litter every request site with the event listeners.

I have a few questions regarding the implementation:

- Should these go under `hooks`? These are actually EE listeners, so it might be better to give them a discrete top-level config key like `listeners`. This would also disambiguate them from hooks, which are stated to allow modification during the pipeline.
- The EE listeners can't modify anything, so they should not be awaited even if they are async.
- I don't think it makes sense to have some aspects of instance configuration be frozen while others are not. Pick one - frozen or unfrozen. I don't see a practical reason why the configuration is frozen... Freezing makes perfect sense when there is state synchronization that happens during instance setup, which would cause modifications to the configuration to be ignored by the instance, but I don't think that's the case with got, so I'd say got should not freeze configuration. If a user wants to ensure the config is not modified, they can freeze it themselves when they create the instance.

source/request-as-event-emitter.js
jstewmon: Actually, since `onSocketConnect` and `onAbort` should be attached to the request, I don't think these are sensible hooks for `got` to configure directly. If you want to configure these for a got instance, you really just need to configure a listener for `request` and attach the request event listeners in the request listener.

test/arguments.js
```diff
@@ -132,7 +132,7 @@ test('throws TypeError when known `hooks` array item is not a function', async t
 });
 
 test('allows extra keys in `hooks`', async t => {
-	await t.notThrows(() => got(`${s.url}/test`, {hooks: {extra: {}}}));
+	await t.notThrows(() => got(`${s.url}/test`, {hooks: {extra: []}}));
```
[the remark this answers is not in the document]
szmarczak: Why would someone assign non-hook object to hooks? What's the use case? What big problem does this solve? Sorry, but I don't see that.

source/request-as-event-emitter.js
szmarczak: > I'm not sure what the origin of this conditional block are, but I think the right way to handle this event is to remove the if block

#429 

> Actually, since onSocketConnect and onAbort should be attached to the request, I don't think these are sensible hooks for got to configure directly. If you want to configure these for a got instance, you really just need to configure a listener for request and attach the request event listeners in the request listener.

Why would you do the same thing twice?

wise

source/normalize-arguments.js
szmarczak: Nope. https://github.com/szmarczak/got/blob/b22c3abbe8daf905aae4bed66cddfc7b9ed72170/source/normalize-arguments.js#L191-L199

source/normalize-arguments.js
```diff
@@ -187,24 +186,14 @@ module.exports = (url, options, defaults) => {
...
+				for (const [index, hook] of Object.entries(hooks)) {
```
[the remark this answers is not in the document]
szmarczak: Thanks for the tip! :)

test/arguments.js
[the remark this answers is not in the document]
jstewmon: It allows for encapsulation. Here's a contrived example:

```js
class GotHooks {

  constructor () {
    this.message = 'Encapsulated Got Hooks';
    this.beforeRequest = [
      () => this.runRook()
    ];
  }

  async runRook() {
    console.log(this.message);
  }
}
```

If got validates only the properties it is concerned with, then everything is fine. If got extraneously demands that the object only have properties that correspond to known hooks, then the user has to jump through some hoops to create an object that accomplishes the same thing.

szmarczak:
> Should these go under hooks? These are actually EE listeners, so it might be better to give them a discrete top-level config key like listeners. This would also disambiguate them from hooks, which are stated to allow modification during the pipeline.

You're right. Instead of `onAbort` it should be `onCancel`. 

> The EE listeners can't modify anything, so they should not be awaited even if they are async.

I don't understand. EE listeners aren't awaited.

> I don't think it makes sense to have some aspects of instance configuration be frozen while others are not. Pick one - frozen or unfrozen. I don't see a practical reason why the configuration is frozen... 

It's frozen to prevent unwanted changes to the instance. But yeah, some people might want to keep them frozen and some people might want to not. My proposal: to disable freezing people will need to set `defaults.freezeOptions` to `false` (`true` by default).

source/request-as-event-emitter.js
jstewmon: Thanks for the reference to #429. Maybe I'm missing something, but I don't think that was the right fix b/c `if req.connection` is just testing a race condition. Shouldn't it be:

```js
req.once('socket', (sock) => {
  sock.once('connect', conn => {
    // ...
  }
};
```

I can see the convenience of low-friction hooks. Maybe there's a way to make the configuration a sort of DSL that got can use to wire these up. Something like this:

```js
{
  events: {
    request: {
      ['on|once']: {
        socket: {
          null: [(sock) => { console.log('request.on(socket)') }],
          connect: [(conn) => { console.log('socket.on(connect)') }]
        }
      },
    }
  }
}
```

With the above, the configuration can be traversed and the listeners can be subscribed without having to add handling code to got for every event.

I used `null` as a sentinel value to mean the subscriber for the event as opposed to a child subscription. Something else (symbol?) would work also.

source/normalize-arguments.js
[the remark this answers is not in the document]
szmarczak: *It's not a bug. It's a feature!*

source/normalize-arguments.js
jstewmon: Ah, thanks - I got tripped up reviewing the diff.

jstewmon:
> I don't understand. EE listeners aren't awaited.

You're using `callAll`, which awaits the hook function, in response to an EE event.

jstewmon:
Can we agree that the options should either be frozen or not? If so, then I think you can cut down this PR by reverting the changes to how the configuration is setup and normalized, right? That will allow us to focus on the interface for the event listeners.

szmarczak:
I'm closing this due to a [GitHub crash](https://twitter.com/githubstatus/status/1018911359881416705). This PR doesn't see new commits. I'll create another PR when I'm done with fixing things. Big thanks to @jstewmon for letting me know what I've done wrong! :raised_hands:

source/request-as-event-emitter.js
szmarczak: > Shouldn't it be:

szmarczak: > I can see the convenience of low-friction hooks.

Great! We'll discuss this later. I was gonna push a PR with rewrited hooks from scratch but I decided current implementation is better. :)

szmarczak:
> You're using callAll, which awaits the hook function, in response to an EE event.

It's fixed in the upcoming PR.

szmarczak:
> Can we agree that the options should either be frozen or not?

My answer: `defaults.preventChanges`. The default value still needs to be discussed.

PR #524, opened 2018-07-16
Better hooks and optional freezable defaults

szmarczak:
Migrated from #523 due to a GitHub crash.

- [x] `got.hooks = got.defaults.options.hooks`
- [x] By default hooks are always empty<br>(before: if you hadn't specified `hooks` in the defaults, `got.hooks` would be `undefined`)
- [x] `defaults.preventChanges` (documented in `advanced-creation.md`)


szmarczak:
@jstewmon ping

test/create.js
```diff
@@ -118,11 +118,25 @@ test('no tampering with defaults', t => {
...
 	});
 
-	t.is(instance.defaults.options.baseUrl, 'example');
+	t.is(instance.defaults.options.baseUrl, 'http://google.com');
 	t.is(instance2.defaults.options.baseUrl, 'example');
 });
+
+test('can prevent unwanted changes', t => {
+	const instance = got.create({
+		preventChanges: true,
```
sindresorhus: It's not clear what this does exactly? What changes does it prevent? All? I think it needs a more explicit name.

source/define-const-property.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+const is = require('@sindresorhus/is');
+
+module.exports = (where, properties, deep = false, excluded = []) => {
```
sindresorhus: This method needs some internal docs in form of a code comment above. What it does exactly.

For readability, I think `deep = false, excluded = []` should be converted into an options object. `deep` is a boolean trap: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

source/create.js
```diff
@@ -51,18 +53,18 @@ const create = defaults => {
 		return defaults.handler(url, options, next);
 	};
 
+	got.hooks = defaults.options.hooks = { // eslint-disable-line no-multi-assign
```
sindresorhus: The ESLint rule is there for a reason ;)

szmarczak: I don't understand what's wrong with that :P I'll change it :)

sindresorhus: In this case, nothing major, but very often it creates harder to read code.

source/define-const-property.js
szmarczak: Thanks for telling me about that, very interesting :3

jstewmon:
@szmarczak can you make separate pull requests out of these changes? I think there are three things happening here that aren't strictly related:
- new option to not freeze options
- hooks option default assignment is moved into got.create
- whitespace reformatting

There are some changes leftover from the other PR that appear to be extraneous - `defineConstPoperty` and changes to hooks normalization.

szmarczak:
#### 1

> new option to not freeze options

That's part of the improved hooks. The current hooks you provided are deep frozen. It shouldn't. The ability to not freeze options is by-the-way.

It's strictly related here.

#### 2

> hooks option default assignment is moved into got.create

That's part of the improved hooks.
It's strictly related here.

#### 3

> whitespace reformatting

Only `hooks.js`. It's strictly related here.

I've just made your tests more readable, so it looks more like the other tests.

#### 4

> There are some changes leftover from the other PR that appear to be extraneous - `defineConstPoperty`

In favor of the hooks. See the first point.

#### 5

> changes to hooks normalization

See point two.

## Summarizing:

I'm not splitting anything. All the things are related to hooks. I recommend looking at the PR history to see how the things had been done before.

jstewmon:
Per the previous discussion on #523:
- options should be frozen or not, without exceptional cases, so we don't need any of the changes that allow for hooks to be modified if the config is frozen.
- only known hooks should be validated, so that callers are not unnecessarily constrained on what is a valid object
- Object.entries is not an equivalent for `for (... of ...)` or `Array.prototype.forEach`

I thought we were in agreement on those points, which is makes the three changes I mentioned above unrelated. I think @sindresorhus agrees (based on thumbs up on my request to split)?

szmarczak:
> options should be frozen or not, without exceptional cases, so we don't need any of the changes that allow for hooks to be modified if the config is frozen.

I think you're right 👍 If `defaults.preventChanges` is set to true, hooks should be frozen. If it's set to false, hooks shouldn't be frozen. Good point!

> only known hooks should be validated, so that callers are not unnecessarily constrained on what is a valid object

The situation you describe was active in the previous PR. [It's not active anymore](https://github.com/szmarczak/got/blob/c4de6a4b4558fb8c821f5981a5fee6b95f182cff/source/normalize-arguments.js#L191). :)

> Object.entries is not an equivalent for `for (... of ...)` or Array.prototype.forEach

I still don't see what's wrong with that. People can use this feature to name hooks.

### *It's not a bug. It's a feature!*

`got.hooks.beforeRequest.mySuperHook = ...`

szmarczak:
Or you know what? Fine. I'll make a separate PR for "reformatting".

szmarczak:
@jstewmon Is it better now? I could take off the shortcut for hooks and make another PR for that, but that's just only a few lines so I don't think it's needed.

@sindresorhus Can you say something regarding this PR?

source/create.js
```diff
@@ -51,18 +53,19 @@ const create = defaults => {
 		return defaults.handler(url, options, next);
 	};
 
+	defaults.options.hooks = {
+		beforeRequest: [],
+		...(defaults.options.hooks || {})
+	};
+	got.hooks = defaults.options.hooks;
```
jstewmon: Does this facilitate something? I'd avoid duplicating access points.

source/create.js
```diff
@@ -51,18 +53,19 @@ const create = defaults => {
...
 	}
 
-	Object.assign(got, errors);
-	Object.defineProperty(got, 'defaults', {
-		value: deepFreeze(defaults),
-		writable: false,
-		enumerable: true,
-		configurable: true
-	});
+	defineConstProperty(got, {...errors, create}, {deepFreeze: true});
```
jstewmon: I think this is overkill. If it's not overkill, I guess I'd say just call `defineConstProperty(got, got)` b/c it looks like `stream` and `extend` were arbitrarily omitted.

source/define-const-property.js
```diff
@@ -0,0 +1,23 @@
...
+	return Object.freeze(obj);
+};
+
+module.exports = (where, properties, options = {}) => {
+	for (const [key, value] of Object.entries(properties)) {
+		Object.defineProperty(where, key, {
+			value: options.deepFreeze ? deepFreeze(value) : Object.freeze(value),
+			writable: false,
+			enumerable: true,
+			configurable: true
```
jstewmon: I see it was like this before, but why is `configurable` set to `true`? We're locking it unless you're serious enough about changing it to redefine the property?

jstewmon:
> I still don't see what's wrong with that. People can use this feature to name hooks.

They can't b/c we iterate the hooks as an array to evaluate them. The API describes that the hooks are `Array<Function>`, so that's what I suggest we validate.

I don't have a great example why someone would attach a property to an Array, but I'm sure it happens.

Honestly, I can't believe we discussing whether it's sensible to iterate an Array with Object.entries. I think it's nonsensical.

source/create.js
szmarczak: Stream should be in. Extend no, because if someone doesn't like that function he can overwrite it :) What do you think?

source/define-const-property.js
szmarczak: Yeah. It should be set to `false` (default value).

source/create.js
jstewmon: `got.defaults.options.hooks` is duplicated. Consider why all options properties aren't copied to the got instance.

source/create.js
jstewmon: I would just get rid of it. When an object has a custom interface, those customizations should be documented, so that users have a way to learn about them. Documenting this stuff adds clutter to the documentation, adding friction for users.

advanced-creation.md
```diff
@@ -114,3 +114,10 @@ const unicorn = got.create(settings);
 // Same as:
 const unicorn = got.extend({headers: {unicorn: 'rainbow'}});
 ```
+
+##### preventChanges
```
brandon93s: I'm not sure we want to allow things to be mutated by default, or via a configuration for that matter. Why was this necessary or even wanted?

source/define-const-property.js
```diff
@@ -0,0 +1,23 @@
...
+const deepFreeze = obj => {
+	for (const [key, value] of Object.entries(obj)) {
+		if (is.object(value)) {
+			deepFreeze(obj[key]);
+		}
+	}
+
+	return Object.freeze(obj);
+};
+
```
brandon93s: A method named `defineConstProperty` with configurable behavior doesn't feel right.

szmarczak:
I've just realized this PR doesn't change anything.

source/create.js
szmarczak: No, it's not. Consider looking at the JS docs ;) Or google `object reference javascript`.

```js
let a = {};
let b = a;
a.c = 123;
console.log(a);
```

jstewmon: You didn't parse my statement the way it was meant:

> I'd avoid duplicating access points.

I was saying you're creating more than one way of accessing the same object: `got.hooks` and `got.defaults.options.hooks`. The paths to access are duplicated.

Miscommunication and disagreement are bound to happen in async discourse, but we can figure things out without condescension. We can better understand each other by asking questions until the meaning is clear without presuming the other person is an idiot. 🤗

szmarczak: > I'd avoid duplicating access points.

It's not named `access points`. There's no such thing in Node. I think you meant [`object references`](https://stackoverflow.com/questions/37290747/pass-by-reference-javascript-objects). Google it.

> I was saying you're creating more than one way of accessing the same object: got.hooks and got.defaults.options.hooks.

What's wrong with that? That's a shortcut. `got.defaults.options.hooks` is a long word, isn't it?

> We can better understand each other by asking questions until the meaning is clear without presuming the other person is an idiot.

Agreed. Using correct naming would end the argument. :tada: I've searched for `access points` but found nothing.

jstewmon: Technical conversations are usually a mixture of colloquial and technical terms.

I think it's reasonable to expect to be able to converse like we're people, without being rude or disrespectful.

Apparently, those expectations are not going to be met here. :-(

sindresorhus: @szmarczak I know you don't mean anything bad by it, but your tone does come off a bit condescending. If you don't understand a comment, ask for clarification instead of assuming the other part doesn't know what they're talking about. It's easy to talk past each other. Clarity is key in technical conversations.

szmarczak: @sindresorhus Yeah, you're right. My fault. I should better haven't said that. Sorry for being offensive.

@jstewmon Please take my sincere apologies.

> Technical conversations are usually a mixture of colloquial and technical terms.

On StackOverflow that behaviour is not welcome, just saying as I've experienced it myself.

jstewmon: Thanks, no hard feelings

PR #525, opened 2018-07-17
Use events to wire upload tracking instead of racing with setImmediate

jstewmon:
I could not determine why `setImmediate` was being used to defer some of the statements in the event emitter wiring, other than the socket connection check, which can be safely done by subscribing to the right events.

### DIFF
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -91,13 +91,11 @@ module.exports = (options = {}) => {
 				return;
 			}
 
-			setImmediate(() => {
-				try {
-					getResponse(response, options, emitter, redirects);
-				} catch (error) {
-					emitter.emit('error', error);
-				}
-			});
+			try {
+				getResponse(response, options, emitter, redirects);
+			} catch (error) {
+				emitter.emit('error', error);
+			}
 		});
 
 		cacheReq.on('error', error => {
@@ -110,7 +108,7 @@ module.exports = (options = {}) => {
 
 		cacheReq.once('request', req => {
 			let aborted = false;
-			req.once('abort', _ => {
+			req.once('abort', () => {
 				aborted = true;
 			});
 
@@ -133,9 +131,7 @@ module.exports = (options = {}) => {
 				timedOut(req, options.gotTimeout);
 			}
 
-			setImmediate(() => {
-				emitter.emit('request', req);
-			});
+			emitter.emit('request', req);
 		});
 	};
 


source/request-as-event-emitter.js
```diff
@@ -122,8 +120,9 @@ module.exports = (options = {}) => {
 
 		cacheReq.once('request', req => {
 			let aborted = false;
-			req.once('abort', _ => {
+			req.once('abort', () => {
```
szmarczak: Refactoring should be done in a separate PR. Just Kidding :P 👍

source/request-as-event-emitter.js
```diff
@@ -122,8 +120,9 @@ module.exports = (options = {}) => {
...
 				aborted = true;
+				emitter.emit('abort');
```
szmarczak: Do we really need that? It's aborted only when receives an error/someone cancels the request. The error `got` throws has nice details though.

jstewmon: It's not needed. I don't know why I put it there. 🤪

source/request-as-event-emitter.js
jstewmon: I considered it, but when I looked at the diff, I didn't think the one line change raised the cognitive burden enough to warrant a separate commit. 😅

source/request-as-event-emitter.js
szmarczak: I don't know.

source/request-as-event-emitter.js
```diff
@@ -103,13 +103,11 @@ module.exports = (options = {}) => {
 				return;
 			}
 
-			setImmediate(() => {
-				try {
-					getResponse(response, options, emitter, redirects);
-				} catch (error) {
-					emitter.emit('error', error);
-				}
-			});
```
brandon93s: I suspect this was added to control the call stack, ensuring that the method ran to completion before any errors were emitted. This changes that behavior. There doesn't appear to be any tests failing from the change, but I would guess that this introduces behavioral changes.

brandon93s: Originally added in https://github.com/sindresorhus/got/pull/117.

jstewmon: @brandon93s Thanks for the reference to #117.

It looks like the intent was to ensure the response data was handled after the event listeners that clear timeouts were run.

If a callback is provided for `request`, it will be the first handler to run.

I can make another pass at this, to evaluate when the timeout clearing handlers are being run.

Does my analysis sound right? See any other issues?

source/request-as-event-emitter.js
```diff
@@ -148,8 +146,7 @@ module.exports = (options = {}) => {
 					total: uploadBodySize
 				});
 
-				const socket = req.connection;
-				if (socket) {
+				req.once('socket', socket => {
```
szmarczak: Incorrect implementation here. See why:

1. 30c39bc1c3042a80ed3f6752677e0ab42cb1ca98
2. 13d6b68ba02413374519f490d75741a427d705ad

jstewmon: I've reviewed those issues and do not think this change regresses either of them.

- The problem reported in #427 and fixed in #429 was that `req.connection` may be undefined. I've removed the reference to `req.connection` in favor of subscribing to the `socket` event to get the `connection`. So, this should not regress that issue b/c we don't try to access a property which may be undefined.
- The memory leak fix from #465 is addressed by the `if (socket.connecting)` check. It prevents unbounded event subscriptions with keep-alive is enabled on the agent.

szmarczak: I've looked deeply and I think it's safe to do so. Good job!

source/request-as-event-emitter.js
szmarczak: You both are right. The main intent was to emit response ASAP, so @floatdrop used `setImmediate` as it executes immediately after I/O operations. I'd leave `setImmediate` unchanged here.

> If a callback is provided for request, it will be the first handler to run.

I think @jstewmon is right here. We don't need `setImmediate` in this case (line 198).

source/request-as-event-emitter.js
```diff
@@ -183,26 +180,22 @@ module.exports = (options = {}) => {
...
-				}
+				});
 			});
 
 			if (options.gotTimeout) {
 				clearInterval(progressInterval);
 				timedOut(req, options.gotTimeout);
 			}
 
-			setImmediate(() => {
```
jstewmon: Note: this setImmediate deferral was introduced in https://github.com/jstewmon/got/commit/a1eb3f7 in exchange for `get` being deferred. `get` is currently deferred (didn't track down the commit), so this does not need to be deferred.

Either `get` or `emitter.emit('request', req)` needs to be deferred to give callers a chance to subscribe to the `request` event.

source/request-as-event-emitter.js
jstewmon: > The main intent was to emit response ASAP

That's not how I read it. There are 3 emitters involved, so I'm not sure which one you mean. The `emitter` response event is emitted from `getResponse`, so `setImmediate` here delays that. The `req` emitter's event isn't affected by this `setImmediate`.

This is the background information I found:

The version of `timed-out` being used at the time of #117 listened for the `response` event of the request to clear the timeout:
https://github.com/floatdrop/timed-out/blob/v2.0.0/index.js#L31

The `request` object was passed to timed-out, which would setup that listener.

The statement that was deferred with `setImmediate` called `unzipResponse` from the `http.request` callback, which would have run before timed-out's listener.

So my assessment was that the intent was to allow timed-out's listener to run before the response stream was connected to unzipResponse.

*Now that `timeout.request` is waiting for `response.end`, I think deferring the call to `getResponse` actually increases the possibility that the timeout will breach.*

Since it is quite common to use a request callback, I think a better solution to checking a timeout against the `response` event would be to use setImmediate in the timer callback to defer the error emission to the [`check` phase](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/) of the event loop in case the response had been received but was still queued due to load. At a minimum, the listener that clears the timeout could be subscribed with `prependListener` to ensure it runs earlier.

I thought there might be some practical purpose other than the ordering of the response handlers (like allowing subscriptions to some event emitter), but I don't see any such purpose, and I think we can clearly see that the `response` event listener order is no longer applicable.

szmarczak: > That's not how I read it.

If an error was emitted before the response, then boom.

> The emitter response event is emitted from getResponse, so setImmediate here delays that.

When #117 was merged there was sense to do that.

> I think a better solution to checking a timeout against the response event would be to use setImmediate in the timer callback

Can you elaborate? The timer callback clears the timeout AFAIK.

jstewmon: > If an error was emitted before the response, then boom.

I don't follow. The code in question is in the request callback, which is listening for the response event.

Here's some code to illustrate what I meant about using setImmediate with timers. The big comment elaborates on the rationale.

```js

	let connected = false;
	let connectTimeout = setTimeout(
		() => {
			if (!connected) {
				// Timers run first in the event loop, so schedule the error
				// after the IO events in this cycle in case the connection is
				// ready. When load is low, this doesn't matter, but when load
				// is high, there may be many IO events that are ready on each
				// cycle, and the intent of the timeout is to cancel operations
				// which are not ready.
				setImmediate(
					() => {
						if (!connected) {
							req.abort();
							const e = new Error('Connection timed out on request' + host);
							e.code = 'ETIMEDOUT';
							req.emit('error', e);
						}
					}
				)
			}
		},
		delays.connect
	);
	req.once('socket', socket => {
		connected = !socket.connecting;
		if (connected) {
			clearTimeout(connectTimeout);
		} else {
			socket.once('connect', () => {
				connected = true;
				clearTimeout(clearTimeout);
			})
		}
	});
```

szmarczak: So, you just delay throwing the error in favor of IO job (e.g. emitting the response)? If it emits the response, `got` instance resolves, but the error will be thrown anyway. The test you made (#528) would fail.

@floatdrop added here `setImmediate` to emit `response` ASAP (to clear the timeouts, before there are any errors thrown).

jstewmon: I'm using a [double-check locking pattern](https://en.wikipedia.org/wiki/Double-checked_locking) to conditionally emit the error IFF the timer expired and the awaited resource was not ready in the poll phase queue. Take note that `connected` is checked in both the setTimeout callback and the setImmediate callback.

> @floatdrop added here setImmediate to emit response ASAP (to clear the timeouts, before there are any errors thrown).

Do you mean that setImmediate was added to allow the timeout event listeners to run before the response stream is handled by getResponse?

I gave a detailed explanation above about why the event emission order is not actually affected by the setImmediate call. I also explained why the order of the listeners is not currently relevant b/c there are no timeouts that depend on the timing of the `response` event.

I am interested in adding a timeout for the `response` event b/c it's useful to abort the request if the server takes too long to respond to the request.

I'm suggesting that it is more expressive and, thus, more maintainable to address the problem of timeout errors being spuriously emitted when a system is under load by deferring timeout error emission until the poll phase following the timer expiration, which is what I've postulated was the intent behind wrapping `getResponse` (formerly `unzipResponse`) with `setImmediate`.

szmarczak: Sorry, I just missed the second `if` somehow.

> Do you mean that setImmediate was added to allow the timeout event listeners to run before the response stream is handled by getResponse?

Yes.
> On request-heavy applications (with short timeout option) lots of requests will be rejected, because of response handlers.

Before:

1. First request:
   - timeout has been cleared
   - [this thing](https://github.com/sindresorhus/got/blob/001351cb067e3553bb551b806fc569fffb385b37/index.js#L90-L114) makes some delay, because there's no `setImmediate` [here](https://github.com/sindresorhus/got/blob/001351cb067e3553bb551b806fc569fffb385b37/index.js#L56)
2. Second request: `sorry, but too much time has passed`, throws an error.

After:
1. First request:
   - timeout has been cleared
   - [this thing](https://github.com/sindresorhus/got/blob/9ffbbdf5aba98892519282ad7799a991c0b04483/index.js#L92-L116) doesn't make a big delay, because `setImmediate` is used [here](https://github.com/sindresorhus/got/blob/9ffbbdf5aba98892519282ad7799a991c0b04483/index.js#L56-L58)
2. Second request: `ok, go on`
3. First request: success.

> I am interested in adding a timeout for the response event b/c it's useful to abort the request if the server takes too long to respond to the request.

Use `socket` timeout. If you use `response.setTimeout()` it'll throw unless whole response is received I think.

szmarczak: I just want to note we're talking about completely different things. They are not the same. Your example just gives the last chance for establishing the connection, but I don't know if that's the correct behavior. `connect` event doesn't make any delays here.

jstewmon: I think we're stuck talking in a circle. Maybe someone else can chime in an help us converge our understanding.

@sindresorhus @brandon93s @knksmith57 any clarifying insights you can add?

szmarczak: [Your example](https://github.com/sindresorhus/got/pull/525#discussion_r203699062) just gives the last chance for establishing the connection after the timeout has been called. Am I wrong?

jstewmon: I wouldn't say you're wrong, but I think it's more subtle than that. Have you carefully reviewed the doc I linked earlier about the [event loop phases]?

The timers get to run first, then the data callbacks, then setImmediate callbacks.

So, when the timer breaches, if the system has been busy, the data callbacks may be ready to execute, but they haven't had their turn. So, if we move the emit error call to setImmediate with that double-check pattern, we can allow the busy system to process the data callbacks that are ready but haven't had their turn.

The point is that if we're checking `connect` timeout. The connection may have already been established, but its event hasn't fired when the timeout callback is run.

[event loop phases]: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/

szmarczak: > I think a better solution to checking a timeout against the response event would be to use setImmediate in the timer callback to defer the error emission to the check phase of the event loop in case **the response had been received but was still queued due to load**.

So the connection was established, because there's a response.

> Here's some code to illustrate what I meant about using setImmediate with timers.

You were trying to illustrate the first sentence using the `connect` timeout, though it had been connected already. Awesome.

Is it possible to receive `connect` event and `response` (with some queued data) in the same loop?

szmarczak: > Have you carefully reviewed the doc I linked earlier about the event loop phases?

Yes, I have read it several times, not only today. 


> The point is that if we're checking connect timeout. The connection may have already been established, but its event hasn't fired when the timeout callback is run.

Then the connection is not established. From the [docs](https://nodejs.org/api/net.html#net_event_connect):

> Event: 'connect'
> Emitted when a socket connection is successfully established.

> socket.connecting
> If true - socket.connect(options[, connectListener]) was called and haven't  yet finished. Will be set to false before emitting 'connect' event and/or  calling socket.connect(options[, connectListener])'s callback.

jstewmon: This is a tricky topic because of the changes in behavior over time.

When I've referred to the `response` event, I've been talking about the behavior and intent at the time (circa 2015) that the setImmediate deferral was added in #117. The `response` event is not currently used to clear any timeouts, so it is not relevant to this change.

> You were trying to illustrate the first sentence using the connect timeout, though it had been connected already.

No, the statement you quoted was in my original analysis. I was suggesting that there was a more precise way of mitigating #117 than deferring the response piping with setImmediate.

I didn't want to use the response event to illustrate how that suggestion would be implemented because none of the current timeouts depend on the timing of that event. I used the `connect` event as an example in the current code because it is used to clear the connect timeout.

> Is it possible to receive connect event and response (with some queued data) in the same loop?

I don't see any reason why that would not be possible, but their listeners would be notified in FIFO order (connect before response). Does it matter?

jstewmon: > Then the connection is not established

The underlying system operation completes before we're notified. That's the purpose of the poll phase - to notify us about the IO operations we listened for are complete.

sindresorhus:
Closing in favor of #534

PR #526, opened 2018-07-18
Link to current docs

millette:
I hadn't used got in a while. Saw the docs, many fancy updates looking great, but when it came time to use them, there were not implemented (in got 8.3.2). A tip at the top of the docs liking to current version might be helpful.


sindresorhus:
Thanks for the suggestion, but this is expected knowledge and how most projects on GitHub operate.

PR #530, opened 2018-07-19
Expose `assignOptions`

szmarczak:
IOW: the correct way to inherit from options.

### DIFF
--- a/advanced-creation.md
+++ b/advanced-creation.md
@@ -11,16 +11,8 @@ Configure a new `got` instance with the provided settings.<br>
 
 ##### [options](readme.md#options)
 
-To inherit from parent, set it as `got.defaults.options` or use `got.assignOptions(defaults.options, options)`.<br>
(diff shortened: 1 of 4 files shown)


advanced-creation.md
```diff
@@ -11,7 +11,16 @@ Configure a new `got` instance with the provided settings.<br>
 
 ##### [options](readme.md#options)
 
-To inherit from parent, set it as `got.defaults.options` or use [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals).
+To inherit from parent, set it as `got.defaults.options` or use `got.assignOptions(defaults.options, options)`.<br>
+Avoid using [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) as it may give unwanted result:
```
sindresorhus: `as it may give unwanted result:` => `as it doesn't work recursively:`

sindresorhus:
Makes sense to me, but then `assignOptions` needs to be documented properly in the API docs: https://github.com/sindresorhus/got#api

szmarczak:
Yeah, completely forgot about this.

PR #531, opened 2018-07-19
Move upload progress plumbing to its own module

jstewmon:
This PR is based on #525 .

The inlining of upload progress bloated the essential request wiring, making it hard to grok.

Moving the upload progress feature to its own module makes both sections of code easy to grok.


jstewmon:
I rebased this on master, since #525 is still being discussed, and the two PRs are not coupled.

PR #532, opened 2018-07-19
Pass normalized options to the handler

szmarczak:
That's just an improvement.

### DIFF
--- a/source/normalize-arguments.js
+++ b/source/normalize-arguments.js
@@ -16,22 +16,24 @@ module.exports = (url, options, defaults) => {
 		throw new TypeError('Parameter `url` is not an option. Use got(url, options)');
 	}
 
-	if (options.baseUrl && (is.string(url) || is(url) === 'URL')) {
-		url = urlToOptions(new URLGlobal(url, options.baseUrl));
-	} else if (!is.string(url) && !is.object(url)) {
+	if (!is.string(url) && !is.object(url)) {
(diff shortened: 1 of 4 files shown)


source/normalize-arguments.js
```diff
@@ -15,7 +16,9 @@ module.exports = (url, options, defaults) => {
 		throw new TypeError('Parameter `url` is not an option. Use got(url, options)');
 	}
 
-	if (!is.string(url) && !is.object(url)) {
+	if (options.baseUrl && (is.string(url) || is(url) === 'URL')) {
+		url = urlToOptions(new URLGlobal(url, options.baseUrl));
+	} else if (!is.string(url) && !is.object(url)) {
 		throw new TypeError(`Parameter \`url\` must be a string or object, not ${is(url)}`);
```
sindresorhus: I think this should be moved out of the if/else into a standalone `if` above:

```js
if (!is.string(url) && !is.object(url)) {
	throw new TypeError(`Parameter \`url\` must be a string or object, not ${is(url)}`);
}

if (options.baseUrl && (is.string(url) || is(url) === 'URL')) {
...
```

sindresorhus:
Other than the https://github.com/sindresorhus/got/pull/532#discussion_r204534554 nitpick, this looks good to me :)

source/normalize-arguments.js
szmarczak: Yep. Doing right now.

PR #534, opened 2018-07-23
Customize timeouts and generally improve the whole thing

jstewmon:
- Refactor timed-out to optimistically defer errors until after the poll
  phase of the current event loop tick.
- Timeouts begin and end when their respective phase of the lifecycle
  begins and ends.
- Timeouts result in a `got.TimeoutError`

This includes the controversial  change discussed in #525: removal of `setImmediate` around response handling and the addition of `setImmediate` in the timeout handling. Those change are not strictly required and can be reverted if necessary and everything in this PR will still work.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -188,7 +188,13 @@ Type: `number` `Object`
 
 Milliseconds to wait for the server to end the response before aborting request with `ETIMEDOUT` error (a.k.a. `request` property). By default there's no timeout.
 
-This also accepts an object with separate `connect`, `socket`, and `request` fields for connection, socket, and entire request timeouts.
+This also accepts an object with separate `lookup`, `connect`, `socket`, `response` and `request` fields to specify granular timeouts for each phase of the request.
+
+- `lookup` starts when a socket is assigned and ends when the hostname has been resolved. Does not apply when using a unix domain socket.
+- `connect` starts when `lookup` completes (or when the socket is assigned if lookup does not apply to the request) and ends when the socket is connected.
+- `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
+- `response` starts when the socket is connected and ends when the response headers are received.
+- `request` starts when the request is initiated and ends when the response's end event fires.
 
 ###### retry
 
--- a/readme.md
+++ b/readme.md
@@ -190,7 +190,7 @@ Milliseconds to wait for the server to end the response before aborting request 
 
 This also accepts an object with separate `lookup`, `connect`, `socket`, `response` and `request` fields to specify granular timeouts for each phase of the request.
 
-- `lookup` starts when a socket is assigned and ends when the hostname has been resolved. Does not apply when using a unix domain socket.
+- `lookup` starts when a socket is assigned and ends when the hostname has been resolved. Does not apply when using a Unix domain socket.
 - `connect` starts when `lookup` completes (or when the socket is assigned if lookup does not apply to the request) and ends when the socket is connected.
 - `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
 - `response` starts when the request has been written to the socket and ends when the response headers are received.
--- a/readme.md
+++ b/readme.md
@@ -193,7 +193,7 @@ This also accepts an object with separate `lookup`, `connect`, `socket`, `respon
 - `lookup` starts when a socket is assigned and ends when the hostname has been resolved. Does not apply when using a unix domain socket.
 - `connect` starts when `lookup` completes (or when the socket is assigned if lookup does not apply to the request) and ends when the socket is connected.
 - `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
-- `response` starts when the socket is connected and ends when the response headers are received.
+- `response` starts when the request has been written to the socket and ends when the response headers are received.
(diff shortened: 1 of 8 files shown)


readme.md
```diff
@@ -188,7 +188,13 @@ Type: `number` `Object`
 
 Milliseconds to wait for the server to end the response before aborting request with `ETIMEDOUT` error (a.k.a. `request` property). By default there's no timeout.
 
-This also accepts an object with separate `connect`, `socket`, and `request` fields for connection, socket, and entire request timeouts.
+This also accepts an object with separate `lookup`, `connect`, `socket`, `response` and `request` fields to specify granular timeouts for each phase of the request.
```
jstewmon: Being able to set timeouts around each phase is more useful that timing each phase from the beginning of the request.

A concrete example is being able to time the `response` phase (TTFB) when the target is a load-balanced java service which may be undergoing garbage collection. We often want to set a very aggressive response timeout in this case, so that we can retry the request immediately against another instance.

source/request-as-event-emitter.js
```diff
@@ -119,23 +117,23 @@ module.exports = (options = {}) => {
 					return;
 				}
 
-				const err = new RequestError(error, options);
-				emitter.emit('retry', err, retried => {
+				if (!(error instanceof GotError)) {
```
jstewmon: I wasn't sure that this was the precise change needed. My objective was to expose the TimeoutError to a client using the Promise interface.

source/request-as-event-emitter.js
```diff
@@ -91,13 +91,11 @@ module.exports = (options = {}) => {
...
-					getResponse(response, options, emitter, redirects);
-				} catch (error) {
-					emitter.emit('error', error);
-				}
-			});
+			try {
+				getResponse(response, options, emitter, redirects);
+			} catch (error) {
+				emitter.emit('error', error);
+			}
```
szmarczak: Can we leave this unchanged as this is being discussed in #525? Try not to duplicate PRs.

readme.md
```diff
@@ -188,7 +188,13 @@ Type: `number` `Object`
...
+
+- `lookup` starts when a socket is assigned and ends when the hostname has been resolved. Does not apply when using a unix domain socket.
+- `connect` starts when `lookup` completes (or when the socket is assigned if lookup does not apply to the request) and ends when the socket is connected.
+- `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
+- `response` starts when the socket is connected and ends when the response headers are received.
+- `request` starts when the request is initiated and ends when the response's end event fires.
```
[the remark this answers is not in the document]
jstewmon: I think the names should be as transparent as possible. lookup, connect and response are nice because they map directly to the event that ends the phase.

request and socket are the outliers, but I left them as they were to maximize backwards compatibility.

If we want to forego backwards compatibility, I'd suggest renaming `request` to `overall` or `total`. `socket` may be more expressive as `idleSocket`, but I think it's too pedantic to justify the incompatibility.

source/timed-out.js
```diff
@@ -1,79 +1,120 @@
...
+				cancelers.push(cancelTimeout);
+				return cancelTimeout;
+			};
+			if (socket.connecting) {
+				socket.once('connect', () => {
+					req.once('response', timeResponse());
+				});
+			} else {
+				req.once('response', timeResponse());
+			}
```
szmarczak: ❤️ 🙌

readme.md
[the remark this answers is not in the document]
szmarczak: > I'd suggest renaming request to overall or total.

👍 

> socket may be more expressive as idleSocket, but I think it's too pedantic to justify the incompatibility.

Agreed. `socket` is enough IMO.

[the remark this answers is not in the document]
jstewmon: @sindresorhus what do you think about changing `request` to `overall` or `total`?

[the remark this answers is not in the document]
jstewmon: Following up on this, I realized that the response timer would start too soon if the request spent significant time writing to the socket, so I implemented a fix for that by emitting a custom event on the request when the body has been written.

This also enabled the addition of a `send` timer.

If we were starting from scratch I would name the `send` timer `request` and I would name the `request` timer `overall`. But, I worry that changing the meaning of those names is subtle enough to go overlooked by users, leading to bugs that can be avoided by just keeping the imperfect names.

jstewmon:
@szmarczak thanks for the flattering review :-)

test/timeout.js
```diff
@@ -27,69 +45,211 @@ test('timeout option (ETIMEDOUT)', async t => {
...
+		}),
+		{
+			...errorMatcher,
+			message: `Timeout awaiting 'response' for 1ms`
+		}
+	);
+});
+
+test('response timeout (slow upload)', async t => {
+	const body = fs.createReadStream('/dev/urandom', {
```
jstewmon: Ugh, I forgot about AppVeyor. I'll look for an alternative. Suggestions welcome.

sindresorhus: I made https://github.com/sindresorhus/random-bytes-readable-stream, but I see you already solved the problem :)

jstewmon: That looks handy :-)

source/request-as-event-emitter.js
sindresorhus: I agree, this should not have been included here, but since I'm gonna accept #525, I'll allow it.

sindresorhus:
@jstewmon This is great! Thanks for doing such a thorough investigating into timeouts.

PR #535, opened 2018-07-25
docs: DRY the timeout object section

jstewmon:
I noticed that `send` was not in the first list, then realized we probably don't need to enumerate the object keys twice.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -186,7 +186,7 @@ Query string object that will be added to the request URL. This will override th
 
 Type: `number` `Object`
 
-Milliseconds to wait for the server to end the response before aborting request with `ETIMEDOUT` error (a.k.a. `request` property). By default there's no timeout.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -186,9 +186,9 @@ Query string object that will be added to the request URL. This will override th
 
 Type: `number` `Object`
 
-Milliseconds to wait for the server to end the response before aborting request with `ETIMEDOUT` error (a.k.a. `request` property). By default there's no timeout.
+Milliseconds to wait for the server to end the response before aborting request with [`got.TimeoutError`](#gottimeouterror) error (a.k.a. `request` property). By default there's no timeout.
```
jstewmon: I updated these when I was working on #534, but I apparently lost the diff somewhere along the way.

szmarczak: Ok. Seems good. Is it ready to merge?

jstewmon: Yes, sorry for tacking that on to this PR. I didn't expect you to review so quickly. I'm not working on any more changes. :-)

PR #537, opened 2018-07-25
Fix: remove ESOCKETTIMEDOUT from retry whitelist and update docs

jstewmon:


readme.md
```diff
@@ -216,10 +216,9 @@ Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.ra
 Option `retries` can be a `number`, but also accepts a `function` with `retry` and `error` arguments. Function must return delay in milliseconds (`0` return value cancels retry).
 
 **Note:** It retries only on the specified methods, status codes, and on these network errors:
-- `ETIMEDOUT`: Connection was not estabilished after a period of time.
+- `ETIMEDOUT`: One of the [timeout](#timeout) limits was breached.
```
szmarczak: Wouldn't be this better:
> One of the [timeouts](#timeout) has occured.

I don't think `breach` is a good word here to describe that.

sindresorhus:
Are we absolutely sure `ESOCKETTIMEDOUT` can never be thrown?

jstewmon:
I'm not certain it will never be thrown. I searched through node and libuv without finding any reference to it, but that's not proof.

I ended up removing it b/c I couldn't think of how to describe it in the docs.

szmarczak:
> Are we absolutely sure ESOCKETTIMEDOUT can never be thrown?

Yes. It has been created by [`request`](https://github.com/request/request/issues/2320) originally. I don't see that anywhere else.

@jstewmon We should throw `ECONNRESET: socket hang up` instead of `ETIMEDOUT` when the socket is connected and one of the timeouts has occurred, as Node.js does.

jstewmon:
I'm not convinced we should use `ECONNRESET` for the client's idle socket timeout. It's true that node uses that code, but when node raises the error, it's ambiguous as to which side hung up.

I think we should keep our socket timeout error code as `ETIMEDOUT` to disambiguate it from a hangup by the server.

```
const http = require('http');
const {promisify} = require('util');

const s = http.createServer((req, res) => {
  res.setTimeout(1);
});
const listen = promisify(s.listen.bind(s));

(async () => {
  await listen()
  const request = http.request({port: s.address().port});
  request.on('error', logError)
  request.end();
})();

function logError (error) {
  console.error('%o', error);
}
```

```
{ Error: socket hang up
    at createHangUpError (_http_client.js:331:15)
    at Socket.socketOnEnd (_http_client.js:423:23)
    at emitNone (events.js:111:20)
    at Socket.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1064:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  [stack]: 'Error: socket hang up\n    at createHangUpError (_http_client.js:331:15)\n    at Socket.socketOnEnd (_http_client.js:423:23)\n    at emitNone (events.js:111:20)\n    at Socket.emit (events.js:208:7)\n    at endReadableNT (_stream_readable.js:1064:12)\n    at _combinedTickCallback (internal/process/next_tick.js:138:11)\n    at process._tickCallback (internal/process/next_tick.js:180:9)',
  [message]: 'socket hang up',
  code: 'ECONNRESET' }
```

szmarczak:
Yeah, then better use `ETIMEDOUT`.

PR #538, opened 2018-07-25
Make `got.mergeOptions()` behavior more obvious and document its behavior

jstewmon:
Option merging is now consistent, so null header removal is done during
options normalization. null could not be used to indicate property
removal because null is a significant value for some settings.


source/assign-options.js
```diff
@@ -1,27 +1,36 @@
...
 
-	if (Reflect.has(options, 'headers')) {
-		for (const [key, value] of Object.entries(options.headers)) {
-			if (is.nullOrUndefined(value)) {
-				delete returnOptions.headers[key];
-			}
-		}
+function customizer(objValue, srcValue) {
+	if (is.array(srcValue) || is.array(objValue)) {
+		return cloneDeep(srcValue);
```
jstewmon: This might not be necessary. Maybe we can just pick the right target based on whether is.array(objValue). I was running short on time, so I did this to get it working.

I think we could ditch `cloneDeep` as a dep if that works.

source/assign-options.js
```diff
@@ -1,27 +1,30 @@
...
 module.exports = (defaults, options = {}) => {
-	const returnOptions = extend(true, {}, defaults, options);
+	return merge({}, defaults, options);
+};
 
-	if (Reflect.has(options, 'headers')) {
-		for (const [key, value] of Object.entries(options.headers)) {
-			if (is.nullOrUndefined(value)) {
-				delete returnOptions.headers[key];
+function merge(target = {}, ...sources) {
```
szmarczak: `target = {}` is useless here

test/create.js
```diff
@@ -76,6 +77,36 @@ test('custom headers (extend)', async t => {
 	t.is(headers.unicorn, 'rainbow');
 });
 
+test('extend overwrites arrays', t => {
+	const statusCodes = [408];
+	const a = got.extend({retry: {statusCodes}});
+	t.deepEqual(a.defaults.options.retry.statusCodes, statusCodes);
+	t.is(a.defaults.options.retry.statusCodes, statusCodes);
+});
```
szmarczak: :+1:

test/create.js
```diff
@@ -76,6 +77,36 @@ test('custom headers (extend)', async t => {
...
+		b.defaults.options.headers,
+		{...got.defaults.options.headers, ...{baz: 'baz', bar: null}}
+	);
+});
+
+test('extend merges URL instances', t => {
+	const a = got.extend({baseUrl: new URL('https://example.com')});
+	const b = a.extend({baseUrl: '/foo'});
+	t.is(b.defaults.options.baseUrl.toString(), 'https://example.com/foo');
+});
```
szmarczak: :+1:

source/assign-options.js
```diff
@@ -1,27 +1,30 @@
-const extend = require('extend');
+const url = require('url');
```
szmarczak: use `const {URL} = require('url');`

test/create.js
szmarczak: Consider this example:

```js
test('extend overwrites arrays', t => {
	let statusCodes = [408];
	const a = got.extend({retry: {statusCodes}});
	statusCodes[0] = 500;
	t.deepEqual(a.defaults.options.retry.statusCodes, [408]);
	t.not(a.defaults.options.retry.statusCodes, statusCodes);
});
```

It shouldn't fail. It should deep copy arrays IMO.

szmarczak: Oh. Defaults are deep frozen here: https://github.com/sindresorhus/got/blob/da7f055749780438999d459dca5455a61abab8ba/source/create.js#L45

and ain't cloned before.

test/create.js
```diff
@@ -76,6 +77,36 @@ test('custom headers (extend)', async t => {
...
+
+test('extend removes object values set to undefined (root keys)', t => {
+	t.true('headers' in got.defaults.options);
+	const a = got.extend({headers: undefined});
+	t.false('headers' in a.defaults.options);
+});
```
szmarczak: :-1: It should keep the old value.

test/create.js
```diff
@@ -76,6 +77,36 @@ test('custom headers (extend)', async t => {
...
+test('extend removes object values set to undefined', t => {
+	const a = got.extend({
+		headers: {foo: 'foo', bar: 'bar', baz: 'baz'}
+	});
+	const b = a.extend({headers: {foo: undefined, bar: null}});
+	t.deepEqual(
+		b.defaults.options.headers,
+		{...got.defaults.options.headers, ...{baz: 'baz', bar: null}}
+	);
+});
```
szmarczak: :-1: The feature is meant **only** for `headers`. Otherwise, if the new value is `undefined`, it shouldn't be replaced.

test/create.js
jstewmon: Yeah, I made it only deep copy plain object, since all other types are merged by replacement.

test/create.js
jstewmon: I'm fine either way on this one.

I interpreted the previous behavior as `undefined` being a sentinel value that can be used to remove a setting, i.e. a discrete feature of got's extension behavior.

There is a test that verifies that null headers are omitted from the normalized arguments, but I've addressed that case in normalize-arguments.

@sindresorhus do you have a preference?

test/create.js
jstewmon: Ack. This is the same behavior as we're discussing above, so we'll resolve this when we resolve that.

test/create.js
szmarczak: > Just to clarify, based on your complete suggestion below, you think undefined values in the source should be noop and the target value should be preserved, right?

yes.

readme.md
```diff
@@ -408,14 +406,23 @@ client.get('/demo');
 
 #### got.assignOptions(parentOptions, newOptions)
 
-Extends parent options. Avoid using [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) as it doesn't work recursively:
```
szmarczak: I'd leave this note about avoiding `object spread` because mostly that's not what users want.

jstewmon: I pruned it because I didn't think it was necessary for us to document the language feature, and the reader's attention to detail is inversely related to the length of the docs.

I can put it back if you feel strongly that it will meaningfully assist users in making the right choice.

szmarczak: @sindresorhus What are your thoughts on that?

sindresorhus: I think it's worth keeping. I see the mistake of using Object.assign/object-spread for deep objects all the time.

test/create.js
sindresorhus: Both arguments are valid and both are common. `Object.assign`/object-spread will overwrite with `undefined`, while many extends libs will not. I definitely find it more convenient when `{foo: true}, {foo: undefined}` end up as `{foo: true}`, but that also means there's no way to explicitly unset a value, like with headers. Maybe we do a middle ground and say that `null` overrides, while `undefined` does not. This is how default function arguments work. They only work with `undefined` not `null`. So to unset a header you set `null`.

```
{foo: true}, {foo: false}
// {foo: false}

{foo: true}, {foo: undefined}
// {foo: true}

{foo: true}, {foo: null}
// {foo: null}
```

Thoughts?

szmarczak: I strongly agree :) This behaviour is more clear IMO.

szmarczak: @jstewmon Could you allow edits from maintainers? I did what @sindresorhus said, I just wanted to push one commit. You'll change that (or revert) if you'd like to.

test/headers.js
```diff
@@ -141,13 +141,3 @@ test('remove null value headers', async t => {
...
-
-test('remove undefined value headers', async t => {
-	const {body} = await got(s.url, {
-		headers: {
-			'user-agent': undefined
-		}
-	});
-	const headers = JSON.parse(body);
-	t.false(Reflect.has(headers, 'user-agent'));
-});
```
szmarczak: I wouldn't support this, as it might be confusing with `got.extend({ ... })`, where setting to `undefined` keeps the old value.

test/create.js
jstewmon: I lightly object b/c ignoring undefined precludes the possibility of removing a setting (hypothetically, null may be a significant of invalid value). I can't think of a reason why that would be a problem today. But, I also can't think of a reason why an object would contain a key with the value set to undefined in this context b/c the key could simply be omitted to get the same behavior.

In other words, the key's absence or presence is not significant if the value is undefined, which is not something I'd expect as a user. If I set the key in the passed options, I did so with a purpose.

@szmarczak , I backed off your change b/c it overreached what we had discussed. I don't mind maintainer commits for nits or agreed upon changes, but please don't make behavioral changes when consensus has not been reached.

I think the only significant change I reverted was the deep merging of arrays. I did that because got does not make changes to those arrays, so there is no need to make a copy of them, so copying adds unnecessarily to the runtime of the method.

test/headers.js
jstewmon: This test had to be rewritten to use a header key that doesn't have a default value. The behavior is consistent as you wished. :-)

test/create.js
jstewmon: We're trying to build consensus around these changes. Making a fork at this point is counterproductive.

If there's a change you want to see, we should talk about it.

szmarczak: Already did.

jstewmon: You 👍 my explanation for why arrays are directly assigned, then changed the behavior in a push to my branch.

You rewrote the docs I wrote without having provided any remarks.

You implemented @sindresorhus proposition on my branch without giving me a chance to respond.

I can appreciate that you are eager to land this, but we've had a productive conversation and are very close to achieving consensus on the implementation. By forging ahead, you undercut that process and then we have to have these unproductive meta conversations.

What's the rationale for, or problem solved, by deep merging arrays?

szmarczak: > What's the rationale for, or problem solved, by deep merging arrays?

You don't even simply clone them, so let's start with that. With arrays there should be no exception. I'm not gonna explain simple things.

I really appreciate you want to do it on your own :)

sindresorhus: > I'm not gonna explain simple things.

Arguments are worthless if the other part doesn't understand you. Not everyone is in your exact mindset. We all need to be on the same page. I was also confused by your comment.

sindresorhus: I just realized that my proposal is how `lodash.merge` works.

> But, I also can't think of a reason why an object would contain a key with the value set to undefined in this context b/c the key could simply be omitted to get the same behavior.

It allows you to do inline conditionals:

```js
foo({
	unicorn: checkForUnicorns() && '🦄'
});
```

Without, I would have to do:

```js
const options = {};

if (shouldUseUnicorns()) {
	options.unicorn = '🦄';
}

foo(options);
```

> I backed off your change b/c it overreached what we had discussed. I don't mind maintainer commits for nits or agreed upon changes, but please don't make behavioral changes when consensus has not been reached.

@szmarczak Don't modify other people's PRs before we have agreed on a solution.

jstewmon: @sindresorhus thanks for providing that example. I can appreciate the convenience that provides.

It occurred to me that, in the future, we could expose a symbol support deleting a key if the need for it arises. Similarly, we could expose a symbol to indicate that the result should be `undefined` if there's a need to explicitly set a key to `undefined`.

Using symbols for those cases would have the disadvantage of being a unique approach, but it would have the advantage of our interface being optimized for the most common scenarios.

If we're settled on ignoring `undefined`, we could rename `assignOptions` to `mergeOptions` to help clarify that its behavior is not congruent with Object.assign.

readme.md
jstewmon: On closer inspection, I didn't actually remove this - I just inserted the function description before the warning. :-)

test/create.js
szmarczak: > Arguments are worthless if the other part doesn't understand you. Not everyone is in your exact mindset. We all need to be on the same page. I was also confused by your comment.

I'll sum up: going through shortcuts won't end well, just a hint. Another hint: references, arrays, deep freeze. This says a lot.

> Don't modify other people's PRs before we have agreed on a solution.

I've just proposed a solution based on yours. Do what you want, I'm off. I'll be happy with any (current one is cool too).

test/create.js
szmarczak: > Yeah, I made it only deep copy plain object, since all other types are merged by replacement.

I don't think obvious things need an explanation, do they? You've just repeated yourself: that's what your code does, doesn't it? Please come up with a real argument. I'll illustrate if you have problems with understanding mine:

1. Someone creates an array.
2. The array is passed to assignOptions through `got.extend()`.
3. It is passed by reference (please keep that word in mind my friend).
4. It's deep frozen (got.extend() -> got.create() -> deepFreeze())
5. Someone wants to reuse the array.
6. :boom:

jstewmon: I understand that concern, but I intentionally kept the implementation simple, so that the results are predictable.

I was also concerned with performance, since this method is called not only by `got.extend` and `got.create`, but also by every call to `got()`.

There are an infinite number of possible reference types that inherit from object, which may also have unexpected behavior when cloned, especially when only an object's own properties are cloned.

I reluctantly chose to recursively merge source values that are plain objects when the target value is not a plain object for this reason because normalize-arguments currently requires that the options it receives are mutable. I would rather have refactored normalize-arguments, so that it selectively builds a new object from the input, but I thought that was too large a change to introduce here.

Do you think it would be helpful to explicitly mention in the `got.extend` and `got.create` that the resulting object is frozen, so that any reference types will be frozen? We could also mention that a workaround is to use something like `_.deepClone` if users need to preserve a mutable instance of a config setting.

We could also reconsider whether freezing the options tips the balance of predicable behavior nearer or farther from the user's expectations.

jstewmon: Just clarify my position - I'm not strictly inflexible on the implementation, but I want to make sure my rationale is clearly expressed before we decide to change it.

@sindresorhus do you have an opinion on this topic?

szmarczak: > I was also concerned with performance, since this method is called not only by got.extend and got.create, but also by every call to got().

Then why? There are many nitpicks that could improve performance **a bit** (it's unnoticeable) like this one and you're nothing about that. Please, be serious.

> Do you think it would be helpful to explicitly mention in the got.extend and got.create that the resulting object is frozen, so that any reference types will be frozen?

No. That's dumb. All plain objects are frozen and other objects like arrays are not? That doesn't have any sense. BTW, It's OK to pass references to other types than objects and arrays since `got` doesn't use them.

szmarczak: @jstewmon Nice copycat work. GL ITF.

test/headers.js
```diff
@@ -145,9 +145,9 @@ test('remove null value headers', async t => {
 test('remove undefined value headers', async t => {
```
szmarczak: Wait. What's with `user-agent`? You need to change the test name if you change the behaviour too.

test/create.js
```diff
@@ -76,6 +77,42 @@ test('custom headers (extend)', async t => {
...
+
+test('extend merges URL instances', t => {
+	const a = got.extend({baseUrl: new URL('https://example.com')});
+	const b = a.extend({baseUrl: '/foo'});
+	t.is(b.defaults.options.baseUrl.toString(), 'https://example.com/foo');
+});
+
+test('extend ignores object values set to undefined (root keys)', t => {
+	t.true('headers' in got.defaults.options);
```
szmarczak: Use `Reflect.has()` for consistency.

szmarczak: Also I suggest figuring out a better test name.

test/create.js
```diff
@@ -76,6 +77,42 @@ test('custom headers (extend)', async t => {
...
 });
 
+test('extend overwrites arrays', t => {
+	const statusCodes = [408];
+	const a = got.extend({retry: {statusCodes}});
+	t.deepEqual(a.defaults.options.retry.statusCodes, statusCodes);
+	t.is(a.defaults.options.retry.statusCodes, statusCodes);
+});
+
+test('extend overwrites null', t => {
```
szmarczak: After I read the test name I still don't know what's this about.

readme.md
```diff
@@ -358,10 +358,8 @@ Sets `options.method` to the method name and makes a request.
 
 #### got.extend([options])
 
-Configure a new `got` instance with default `options` and (optionally) a custom `baseUrl`:
+Configure a new `got` instance with default `options`. `options` are merged with the extended instance's `defaults.options` as described in [`got.assignOptions`](#got.assignoptionsparentoptions,-newoptions).
```
szmarczak: Stop guessing. You need to change the URL to `#gotassignoptionsparentoptions-newoptions`.

test/create.js
sindresorhus: > Do you think it would be helpful to explicitly mention in the got.extend and got.create that the resulting object is frozen, so that any reference types will be frozen? We could also mention that a workaround is to use something like _.deepClone if users need to preserve a mutable instance of a config setting.

Yes, that would be useful.

sindresorhus:
I've decided. Let's go with ignoring `undefined` and rename the method to `mergeOptions`.

I think we should also deep clone the array, not because it solves all the problems, but it makes it less likely that the user will encounter a problem with references, and I think that is worth the neglible performance we save by not doing it. This is also how `lodash.merge` works and what users would expect:

> Array and plain object properties are merged recursively. Other objects and value types are overridden by assignment. Source objects are applied from left to right. - `lodahs.merge` docs

We can pretty much reuse the above description for our `mergeOptions` docs.

jstewmon:
@sindresorhus I wasn't sure whether you meant to just make a copy of the source array or merge the array into the target's property value (like lodash). I assumed (and implemented) the former, since that's closest to what we've been discussing, but your comment about lodash gave me some doubt.

szmarczak:
@jstewmom You forgot `git add` ;) There's no `merge-options` file.

readme.md
szmarczak: Glad you've taken my tip ^_^

sindresorhus:
> I wasn't sure whether you meant to just make a copy of the source array or merge the array into the target's property value

I meant, make a copy.

sindresorhus:
Thanks for the PR, @jstewmon :)

PR #539, opened 2018-07-27
Improve merging options

szmarczak:
Compared to #538:

- clones `defaults` on instance creation,
- more clear docs a bit,
- `merge()` clones array properties too, which is useful when hooks need to access `hooks.beforeRequest.someProperty`,
- `merge()` is a different file now,
- `[]` instead of `new Array` in `merge()`,
- added `use strict` to `merge.js`,
- `merge()` is now an arrow function.


sindresorhus:
Closing this in favor of #538

There's really no point in opening a new PR. We have not yet decided which direction to go and when we do, you should give @jstewmon a chance to implement it first.

szmarczak:
Let's see where it heads then.

sindresorhus:
Your implementation might be better but you have no description in this PR on how it's better. It's also forking our existing discussion into a new place, which makes it more difficult to keep track of.

sindresorhus:
> Let's see where it heads then.

If we're not happy with the other implementation and it's going nowhere, we can just reopne this, but I'd like to give people a chance to finish their work.

szmarczak:
I completely agree with you.

> I'd like to give people a chance to finish their work.

There's nothing wrong with helping, is it? The thing is I pushed a commit. He could look at what I did and improve it, but he **did not** want it, he reverted all the changes. He wanted to rediscover things which already have been discovered, so feel free. The same goal is achieved, but it's been delayed for days by this way.

sindresorhus:
> There's nothing wrong with helping, is it? The thing is I pushed a commit. He could look at what I did and improve it, but he did not want it, he reverted all the changes. He wanted to rediscover things which already have been discovered, so feel free. The same goal is achieved, but it's been delayed for days by this way.

You pushed a commit with something we hadn't agreed on into his branch, which made it impossible for him to continue with his own solution. The correct way would have been to do your changes in a branch and link to your branch and explain in his PR how your way is different and better.

szmarczak:
> The correct way would have been to do your changes in a branch and link to your branch and explain in his PR how your way is different and better.

I guess he would close it without reading any content, as he told me to do not so before. Then he said on gitter he will follow the way I had been talking before (fastify similar hooks) :) Seriously, no sense.

readme.md
```diff
@@ -124,7 +124,7 @@ Default: `{}`
 
 Request headers.
 
-Existing headers will be overwritten. Headers set to `null` or `undefined` will be omitted.
```
szmarczak: I've removed that part to not confuse people with the behaviour of `got.mergeOptions`, where setting to `undefined` may keep the old value.

sindresorhus:
> hooks are called by reference,

Not sure I see the point of this. Example use-case? I generally prefer not exposing `this` unless absolutely necessary.

> merge() is a different file now,

Can you leave it where it is for now so we get a diff?

> [] instead of new Array in merge().

👍

szmarczak:
> Not sure I see the point of this. Example use-case? I generally prefer not exposing this unless absolutely necessary.

Simple example:

```js
const {beforeRequest} = options.hooks;

const handler = function (options) {
	options.headers['some-data'] = this.data; // this = options.hooks.beforeRequest
};

beforeRequest.data = 'someData';

beforeRequest.push(handler);
```

> Can you leave it where it is for now so we get a diff?

Working on it.

szmarczak:
> I think it should be up to users to bind methods if they want to use this.

Right, I'll revert that :)

test/headers.js
```diff
@@ -142,12 +142,22 @@ test('remove null value headers', async t => {
 	t.false(Reflect.has(headers, 'user-agent'));
 });
 
-test('remove undefined value headers', async t => {
+test('setting a header to undefined keeps the old value', async t => {
 	const {body} = await got(s.url, {
 		headers: {
 			foo: undefined
```
sindresorhus: Shouldn't this be `'user-agent'`?

readme.md
```diff
@@ -403,30 +403,28 @@ client.get('/demo');
 
 *Need more control over the behavior of Got? Check out the [`got.create()`](advanced-creation.md).*
 
-**Both `got.extend(options)` and `got.create(options)` will freeze the instance's default options. For `got.extend()`, the instance's default options are the result of `got.mergeOptions`, which effectively copies plain `Object` and `Array` values. Therefore, you should treat objects passed to these methods as immutable.**
```
sindresorhus: Why was this removed? I guess we could shorten it a little bit, but I like to keep the note about `you should treat objects passed to these methods as immutable.` somewhere.

test/headers.js
szmarczak: Yeah, must missed that when was merging with master. Fixed.

readme.md
szmarczak: I consider it's no longer needed, because the defaults are cloned, so your object is safe, nothing will change :)

PR #541, opened 2018-08-02
Ignore JSON option when using `got.stream()`

szmarczak:


readme.md
```diff
@@ -137,7 +137,7 @@ Returns a `Stream` instead of a `Promise`. This is equivalent to calling `got.st
 
 Type: `string` `Buffer` `stream.Readable` [`form-data` instance](https://github.com/form-data/form-data)
 
-*This is mutually exclusive with stream mode.*
+*If you provide this option, `got.stream()` will be read-only.*
```
sindresorhus: How is this related to the JSON change and what does `read-only` mean here?

szmarczak: AFAIK `form` option is converted to `body` option [here](https://github.com/sindresorhus/got/blob/80a02fec0ee69da1079cc89d94fba0bcf0aff423/source/normalize-arguments.js#L97). `read-only` means the stream is no longer writable.

szmarczak: I've just changed the sentence to make things clearer. The behaviour of `form` option remains the same.

sindresorhus: But you changed the meaning. Previously it said that the option cannot be used with `got.stream()` at all. Now you're saying that it can, but that the stream will be read-only. If this is true, can you point to a test proving that?

szmarczak: > Previously it said that the option cannot be used with got.stream() at all.

Yeah, but practically it wasn't true.

> can you point to a test proving that?

Sure :)

szmarczak: Here you go: https://github.com/sindresorhus/got/blob/b504346f92b7e59942e6fdfd1da264762b300267/test/stream.js#L62-L66

See? I hope it proves it :)

sindresorhus: 👍 Great :)

PR #490, opened 2018-06-02
fix Buffer.byteLength(req._header) throwing error

egorovli:
according to #458, `got` throws error in some cases when trying to pipe one `got.stream` into another one:

```bash
buffer.js:514
    throw new ERR_INVALID_ARG_TYPE(
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be one of type string, Buffer, or ArrayBuffer. Received type object
    at Function.byteLength (buffer.js:514:11)
    at Timeout.setInterval [as _onTimeout] (/Users/egorovli/Dev/project/api/node_modules/got/index.js:207:35)
    at ontimeout (timers.js:427:11)
    at tryOnTimeout (timers.js:289:5)
    at listOnTimeout (timers.js:252:5)
    at Timer.processTimers (timers.js:212:10)
```

this commit should fix this behavior, according to https://github.com/sindresorhus/got/issues/458#issuecomment-392847828

PR #542, opened 2018-08-02
Increase coverage

szmarczak:


szmarczak:
I'll leave this part:

https://github.com/sindresorhus/got/blob/6ba9e68c7f496a69641a8aac51be0781f1ae0eac/source/timed-out.js#L30-L32

this makes us sure `timed-out` isn't called twice or more, although it's obvious we don't do it.

szmarczak:
We also need tests for `electron` support.

szmarczak:
I just wonder if this piece of code is needed:

https://github.com/sindresorhus/got/blob/6ba9e68c7f496a69641a8aac51be0781f1ae0eac/source/get-body-size.js#L31-L33

because I tried passing a `Buffer` as body and this doesn't even get called.

source/normalize-arguments.js
```diff
@@ -165,10 +165,10 @@ module.exports = (url, options, defaults) => {
 
 			if (Reflect.has(error, 'headers') && Reflect.has(error.headers, 'retry-after') && retryAfterStatusCodes.has(error.statusCode)) {
 				let after = Number(error.headers['retry-after']);
-				if (is.number(after)) {
-					after *= 1000;
+				if (is.nan(after)) {
+					after = Date.parse(error.headers['retry-after']) - Date.now();
```
szmarczak: This fixes a bug: at first I thought `is.number` won't return `true` for NaN, but I was wrong.

`Math.max` is not needed BTW.

source/request-as-event-emitter.js
```diff
@@ -142,13 +142,7 @@ module.exports = (options = {}) => {
 
 		if (backoff) {
 			retryCount++;
-			setTimeout(options => {
-				try {
-					get(options);
-				} catch (error2) {
-					emitter.emit('error', error2);
-				}
-			}, backoff, options);
```
sindresorhus: Why was this removed?

sindresorhus:
> because I tried passing a Buffer as body and this doesn't even get called.

You have to pass a stream. The buffer testing is testing a property on the stream.

source/request-as-event-emitter.js
szmarczak: `get()` does not throw any errors directly. It uses `emitter.emit('error', error);`

szmarczak:
@sindresorhus But it's converted to a stream here:

https://github.com/sindresorhus/got/blob/6ba9e68c7f496a69641a8aac51be0781f1ae0eac/source/normalize-arguments.js#L111-L114

source/request-as-event-emitter.js
sindresorhus: Good point. I think we need a try/catch here though: https://github.com/sindresorhus/got/pull/542/files#diff-62bdc57f6f22ae58f495daef16f21f8bR141 In case the `options.gotRetry.retries` function throws.

sindresorhus:
@szmarczak Hmm, then maybe it returns here: https://github.com/sindresorhus/got/blob/6ba9e68c7f496a69641a8aac51be0781f1ae0eac/source/get-body-size.js#L10 Add some `debugger` statements to each `if` and figure out where it returns.

source/request-as-event-emitter.js
sindresorhus: Your answer is not confusing, but you're not answering my question. Too be clear, I think we need to try/catch this line because it could throw: `const backoff = options.gotRetry.retries(++retryTries, error);`

szmarczak:
> Hmm, then maybe it returns here:

Yeah, you're right, checked it. So, is it needed? IMO we can get rid of that.

source/request-as-event-emitter.js
szmarczak: > you're not answering my question.

Sorry, if you're talking about the second one then it hadn't popped up when I was writing the answer. Pardon my confusion.

> I think we need to try/catch this line because it could throw

Good point! Indeed. 🙌

sindresorhus: > I think we need to try/catch this line because it could throw

sindresorhus: >> I think we need to try/catch this line because it could throw
> Good point! Indeed. 🙌

And of course a test ;)

szmarczak: Done! I hope it satisfies you :)

sindresorhus:
> Yeah, you're right, checked it. So, is it needed? IMO we can get rid of that.

See: https://github.com/sindresorhus/got/blob/4d92eb6bef37637c0d85986faae314e17e9cd9bc/source/request-as-event-emitter.js#L163-L166

test/retry.js
```diff
@@ -247,3 +249,13 @@ test('works when defaults.options.retry is not an object', async t => {
...
 });
+
+test('retry function can throw', async t => {
+	await t.throws(got(`${s.url}/413`, {
+		retry: {
+			retries: () => {
+				throw new Error('Simple error');
+			}
+		}
+	}), 'Simple error');
```
sindresorhus: It's a good practice to put the message in a variable and use it in both to make absolutely sure they are the same and stay the same in the future too.

szmarczak: Yeah.

szmarczak:
> This should be done in a separate PR though

So it will be :)

> It's a good practice to put the message in a variable and use it in both to make absolutely sure they are the same and stay the same in the future too.

Done :unicorn:

sindresorhus:
> We also need tests for electron support.

Nah, don't care. If it prevents us from getting 100% test coverage, we can just mock it.

> Note: Missing test for #490 and for:

Do you intend to do that in this PR? Asking so I know whether this is ready to merge or not.

sindresorhus:
Something is failing here: https://ci.appveyor.com/project/sindresorhus/got/build/1.0.245/job/nunr63w0rttwjv0t

szmarczak:
>  If it prevents us from getting 100% test coverage

The point is: not only that keeps us from 100%. But it's possible to achieve 99% :)

> Do you intend to do that in this PR? Asking so I know whether this is ready to merge or not.

No, that'll be a separate PR too. Yes, it's ready to merge now :)

szmarczak:
See? Fixed :)

sindresorhus:
Yup, AppVeyor is annoying. We had so many problems with it over at https://github.com/avajs/ava too...

sindresorhus:
👍 I'm extra happy that doing this actually caught a couple of potential bugs.

PR #543, opened 2018-08-03
Increase coverage

szmarczak:


szmarczak:
It's hard to test #490 because it occurs randomly.

PR #544, opened 2018-08-03
Unify calculating content-length

szmarczak:


source/get-body-size.js
```diff
@@ -32,5 +32,9 @@ module.exports = async options => {
 		return body._buffer.length;
 	}
 
+	if (body && is.number(body.length)) {
```
szmarczak: I think `&& !is.nodeStream(body)` is missing here...

sindresorhus: I think this check should be on line 21 for clarity and be `if (Buffer.isBuffer(body)) {`

PR #546, opened 2018-08-04
Support baseUrl as a Function

szmarczak:


brandon93s:
Why is this needed?

If your URL is dynamic it's not a `baseURL` and you should just call `got` with a FQD each time instead of using a `baseURL`. A base url should be static.

szmarczak:
> If your URL is dynamic it's not a baseURL and you should just call got with a FQD each time instead of using a baseURL. A base url should be static.

Yeah, you're right 😋

PR #547, opened 2018-08-06
Normalize options.method

szmarczak:


source/index.js
```diff
@@ -33,7 +33,6 @@ const defaults = {
 		},
 		cache: false,
 		decompress: true,
-		useElectronNet: false,
```
sindresorhus: Why is this removed?

szmarczak: Because it's not needed. See https://github.com/sindresorhus/got/blob/master/source/request-as-event-emitter.js#L43

sindresorhus: I think it's still useful to declare all options up-front, even if they are `false` by default.

source/create.js
```diff
@@ -37,7 +37,8 @@ const create = defaults => {
 		return defaults.handler(normalizeArguments(url, options, defaults), next);
 	};
 
-	for (const method of defaults.methods) {
+	const methods = defaults.methods.map(method => method.toLowerCase());
```
sindresorhus: Why are you lowercasing them here, but uppercasing them in `normalize-arguments`?

source/index.js
szmarczak: Ok, I'll leave that then.

source/create.js
szmarczak: > Why are you lowercasing them here,

Imagine calling `got.POST` instead of `got.post` :P

> but uppercasing them in normalize-arguments?

To keep things consistent; because `error.method` is uppercased :)

sindresorhus: Nvm. I missed the context. It's late here...

szmarczak: That's ok, answering questions is my job :P It's a bit late here too :)

source/index.js
szmarczak: Done.

PR #550, opened 2018-08-09
Fix throwing errors when using got.stream()

szmarczak:
If normalizing arguments throws and using `got.stream()`, it'd return a rejecting promise. This PR fixes that.

### DIFF
--- a/source/create.js
+++ b/source/create.js
@@ -20,7 +20,11 @@ const create = defaults => {
 			options = mergeOptions(defaults.options, options);
 			return defaults.handler(normalizeArguments(url, options, defaults), next);
 		} catch (error) {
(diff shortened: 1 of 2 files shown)


source/create.js
```diff
@@ -20,7 +20,11 @@ const create = defaults => {
 			options = mergeOptions(defaults.options, options);
 			return defaults.handler(normalizeArguments(url, options, defaults), next);
 		} catch (error) {
-			return Promise.reject(error);
+			if (options.stream) {
+				throw error;
```
szmarczak: We could return throwing EventEmitter here. What do you think?

sindresorhus: No, it's a user-error and should fail immediately.

Issue #557, opened 2018-08-11
Add ability to time requests


gajus:
Add ability to capture timings of every HTTP request state, e.g.

> - `time` - if `true`, the request-response cycle (including all redirects) is timed at millisecond resolution. When set, the following properties are added to the response object:
>   - `elapsedTime` Duration of the entire request/response in milliseconds (*deprecated*).
>   - `responseStartTime` Timestamp when the response began (in Unix Epoch milliseconds) (*deprecated*).
>   - `timingStart` Timestamp of the start of the request (in Unix Epoch milliseconds).
>   - `timings` Contains event timestamps in millisecond resolution relative to `timingStart`. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
>     - `socket` Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_socket) module's `socket` event fires. This happens when the socket is assigned to the request.
>     - `lookup` Relative timestamp when the [`net`](https://nodejs.org/api/net.html#net_event_lookup) module's `lookup` event fires. This happens when the DNS has been resolved.
>     - `connect`: Relative timestamp when the [`net`](https://nodejs.org/api/net.html#net_event_connect) module's `connect` event fires. This happens when the server acknowledges the TCP connection.
>     - `response`: Relative timestamp when the [`http`](https://nodejs.org/api/http.html#http_event_response) module's `response` event fires. This happens when the first bytes are received from the server.
>     - `end`: Relative timestamp when the last bytes of the response are received.
>   - `timingPhases` Contains the durations of each request phase. If there were redirects, the properties reflect the timings of the final request in the redirect chain:
>     - `wait`: Duration of socket initialization (`timings.socket`)
>     - `dns`: Duration of DNS lookup (`timings.lookup` - `timings.socket`)
>     - `tcp`: Duration of TCP connection (`timings.connect` - `timings.socket`)
>     - `firstByte`: Duration of HTTP server response (`timings.response` - `timings.connect`)
>     - `download`: Duration of HTTP download (`timings.end` - `timings.response`)
>     - `total`: Duration entire HTTP round-trip (`timings.end`)

– https://github.com/request/request

szmarczak:
We could use our `timed-out.js` module for that. It already does measurements. Small changes and it should be done.

sindresorhus:
Would be useful. I think @jstewmon wanted something like this too.

jstewmon:
[...]

Issue #559, opened 2018-08-12
Log HTTP client behaviour


gajus:
To begin with, it would be handy to be able to inspect configuration of every HTTP request. I currently do that by wrapping `got` in a helper function that logs whatever the configuration was used to create a request.

Other things that would be useful to log:

* Response meta data (including redirects)
* Whether response was served from cache
* Request cancellation
* Request timeout

If you used something like https://github.com/gajus/roarr, there would be near-0 performance impact when logging is disabled. Logging can be controlled using environment variables.

Happy to contribute an integration.

sindresorhus:
👍 I agree this would be useful. I think `roarr` is a bit heavyweight to include in Got, but how about we just expose a `metadata` event and people can log using whatever they want? We could also log by default using [`util.debuglog()`](https://nodejs.org/api/util.html#util_util_debuglog_section).

gajus:
> I think roarr is a bit heavyweight to include in Got, [..]

I have refactored Roarr into two packages:

* https://github.com/gajus/roarr
* https://github.com/gajus/roarr-cli

Now Roarr logger package is 304K.

Now the bulk of the size comes from `sprintf-js`, which I cannot do much about.

If the intent is to compete with other packages on package size, then this is going to compromise your position.

> but how about we just expose a metadata event and people can log using whatever they want?

This approach works great for applications, but it sucks for modules. I explain the distinction in https://github.com/gajus/roarr#motivation.

tl;dr;

When you have an application which depends on modules that use got, there is no way to enable got logging without digging through node_modules/ and patching the code.

> We could also log by default using util.debuglog().

I like `util.debuglog()`. However, it does not provide structured logs. Modern logging stacks (ELK, Splunk) ingest structured logs (JSON), which allows to implement monitoring/ alerting and even scaling based on log attributes. This is primary reason for using logs in the first place – nowadays debugging is better achieved with `--inspect` anyway.

sindresorhus:
> When you have an application which depends on modules that use got, there is no way to enable got logging without digging through node_modules/ and patching the code.

Environment variable? Alternatively or in addition to, we could have a singleton emitter, that logs for all instances, so you could just `require('got').on('metadata', () => {})` and get logs even when got is used in sub-dependencies.

gajus:
> Environment variable?

Not if you go with the custom event approach. Environment variables can only toggle logging, not configure the hooks.

[...]

PR #561, opened 2018-08-13
events option to facilitate event listener registration

jstewmon:
@sindresorhus I haven't implemented any tests yet. I'd like to get your feedback on the interface / mechanics before proceeding.

Currently, the event context is only accessible from event handlers. It may be more useful if the context is assigned to the request instance returned by a request method, so that it can be accessed outside of an event handler, but I wasn't sure what one might do with it, so I've punted on that for now. Suggestions welcome.

This will facilitate #557. My suggestion for an implementation to address #557 is to create a peer dependency that exports an events configuration that is applied to a got instance (or at request call-site).


jstewmon:
Note: I didn't do anything related to the `retry` event b/c it's essentially a private event. I think it's possible to infer when a retry occurs using the `options` object (since the instance is unique and persistent across request attempts), but it may be more ergonomic to expose the `retry` event.

szmarczak:
I see no sense of this, because you can achieve the same by listening to `instance.on('request', req => { ... })` and `instance.on('response', req => { ... })`.

> My suggestion for an implementation to address #557 is to create a peer dependency that exports an events configuration that is applied to a got instance

IMO there's a better way. I'd implement the timings in  the `timed-out.js` file.

sindresorhus:
I guess this would be more appropriate for solving #559, but even then, I would prefer something simpler, or something simpler in addition to this. People that need to do low-level advanced stuff can already directly listen to the request/response.

sindresorhus:
With this abstraction, users also loose ability to unsubscribe when they're done with an event.

sindresorhus:
If we were to expose something like this, I would prefer just a single firehose event that includes a type attribute to indicate the type of event of the message.

jstewmon:
The set of events is large, but I didn't really consider that to affect the API surface area. Consider that if instead of an object, you called `addListener`, then there's just one new method.

I thought it was worth providing a porcelain interface, so that users would be saved from creating memory leaks if they're using keepAlive and subscribing to socket events.

I also thought it was really convenient to be able to subscribe to an arbitrary event without having to wire up all the intermediate listeners.

> With this abstraction, users also lose ability to unsubscribe when they're done with an event.

I don't know what you mean by this.

> If we were to expose something like this, I would prefer just a single firehose event that includes a type attribute to indicate the type of event of the message.

That's certainly possible, but it will less efficient because we'll have to subscribe to everything and call the user's handler for every event.

sindresorhus:
> users would be saved from creating memory leaks if they're using keepAlive and subscribing to socket events.

> I also thought it was really convenient to be able to subscribe to an arbitrary event without having to wire up all the intermediate listeners.

Those are some good points. I'm warming up to the idea.

Did you consider any other APIs for this? Not saying we should do it differently, just wondering what you considered.

sindresorhus:
>> With this abstraction, users also lose ability to unsubscribe when they're done with an event.

> I don't know what you mean by this.

For example, a user wants to listen to `response.data` events, but only the first 10 events. I don't see any way to unsubscribe to the event after 10 events. Of course there could just be an if-statement that returns early after 10 events, but that means it's doing work when it's not needed.

sindresorhus:
> That's certainly possible, but it will less efficient because we'll have to subscribe to everything and call the user's handler for every event.

I don't imagine anyone would use this in production, but could be useful to get a complete picture of what's happening. You could do `got.on('firehose', console.log);`. Could be an addition to this though.

readme.md
```diff
@@ -287,6 +287,65 @@ See the [AWS section](#aws) for an example.
...
 **Note**: Modifying the `body` is not recommended because the `content-length` header has already been computed and assigned.
 
+###### events
+
+Type: `Object<string, Function[]>`<br>
+Default: `{}`
+
+Events allow listeners to be attached to all events emitted during the request lifecycle. Listeners are subscribed using `on` or `once` according to the known frequency of the event.
+
+The signature for a given listener can be found in the Node.js docs. All listeners will be passed an additional `context` parameter, which is a plain object, that is initialized with an `options` property containing the normalized options for the request.
```
sindresorhus: > The signature for a given listener can be found in the Node.js docs.

Maybe we can link to at least the main events, so it's quicker for people to look it up.

readme.md
```diff
@@ -287,6 +287,65 @@ See the [AWS section](#aws) for an example.
...
+
+Listeners are free to modify the context object. Use Symbol properties to avoid collisions with multiple listeners.
```
sindresorhus: What's the use-case for modifying the context object?

readme.md
```diff
@@ -287,6 +287,65 @@ See the [AWS section](#aws) for an example.
 
 **Note**: Modifying the `body` is not recommended because the `content-length` header has already been computed and assigned.
 
+###### events
+
+Type: `Object<string, Function[]>`<br>
+Default: `{}`
+
+Events allow listeners to be attached to all events emitted during the request lifecycle. Listeners are subscribed using `on` or `once` according to the known frequency of the event.
```
sindresorhus: Can you place everything below here in a `<details>` tag so it won't take up so much space in the readme?

readme.md
sindresorhus: > Use Symbol properties to avoid collisions with multiple listeners.

Can you include an example of this? Not clear how that would look like.

PR #564, opened 2018-08-15
Follow redirects with encoded URI

### LINKED ISSUE #563: Got 9.0.0 throws when using unescaped characters
With Got 9.0.0,
```
require("got")("http://google.com/search", { query: { q: "friend’s house" } })
   .then(() => console.log("ok"))
```
rejects with:
```
(node:27956) UnhandledPromiseRejectionWarning: RequestError: Request path contains unescaped characters
    at EventEmitter.cacheReq.on.error (node_modules/got/source/request-as-event-emitter.js:105:27)
    at EventEmitter.emit (events.js:182:13)
    at makeRequest (node_modules/cacheable-request/src/index.js:125:9)
    at get (node_modules/cacheable-request/src/index.js:135:14)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

With `got@8.3.2` this snippet works just fine.

jstewmon:
Fixes #563

### DIFF
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -74,7 +74,7 @@ module.exports = options => {
 				redirectUrl = (new URLGlobal(bufferString, urlLib.format(options))).toString();
 
 				try {
-					redirectUrl = decodeURI(redirectUrl);
+					decodeURI(redirectUrl);
 				} catch (error) {
 					emitter.emit('error', error);
 					return;
@@ -89,7 +89,8 @@ module.exports = options => {
 
 				emitter.emit('redirect', response, redirectOpts);
 
-				get(redirectOpts);
+				// Recursive Deferral
+				setImmediate(get, redirectOpts);
 				return;
 			}
 
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -89,8 +89,7 @@ module.exports = options => {
 
 				emitter.emit('redirect', response, redirectOpts);
 
-				// Recursive Deferral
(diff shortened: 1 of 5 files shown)


source/request-as-event-emitter.js
```diff
@@ -74,7 +74,7 @@ module.exports = options => {
 				redirectUrl = (new URLGlobal(bufferString, urlLib.format(options))).toString();
```
jstewmon: I think using `urlLib.format` to get the baseUrl has the same problem as was discussed in https://github.com/sindresorhus/got/pull/519#issuecomment-404950951.

I can take a look at adding a test and fixing today.

jstewmon: The baseUrl won’t have the right query string, but I don’t think there’s any value for new url for which it would affect the result. So this seems fine.

szmarczak:
> redirects are handled with a recursive call to `get`, so deferred
  recursion with setImmediate

See my comment above.

> test server enhancement:

Could this be placed in a separate PR please?

sindresorhus:
> > redirects are handled with a recursive call to get, so deferred
recursion with setImmediate

> See my comment above.

I've reverted this: https://github.com/sindresorhus/got/pull/564/commits/731c470de8cc246b0043400304e33d45c7706027

> > test server enhancement:

> Could this be placed in a separate PR please?

I agree this should have been a separate change, but we decided we can live with it as we want to get out a new minor release today, and it's just nitpick.

sindresorhus:
Thanks for fixing this, @jstewmon 🙌

PR #565, opened 2018-08-15
Allow the `query` option to be a URLSearchParams instance

### LINKED ISSUE #554: Allow `query` option to be a `URLSearchParams` instance
For better WHATWG URL compatibility.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

`URLSearchParams` supports a string too, so we could just always run it through `URLSearchParams` no matter what, for consistency and code simplicity:
https://github.com/sindresorhus/got/blob/f447378763ba623f9b6642b0bf58dba09295b688/source/normalize-arguments.js#L72-L74

lorenzofox3:
Fixes #554

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -183,7 +183,7 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` `[URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)`<br>
+Type: `string` `Object` or [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)`<br>
 
 Query string object that will be added to the request URL. This will override the query string in `url`.
 
--- a/readme.md
+++ b/readme.md
@@ -183,7 +183,7 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` or [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)`<br>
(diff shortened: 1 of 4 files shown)


sindresorhus:
@lorenzofox3 This causes a test failure.

lorenzofox3:
@sindresorhus 
I think the test is failing is because the URLSearchParams implements properly the [standard url encoder parser](https://url.spec.whatwg.org/#concept-urlencoded-parser) and [serializer](https://url.spec.whatwg.org/#concept-urlencoded-serializer) which means `?foo` is correctly parsed and serialized to `?foo=` (which dose not match the redirect url `?foo` of the test server). I see then different solutions.

1. Modify the test and consider `?foo` is equivalent `?foo=` on the test server side (which is the case according to the spec aforementioned if I understand correctly)
2. Change the implementation to have our own toString so that empty values are ignored and the `=` sign is not appended. However this could lead to inconsistent behavior. 
query parameters with following parameters may not output the same queryString:
`?foo`
`?foo=''`
`?foo=''`
`{query:undefined}`
`{query:''}`

On a side note: I am not sure whether it is a problem with the test framework or the reporter but when running the tests I man not able to say it is actually failing: when I run the tests the final output is 
screen shot 2018-08-15 at 11 25 49 pm

szmarczak:
@sindresorhus It's because `(new URLSearchParams('blah')).toString()` doesn't return `blah` but `blah=`. So it ends up with `?blah=` instead of `?blah`, which means the same.

This:

https://github.com/sindresorhus/got/blob/8f30f1fdefaa96ad55758a352730d00697862952/test/redirects.js#L85

should be

```js
 http.on('/relativeQuery?bang=', (request, response) => { 
```

and the tests would pass. I think we should use [lukeed/polka](https://github.com/lukeed/polka) because it's very lightweight and blazing fast. I highly recommend it :)

PR #571, opened 2018-08-18
Fix Retry List

MarkHerhold:
The README states that 500 errors are retried but the code does not have this error code. Fix by removing from README.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -209,7 +209,7 @@ Type: `number` `Object`<br>
 Default:
 - retries: `2`
 - methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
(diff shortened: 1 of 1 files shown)


szmarczak:
It does.

https://github.com/sindresorhus/got/blob/ca2675e3e5c9db97015e827ce652493a8f9cc1f3/source/index.js#L17-L24

MarkHerhold:
Ah, my apologies. I was staring at the README from master while using got v9.0.0.

szmarczak:
Sure, good point. Many people find this confusing.

@sindresorhus Maybe we should point users to a last release branch by default?

sindresorhus:
@szmarczak Nah. Most projects on GitHub use master as the default branch and development branch. For the latest version, users should visit the npm package or the readme in the relevant Git tag. This is expected knowledge.

MarkHerhold:
I agree with @sindresorhus. Just made a novice mistake. Thanks!

PR #583, opened 2018-08-23
Add ability to pass a custom request function

szmarczak:
This would allow an experimental support of http2 using [`http2-wrapper`](https://github.com/szmarczak/http2-wrapper):

```js
const got = require('got');
const wrapper = require('http2-wrapper');

const h2got = got.extend({fn: wrapper});

(async () => {
	const {body} = await h2got('https://nghttp2.org/httpbin/headers');
	console.log(body);
})();
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -266,6 +266,13 @@ Default: `false`
 
 [Cache adapter instance](#cache-adapters) for storing cached data.
 
+###### request
+
+Type: `Function`<br>
+Default: `http.request` `https.request`
+
+Custom request function. By default it's `http.request` or `https.request` (depending on the protocol).
+
 ###### useElectronNet
 
 Type: `boolean`<br>
--- a/readme.md
+++ b/readme.md
@@ -814,12 +814,27 @@ const custom = got.extend({
 
 *Need to merge some instances into a single one? Check out [`got.mergeInstances()`](advanced-creation.md#merging-instances).*
 
+### Experimental HTTP2 support
+
+Got provides an experimental support for HTTP2 using the [`http2-wrapper`](https://github.com/szmarczak/http2-wrapper) module:
+
+```js
+const got = require('got');
+const {request} = require('http2-wrapper');
+
+const h2got = got.extend({request});
+
+(async () => {
+	const {body} = await h2got('https://nghttp2.org/httpbin/headers');
+	console.log(body);
+})();
+```
 
 ## Comparison
 
 |                       |  `got`  | `request` | `node-fetch` | `axios` |
 |-----------------------|:-------:|:---------:|:------------:|:-------:|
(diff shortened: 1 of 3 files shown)


source/request-as-event-emitter.js
```diff
@@ -32,7 +32,12 @@ module.exports = options => {
 			return;
 		}
 
-		let fn = options.protocol === 'https:' ? https : http;
+		let fn;
+		if (options.fn) {
```
sindresorhus: Let's name it `options._requestFunction` so it's clear that it's readable and clear that not something we officially support.

source/request-as-event-emitter.js
```diff
@@ -32,7 +32,12 @@ module.exports = options => {
...
+			fn = options.fn; // eslint-disable-line prefer-destructuring
```
sindresorhus: `({fn} = options);`

szmarczak: Isn't it ugly? :P

sindresorhus: Yes... Ok, don't do it.

sindresorhus: Fixed: https://github.com/xojs/eslint-config-xo/commit/20daf6d7798140b05e690dfab63959e7c21bfed9 You can remove the disable comment now.

sindresorhus:
https://github.com/sindresorhus/got/pull/583#discussion_r212383890

source/request-as-event-emitter.js
```diff
@@ -32,7 +32,12 @@ module.exports = options => {
 			return;
 		}
 
-		let fn = options.protocol === 'https:' ? https : http;
+		let fn;
+		if (options._requestFunction) {
+			fn = options.fn; // eslint-disable-line prefer-destructuring
```
sindresorhus: You need to update this one too.

sindresorhus:
I think we should be explicit an accept the `request` function, not a library with a request function.

readme.md
```diff
@@ -814,12 +821,27 @@ const custom = got.extend({
 
 *Need to merge some instances into a single one? Check out [`got.mergeInstances()`](advanced-creation.md#merging-instances).*
 
+### Experimental HTTP2 support
+
+Got provides an experimental support for HTTP2 using the [`http2-wrapper`](https://github.com/szmarczak/http2-wrapper) module:
```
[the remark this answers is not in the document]
szmarczak: I decided to name it `module` because it's used everywhere in the readme. Should I replace all `module`s with `package`?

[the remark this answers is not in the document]
sindresorhus: The distinction is that a module is what you import, while a package is what you publish to npm. A package can contain multiple modules. I'll fix it in the readme after merge.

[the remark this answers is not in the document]
szmarczak: Good point.

PR #590, opened 2018-08-25
Provide timings

### LINKED ISSUE #557: Add ability to time requests

szmarczak:
Fixes #557

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
 	],
 	"dependencies": {
 		"@sindresorhus/is": "^0.11.0",
+		"@szmarczak/http-timer": "^1.0.2",
 		"cacheable-request": "^4.0.1",
 		"decompress-response": "^3.3.0",
 		"duplexer3": "^0.1.4",
--- a/readme.md
+++ b/readme.md
@@ -89,11 +89,7 @@ It's a `GET` request by default, but can be changed by using different methods o
 
 #### got(url, [options])
 
-Returns a Promise for a `response` object with a `body` property, a `url` property with the request URL or the final URL after redirects, and a `requestUrl` property with the original request URL.
-
-The response object will typically be a [Node.js HTTP response stream](https://nodejs.org/api/http.html#http_class_http_incomingmessage), however, if returned from the cache it will be a [response-like object](https://github.com/lukechilds/responselike) which behaves in the same way.
-
-The response will also have a `fromCache` property set with a boolean value.
+Returns a Promise for a [`response` object](#response) or a [stream](#streams-1) if `options.stream` is set to true.
 
 ##### url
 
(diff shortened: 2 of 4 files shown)


sindresorhus:
Awesome! Can you add some docs?

szmarczak:
> How accurate is the start time? It's set automatically when timer is invoked. How close is this to when the request truly starts?

I think 0~1ms

> Should this be opt-in? The overhead seems pretty minimal, so maybe keeping it simple and always running is OK.

I agree.

sindresorhus:
> Sure, but please check out #592 first :)

Done

szmarczak:
IDK why but after `git pull --rebase && git add . && git commit -m 'message'` some commits got duplicated (from 7eeb352 to a4d0c47)

PR #591, opened 2018-08-25
Add tests for stripping port in `host` header

### LINKED ISSUE #588: Strip port suffix from Host header if the protocol is known
See https://github.com/request/request/pull/2904

szmarczak:
Fixes #588

### DIFF
--- a/test/headers.js
+++ b/test/headers.js
@@ -178,3 +178,28 @@ test('non-existent headers set to undefined are omitted', async t => {
 	const headers = JSON.parse(body);
 	t.false(Reflect.has(headers, 'blah'));
 });
+
+test('preserve port in host header if non-standard port', async t => {
+	const {body} = await got(s.url, {json: true});
+	t.is(body.host, 'localhost:' + s.port);
+});
+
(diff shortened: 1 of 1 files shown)


test/headers.js
```diff
@@ -178,3 +178,28 @@ test('non-existent headers set to undefined are omitted', async t => {
...
 	t.false(Reflect.has(headers, 'blah'));
 });
+
+test('preserve port in host header if non-standard port', async t => {
+	const {body} = await got(s.url, {json: true});
+	t.is(body.host, 'localhost:' + s.port);
+});
+
+test('strip port in host header if explicit standard port (:80) & protocol (HTTP)', async t => {
+	const {body} = await got('http://httpbin.org:80/headers', {json: true});
```
sindresorhus: I don't think we should use remote server when we can just use our local server. Using a remote endpoint slows down the tests and makes them more fragile.

sindresorhus: Right... We could have used `nock`, but I guess this is good enough then.

szmarczak: Wait, I forgot about `nock`. I'll just push a commit.

PR #594, opened 2018-08-28
Proper fix for #469

szmarczak:


sindresorhus:
Can you explain why this is a more proper fix?

szmarczak:
Because `onSocketConnect` doesn't get called if the socket has been destroyed. Initiating an interval and removing  it after 150ms is redutant (but it's a nice workaround if we don't know the cause, however it's not the case now). https://github.com/sindresorhus/got/pull/469#issuecomment-416316745

PR #595, opened 2018-08-29
Add migration guides

szmarczak:


sindresorhus:
I think each migration should be a separate file.

szmarczak:
This PR is not my first priority right now, so it may take a while to finish this.

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+	}
+})();
+```
+
+Now it looks cool, huh?
+
+#### Options in common
+
+Both `request` and `got` accept [`http.request` options](https://nodejs.org/api/http.html#http_http_request_options_callback).
+Did you know that Got has options which are the same in `request`?
```
szmarczak: Maybe `Got has some options in common with `request`?

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+Oh, and one more thing... There's no `time` options. Assume it's always true.
+
+#### Renamed options
+
+Readability is very important to us, so we decided to rename these options:
+
+- `qs` → `query`
+- `strictSSL` → `rejectUnauthorized`
+- `gzip` → `decompress`
+- `jar` → `cookieJar` (accepts `tough-cookie` jar)
```
szmarczak: need to link `tough-cookie`

migration-guides.md
sindresorhus: I'm not a big fan of the "asking questions you'll answer yourself" thing. I would prefer just saying it:

> These Got options are the same in with Request:

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+Now it looks cool, huh?
+
+#### Options in common
+
+Both `request` and `got` accept [`http.request` options](https://nodejs.org/api/http.html#http_http_request_options_callback).
+Did you know that Got has options which are the same in `request`?
+
+- `url`
+- `body`
+- `baseUrl`
```
sindresorhus: Are you sure this has the same behavior?

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+Now it looks cool, huh?
+
+#### Options in common
+
+Both `request` and `got` accept [`http.request` options](https://nodejs.org/api/http.html#http_http_request_options_callback).
+Did you know that Got has options which are the same in `request`?
+
+- `url`
```
sindresorhus: This has different behavior. We support `URL` too.

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+
+> :star: Switching from other HTTP request libraries to Got :star:
+
+### Migrating from Request
+
+You may think, it's too hard to switch. But that's not true. Let's take the very first example from `request` and see how it looks using `got`.
+
+```js
+const request = require('request');
+request('http://www.google.com', function (error, response, body) {
```
sindresorhus:
```suggestion
request('https://google.com', (error, response, body) => {
```

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+  console.log('error:', error); // Print the error if one occurred
+  console.log('statusCode:', response && response.statusCode); // Print the response status code if a response was received
+  console.log('body:', body); // Print the HTML for the Google homepage.
+});
+
+const got = require('got');
```
sindresorhus: No point in having both an example using promises and one using async/await. Just drop this on. We should always use async/await.

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+
+The `timeout` option has some extra features. You can [set timeouts on particular events!](readme.md#timeout).
+
+The `query` option is always serialized using [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) unless it's a `string`.
+
+To use streams, just call `got.stream(url, options)` or `got(url, {stream: true, ...}`).
+
+#### Breaking changes
+
+- no `form` option. You have to pass a [`form-data` instance](https://github.com/form-data/form-data) through the `body` option
```
sindresorhus: `no` => `No`

migration-guides.md
```diff
@@ -0,0 +1,131 @@
...
+- no `jsonReviver`/`jsonReplacer` option (but you can you hooks for that too)
+
+Hooks are powerful, aren't they? [Read more](readme.md#hooks) to know what else you can use the hooks for!
+
+#### More about streams
+
+Let's take a quick look on another example from `request`:
+
+```js
+http.createServer(function (req, res) {
```
sindresorhus:
```suggestion
http.createServer((request, response) => {
```

migration-guides.md
szmarczak: Right. I'll change that.

migration-guides.md
szmarczak: That'll be just a note. The behavior is the same, but we accept additionally `URL`.

migration-guides.md
szmarczak: Ok :fire:

migration-guides.md
```diff
@@ -0,0 +1,128 @@
...
+You may think it's too hard to switch, but it's really not. 🦄
+
+Let's take the very first example from Request's readme:
+
+```js
+const request = require('request');
+
+request('https://google.com', (error, response, body) => {
+	console.log('error:', error); // Print the error if one occurred
+	console.log('statusCode:', response && response.statusCode); // Print the response status code if a response was received
```
sindresorhus: I know this is from the Requests docs, but this makes no sense. Why would you guard `response`, but not `body`?...

migration-guides.md
```diff
@@ -0,0 +1,128 @@
...
+
+To use streams, just call `got.stream(url, options)` or `got(url, {stream: true, ...}`).
+
+#### Breaking changes
+
+- No `form` option. You have to pass a [`form-data` instance](https://github.com/form-data/form-data) through the [`body` option](https://github.com/sindresorhus/got#body).
+- No `oauth`/`hawk`/`aws`/`httpSignature` option. To sign requests, you need to create a [custom instance](advanced-creation.md#signing-requests).
+- No `agentClass`/`agentOptions`/`forever`/`pool` option.
+- No proxy option. You need to [pass a custom agent](readme.md#proxies).
+- No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
```
sindresorhus: Also no `maxRedirects` option. There are many more too that are not listed here.

migration-guides.md
```diff
@@ -0,0 +1,128 @@
...
+				delete options.headers.referer;
+			}
+		]
+	}
+});
+
+gotInstance(url, options);
+```
+
+- No `jsonReviver`/`jsonReplacer` option, but you can use hooks for that too.
```
sindresorhus: Can you show an example of this too? I know this option is fairly common, so would be nice to give people an easy transition.

migration-guides.md
```diff
@@ -0,0 +1,128 @@
...
+
+The [`baseUrl` option](https://github.com/sindresorhus/got#baseurl) appends the ending slash if it's not present.
+
+To use streams, just call `got.stream(url, options)` or `got(url, {stream: true, ...}`).
+
+#### Breaking changes
+
+- No `form` option. You have to pass a [`form-data` instance](https://github.com/form-data/form-data) through the [`body` option](https://github.com/sindresorhus/got#body).
+- No `oauth`/`hawk`/`aws`/`httpSignature` option. To sign requests, you need to create a [custom instance](advanced-creation.md#signing-requests).
+- No `agentClass`/`agentOptions`/`forever`/`pool` option.
```
sindresorhus: Can we provide any transition steps / alternatives for these options?

sindresorhus:
Can you link to this at the top of the readme with a link text like:

> Moving from Request?

migration-guides.md
szmarczak: There is much sense! :D If `response` is `null`/`undefined` you can refer to `undefinded` by `response.statusCode`. It throws. But you can log `undefined`, right?

szmarczak: Try:

```js
const a = undefined;
console.log(a.b);
console.log(a);
```

szmarczak: AFAIK if there's an error the rest is `null`.

migration-guides.md
sindresorhus: > I prefer to operate on parents, not children (options)

What does that mean?

migration-guides.md
sindresorhus: You're right. I'm too used to Got. It's just the Request API that makes no sense for me in general.

migration-guides.md
szmarczak:
```js
agent: new agentClass(agentOptions)
```

[forever-agent](https://github.com/request/forever-agent)


pool implementation:
```js
const agentSymbol = Symbol('agent');
const pools = new Map();

const getKeyFromPool = pool => ...;

const getAgent = (pool, options) => {
	const merged = {...pool, ...options};
	const key = getKeyFromPool(merged);

	if (!pools.has(key)) {
		pools.set(key, new specifiedAgentByUserOrDefault(pool));
	}

	return pools.get(key);
};
```

szmarczak: will do :)

migration-guides.md
szmarczak: > followRedirect - follow HTTP 3xx responses as redirects (default: true). This property can also be implemented as function which gets response object as a single argument and should return true if redirects should continue or false otherwise.
> followAllRedirects - follow non-GET HTTP 3xx responses as redirects (default: false)
> followOriginalHttpMethod - by default we redirect to HTTP method GET. you can enable this property to redirect to the original HTTP method (default: false)

It's counted as only one option: `followRedirect`.

szmarczak: Some Request's options are not spec-compliant.

sindresorhus: > It's counted as only one option: followRedirect. (followAllRedirects and followOriginalHttpMethod are the children of followRedirect, 'ya know what I mean?)

Yes, but is that clear enough to the end-user though? Why not be explicit about it? Some might just do Cmd+F to try to find what they're looking for. I think we should list them all to make it easy to the user to discover.

> Also, some Request's options are not spec-compliant.

Spec-compliant to what? HTTP? And does it matter? Then we list it and say that Got does not support that because it's not spec compliant.

migration-guides.md
```diff
@@ -83,37 +92,37 @@ const instance = got.extend({
...
 
-instance(url, options);
+gotInstance(url, options);
 ```
 
-- no `jsonReviver`/`jsonReplacer` option (but you can use hooks for that too)
+- No `jsonReviver`/`jsonReplacer` option, but you can use hooks for that too.
 
-Hooks are powerful, aren't they? [Read more](readme.md#hooks) to know what else you can use the hooks for!
+Hooks are powerful, aren't they? [Read more](readme.md#hooks) to see what else you can do with hooks.
```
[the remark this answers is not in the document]
sindresorhus: > do with does not mean use.

Not specifically, but in this context it's IMHO clear.

> Or: see what else you can achieve using hooks.

I'm ok with this one too though. 👍

migration-guides.md
szmarczak: I did **not** say we should have no such feature because it isn't spec-compliant. I meant Got's more spec-compliant than Request. Anyway, nevermind. It doesn't matter. I'll just list them.

sindresorhus: I got (pun intended) what you meant. I just think we should list it no matter what.

migration-guides.md
```diff
@@ -0,0 +1,170 @@
...
+					options.body = newBody;
+
+					// #2 solution (slower)
+					options.body = JSON.parse(JSON.stringify(options.body, options.jsonReplacer));
+				}
+			}
+		],
+		afterResponse: [
+			response => {
+				// TODO in Got: we need to make the `options` public somehow
```
sindresorhus: @szmarczak Can you open a new issue about this?

sindresorhus:
Looking good :)

Still some minor things left, but these things can be done in a follow-up PR:

- [ ] https://github.com/sindresorhus/got/pull/595#discussion_r230579249
- [ ] https://github.com/sindresorhus/got/pull/595#discussion_r230174879
- [ ] https://github.com/sindresorhus/got/pull/595#discussion_r238534446

PR #598, opened 2018-08-30
Fix Electron throwing `HTTP trailers are not supported` error

### LINKED ISSUE #461: Using useElectronNet results in error: HTTP trailers are not supported
I'm using this code to upload a file to S3:

```
const response = await got.put(urlString, {
      body: fs.readFileSync(filename),
      useElectronNet: true,
    });
``` 
And this throws an uncaughtException "HTTP trailers are not supported"

This because the response is parsed by mimic-response:  `mimicResponse(res, progressStream);`
and mimic-response uses a `knownProps` array to read values from the `IncomingMessage`.
`knownProps` props contains 'trailers' and 'rawTrailers'. When these are read from the `IncomingMessage`, this throws an error because of this code:
https://github.com/electron/electron/blob/1c0ea0286ed7f8fb9075d1a6c03c758c72e965b8/lib/browser/api/net.js#L59

mimic-response should be made aware of that it is using electron.net or got should do something to prevent the reading of these properties.

If you let me know what you'd prefer, I can make a pull request 😄

szmarczak:
Fixes #461

### DIFF
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -54,6 +54,7 @@ module.exports = options => {
 		let timings;
 		const cacheableRequest = new CacheableRequest(fn.request, options.cache);
 		const cacheReq = cacheableRequest(options, response => {
+			// Fixes https://github.com/electron/electron/blob/master/lib/browser/api/net.js#L59
 			if (options.useElectronNet) {
 				response = new Proxy(response, {
 					get: (target, name) => {
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -1,6 +1,4 @@
 'use strict';
-/* istanbul ignore next: webpack only */
-const r = ({x: require})['yx'.slice(1)];
 const {URL} = require('url'); // TODO: Use the `URL` global when targeting Node.js 10
 const util = require('util');
 const EventEmitter = require('events');
@@ -54,6 +52,7 @@ module.exports = options => {
 
 		/* istanbul ignore next: electron.net is broken */
 		if (options.useElectronNet && process.versions.electron) {
+			const r = ({x: require})['yx'.slice(1)]; // Trick webpack
 			const electron = r('electron');
 			fn = electron.net || electron.remote.net;
 		}
--- a/source/request-as-event-emitter.js
+++ b/source/request-as-event-emitter.js
@@ -1,4 +1,6 @@
 'use strict';
+/* istanbul ignore next: webpack only */
+const r = ({x: require})['yx'.slice(1)];
 /* istanbul ignore next: compatibility reason */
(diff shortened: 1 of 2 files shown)


source/request-as-event-emitter.js
```diff
@@ -47,13 +47,26 @@ module.exports = options => {
...
+					get: (target, name) => {
+						if (name === 'trailers' || name === 'rawTrailers') {
+							return [];
+						}
+
+						const value = target[name];
+						return is.function(value) ? value.bind(target) : value;
+					}
+				});
+			}
```
sindresorhus: Why not just this?

```js
if (options.useElectronNet) {
	response.trailers = [];
	response.rawTrailers = [];
}
```

source/request-as-event-emitter.js
```diff
@@ -47,13 +47,26 @@ module.exports = options => {
 
 		/* istanbul ignore next: electron.net is broken */
 		if (options.useElectronNet && process.versions.electron) {
-			const electron = global['require']('electron'); // eslint-disable-line dot-notation
+			const electron = require('electron');
```
sindresorhus: It was like this for a reason. It's to prevent Webpack from complaining. I guess we should add a comment about that.

szmarczak: IDK why but the "global" approach was throwing an error for me. I'll try that again.

source/request-as-event-emitter.js
sindresorhus: I see. You could use `Object.defineProperty`, but I guess this is safer.

source/request-as-event-emitter.js
```diff
@@ -47,13 +47,26 @@ module.exports = options => {
...
 			fn = electron.net || electron.remote.net;
 		}
 
 		let timings;
 		const cacheableRequest = new CacheableRequest(fn.request, options.cache);
 		const cacheReq = cacheableRequest(options, response => {
+			if (options.useElectronNet) {
```
sindresorhus: Can you add a comment with a link to the Electron issue this fixes?

szmarczak: Yeah, I'll do.

source/request-as-event-emitter.js
szmarczak: This is what I get:
```js
(node:8174) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: global.require is not a function
```

szmarczak: Hmm... Smart :P

source/request-as-event-emitter.js
```diff
@@ -1,4 +1,6 @@
 'use strict';
+/* istanbul ignore next: webpack only */
+const r = ({x: require})['yx'.slice(1)];
```
sindresorhus: This is only used in one place, so just put it inline with the function call at line 50.

sindresorhus:

Issue #606, opened 2018-09-08
`got.extend()` doesn't merge hooks


dmitry-vakhnenko:
I am try to get options dict for modifing it later. But looks like hook never been called.

```JavaScript
'use strict'
const got = require('got')

const client = got
  .extend({
    baseUrl: 'https://jsonplaceholder.typicode.com/todos/1',
    json: true,
    hooks: {
      beforeRequest: [(options) => console.log(111111111111)],
    },
  })
  .get('/')
  .then(console.log)
  .catch(console.error)
```

got version: **9.2.1**
node version: **v10.10.0**

dmitry-vakhnenko:
hm, if i pass hook to `.get('/')` function then it works.

```JavaScript
const got = require('got')

const noop = () => {}

const client = got
  .extend({
    baseUrl: 'https://jsonplaceholder.typicode.com/todos/1',
    json: true,
    hooks: {
      beforeRequest: [(options) => console.log(1111111111111111111111111111111111111111111111111111111)],
    },
  })
  .get('/', {
    hooks: {
      beforeRequest: [(options) => console.log(111111111111111111)],
    },
  })
  .then(noop)
  .catch(noop)
```

szmarczak:
Can you elaborate where's the problem?

dmitry-vakhnenko:
How i can define global hooks?

when i extend handler i got "compiled" options, where query already attached to url.



So my solution is wrong. It just replaces hooks(as described [here](https://github.com/sindresorhus/got#gotmergeoptionsparentoptions-newoptions]))

szmarczak:
What do you mean by "global hooks"?

dmitry-vakhnenko:
> What do you mean by "global hooks"?

Hooks that works for all children of instance. With `extend` we can add hooks. We can also manage hooks inside hooks. Now, when i use `extend` hooks are just replaces.

> Can you elaborate where's the problem?

```JavaScript
'use strict'

const got = require('got')

const noop = () => {}

const hooks = {
  beforeRequest: [console.log],
}

const client = got.create({
  options: {
    baseUrl: 'https://google.com',
    hooks,
  },
})

// hooks here, OK!
console.log(client.defaults.options.hooks)

// but it not works, hook not called
client
  .post('/')
  .then(noop)
  .catch(noop)

// when i pass directly hooks will be called
client
  .post('/', { hooks })
  .then(noop)
  .catch(noop)
```

szmarczak:
[...]

PR #607, opened 2018-09-08
#606 fix hooks merging inside create

dmitry-vakhnenko:
#606

### DIFF
--- a/source/create.js
+++ b/source/create.js
@@ -51,8 +51,8 @@ const create = defaults => {
 	got.stream = (url, options) => got(url, {...options, stream: true});
 
 	for (const method of aliases) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

dmitry-vakhnenko:
Never write tests before. So, i try.

szmarczak:
This is nonsense.

`got` calls `normalizeArguments`, where options are merged.

https://github.com/sindresorhus/got/blob/7ae6939f920d1174b181063ce61056a8a6def4bc/source/create.js#L33

https://github.com/sindresorhus/got/blob/7ae6939f920d1174b181063ce61056a8a6def4bc/source/normalize-arguments.js#L58

szmarczak:
See #608

PR #615, opened 2018-09-16
Fix unhandled stream error in pipe

### LINKED ISSUE #614: Unhandled stream error in pipe caused by ECONNRESET
When I use `form-data` to upload file, in some complicate environments, the code will crash. Below error message will display.

```
internal/streams/legacy.js:59
      throw er; // Unhandled stream error in pipe.
      ^
Error: read ECONNRESET
    at TCP.onread (net.js:622:25)
```

conechan:
Fix #614

### DIFF
--- a/source/as-promise.js
+++ b/source/as-promise.js
@@ -35,6 +35,7 @@ module.exports = options => {
 
 			if (is.nodeStream(options.body)) {
 				options.body.once('end', uploadComplete);
+				request.on('error', error => reject(error));
 				options.body.pipe(request);
 				options.body = undefined;
 				return;


source/as-promise.js
```diff
@@ -35,6 +35,7 @@ module.exports = options => {
 
 			if (is.nodeStream(options.body)) {
 				options.body.once('end', uploadComplete);
+				request.on('error', error => reject(error));
```
szmarczak: This should be before that `if`.

sindresorhus:
@conechan Ping :)

conechan:
Sorry, I've tried to write a test to reproduce it but failed. I will try again when I have spare time. Thanks.

szmarczak:
I'll try to fix this :)

PR #616, opened 2018-09-16
fixed cancellation

0xflotus:


szmarczak:
Both forms are correct. @sindresorhus has discussed it somewhere and prefers to use "cancelation".

PR #622, opened 2018-09-26
Include `body` property in `HTTPError`

mrmlnc:
#### Description

The ability to look at the body of the error response inside the `retries` function .

#### Why?

Some services use the response body to provide an extended description of why the error occurred (for example, reason of [409 code](https://developer.mozilla.org/docs/Web/HTTP/Status/409)).

We would like to use this field for logging.

### DIFF
--- a/source/errors.js
+++ b/source/errors.js
@@ -59,9 +59,9 @@ module.exports.ParseError = class extends GotError {
 };
 
 module.exports.HTTPError = class extends GotError {
-	constructor(request, opts) {
-		const {statusCode} = request;
-		let {statusMessage} = request;
+	constructor(response, opts) {
+		const {statusCode} = response;
+		let {statusMessage} = response;
 
 		if (statusMessage) {
 			statusMessage = statusMessage.replace(/\r?\n/g, ' ').trim();
@@ -72,8 +72,8 @@ module.exports.HTTPError = class extends GotError {
 		this.name = 'HTTPError';
 		this.statusCode = statusCode;
 		this.statusMessage = statusMessage;
-		this.headers = request.headers;
-		this.body = request.body;
+		this.headers = response.headers;
+		this.body = response.body;
 	}
 };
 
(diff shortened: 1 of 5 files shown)


mrmlnc:
Maybe it makes sense to also pass `request` as field of the error?

szmarczak:
@sindresorhus Any thoughts?

source/errors.js
```diff
@@ -59,7 +59,10 @@ module.exports.ParseError = class extends GotError {
 };
 
 module.exports.HTTPError = class extends GotError {
-	constructor(statusCode, statusMessage, headers, opts) {
+	constructor(request, opts) {
```
brandon93s: Shouldn't parameter 1 be `response` instead of `request`?

szmarczak: changed

sindresorhus:
Makes sense.

The docs needs to be updated too: https://github.com/sindresorhus/got#gothttperror

szmarczak:
> The docs needs to be updated too: [sindresorhus/got#gothttperror](https://github.com/sindresorhus/got#gothttperror)

done

nuragic:
Hello! Any chance to bump a patch release for this one today? 😌 Thanks!

razor-x:
@sindresorhus I am adopting Got for our project to replace `request`. So far it looks great, but I need this (and `timings`) in a new version before I can push stable stuff on my side (can't use the git version for internal reasons). 

Been using the git version to develop with no issues, but :smile_cat: Can haz has release  :hamburger:  or ETA plz? :pray:

szmarczak:
@razor-x https://github.com/sindresorhus/got/milestone/7 also see #635

Rowno:
Doesn't this introduce a security issue for people logging errors? Since the body could easily contain sensitive data.

sindresorhus:
```js
class FooError extends Error {
	constructor() {
		super();
		this.unicorn = 'not logged';
    }
}

console.log(new FooError());
// Error
//  at <anonymous>:8:13
```

Rowno:
I mean logging errors with packages like winston or sentry, which also log all the enumerable properties on the Error object.

PR #627, opened 2018-10-05
Fix `tunnel` package reference in the docs

Kikobeats:


sindresorhus:
Those are not the same packages. Please include a description in the PR of why this change is needed. You also included an unnecessary newline.

Kikobeats:
sorry, just because I suppose you are referencing this library in the proxy section as well:

https://github.com/sindresorhus/got#proxies

PR #633, opened 2018-10-15
Don't override headers defined in the `url` argument when it's an object

### LINKED ISSUE #632: Default headers override custom headers from `url` argument
#### Description 
Here we override custom headers (user input aka `...url`) with a default set of headers (aka `...options`) when `url` argument is a object.

https://github.com/sindresorhus/got/blob/8d2e91171509bc2aec71c4483978cfd421ebcf1d/source/normalize-arguments.js#L100-L105

#### Minimal reproducible example of bug

```js
const got = require('got');

const client = got.extend({
    hooks: {
        beforeRequest: [
            async (options) => {
                console.dir('Hook (beforeRequest)', { colors: true });
                console.dir(options.headers, { colors: true });
            }
        ]
    }
});

(async () => {
    try {
        const res = await client.get({
            protocol: 'https:',
            host: 'google.com',
            path: '/',
            headers: {
                'X-Custom-Header': 'some-value'
            }
        });

        console.dir('Requests headers', { colors: true });
        console.dir(res.req.getHeaders(), { colors: true });
    } catch (error) {
        console.dir(error, { colors: true });
    }
})();
```

Output with some `console.log`:

```js
'Before options merging (url.headers)'
{ 'X-Custom-Header': 'some-value' }

'Before options merging (options.headers)'
{ 'user-agent': 'got/9.2.2 (https://github.com/sindresorhus/got)' }

'After options merging'
{ 'user-agent': 'got/9.2.2 (https://github.com/sindresorhus/got)' }

'Hook (beforeRequest)'
{ 'user-agent': 'got/9.2.2 (https://github.com/sindresorhus/got)',
  'accept-encoding': 'gzip, deflate' }

'Requests headers'
{ 'user-agent': 'got/9.2.2 (https://github.com/sindresorhus/got)',
  'accept-encoding': 'gzip, deflate',
  host: 'www.google.com' }
```

mrmlnc:
Fixes #632


szmarczak:
I've even made a PR to not confuse `url` with `options`... https://github.com/sindresorhus/got/pull/514 Maybe it'll be supported, but not for now :)

mrmlnc:
Are you trying to tell me that a `url` argument can only be a `string` or `WHATWG  URL`? May be in this case it is necessary to write about it more explicitly in the documentation? Like, we only support the definition of `protocol`, `host`/`hostname`, and `path` … or we support all properties exclude `headers` and ….

I'm sorry, but it's not very obvious when the documentation explicitly says that a `url` argument can be a string, or a [`https.request` options object](https://nodejs.org/api/https.html#https_https_request_options_callback), or a [WHATWG `URL`](https://nodejs.org/api/url.html#url_class_url).

sindresorhus:
@szmarczak I think you're confused about what this PR is about. Yes, we disallow `got({url: '…'})`, but we *do* document support for `got({host: '…', headers: {…}})`. The fact that this is no longer supported is a regression.

> The URL to request, as a string, a https.request options object, or a WHATWG URL.

The `https.request options object` can contain a `header` field.

szmarczak:
@sindresorhus We disallow the `url` option [no longer](https://github.com/sindresorhus/got/commit/af341ca44c414607b01c3a3eaa066fa643ab23a4#diff-1d0f03b7418c902cb7ec7637a8cd8520R89).

szmarczak:
Yes, then we can reopen this PR :tada:

> Maybe it'll be supported, but not for now :)

The time is now.

PR #635, opened 2018-10-16
Code cleanup, bug fixes

### LINKED ISSUE #630: Usage without caching
So I'm looking to use `got` without any caching, which seems to be the default with `{ cache: false }`. My code calls `got()` hundreds of thousands of times per minute, so performance is critical.

Every call to `got()` currently creates a new instance of `CacheableRequest`, which is really a factory that gets called exactly once to produce a decorator around the raw request. Internally, that in turn creates a new instance of `Keyv` which then (by default) goes on to create a `Map` it's not going to use either. So am I misunderstanding something or does the default behavior allocate all these objects for nothing? Using the current API, I think the best I can do is pass a `NullCache` that doesn't allocate the `Map` over and over, but everything else is not configurable.

It also seems like the `CacheableRequest` factory is meant to be reused rather than creating a new one for every request. Would you accept a PR that memoizes the call so that every `{ request, cache }` tuple uses the same factory?

Thanks!

### LINKED ISSUE #631: Intermittent uncaught errors

```
```

szmarczak:
- [x] Code cleanup
   - [x] Moved the download progress function into `progress.js`
   - [x] DRY (`as-promise.js` and `as-stream.js`)
   - [x] The retry function is no longer an event, it's just a function.
   - [x] Used more `async/await` in `request-as-event-emitter.js` to catch errors
   - [x] Doesn't use `CacheableRequest` when there's no need
   - [x] Uses `urlToOptions` instead of `urlLib.parse` in `request-as-event-emitter.js`
   - [x] Moved the cancelling mechanism into `request-as-event-emitter.js`
   - [x] Modules which should be separate libs have been moved to `utils` folder
   - [x] Renamed `options.gotRetry` (internal) to `options.retry`
   - [x] DRY (`request-as-event-emitter.js` and `normalize-arguments.js` - validating UTF8 URL)
 - [x] [Destroying `got.stream(...)` cancels the request](https://github.com/sindresorhus/got/compare/master...szmarczak:unhandled-fix#diff-6defa9239fb128d7e40c26c0f9373009R25)
 - [x] *Fixed unhandled `ParseError`s (metioned by @aghuddleston in #631)
 - [x] [Fixed unhandled `TimeoutError`s](https://github.com/sindresorhus/got/compare/master...szmarczak:unhandled-fix#diff-675cd4a9c0a14591082c92ae8638b3ccR51) (metioned by @timdp in #631)
 - [x] [Accepts `url` as an option](https://github.com/sindresorhus/got/pull/635/files#diff-1d0f03b7418c902cb7ec7637a8cd8520R89)
 - [x] Optimized normalizing arguments
 - [x] [Fixed default protocol regression (before: `http:` now: `https:`)](https://github.com/sindresorhus/got/pull/635/files#diff-1d0f03b7418c902cb7ec7637a8cd8520R123)
 - [ ] ~~Caches `CacheableRequest` instances~~ (will be done on CacheableRequest's side)

\* Hard to reproduce. Could be [continuing the request after cookie errors](https://github.com/sindresorhus/got/pull/635/files#diff-893b4bab3e6efc693e1e3480b6306005R106), parsing the response after aborting (fixed by adding `request.once('error', () => {})` before `request.abort()`), cache failure...

Fixes #631
Fixes #630


szmarczak:
I've got a few questions:

1. Why is `Transform` used instead of `response.on('data', ...)` for download progress?
2. Is there any better way to proxy errors?
3. We must optimize `normalize-arguments.js` some time (it's very slow).
4. Is it a good idea to let users change the defaults of any instance? Example:

```js
instance.defaults.options.headers.unicorn = 'rainbow';
```

timdp:
> Don't use `CacheableRequest` when there's no need

😍

source/request-as-event-emitter.js
```diff
@@ -188,10 +169,48 @@ module.exports = options => {
...
+			} else if (input && (options.method === 'POST' || options.method === 'PUT' || options.method === 'PATCH')) {
+				input.once('end', uploadComplete);
+				input.pipe(request);
+			} else {
+				request.end(uploadComplete);
+			}
+		};
+
+		if (options.cache) {
+			const cacheableRequest = new CacheableRequest(fn.request, options.cache); // TODO: we need a manager for CacheableRequest instances
```
timdp: Am I missing something or would this work?

```js
const mem = require('mem')

// memoized factory
const createCacheableRequest = mem((request, cache) => new CacheableRequest(request, cache))

if (options.cache) {
  const cacheReq = createCacheableRequest(fn.request, options.cache)(options, handleResponse)
  // ... do stuff ...
}
```

szmarczak: Seems good to me. I'll try that. Thank you ;)

szmarczak: I tried it, and unfortunately it doesn't work...

```js
z = require('mem')((x, y) => Math.random())
[Function]
> a = new Map()
Map {}
> b = new Map()
Map {}
> z(1, a)
0.7761714954957284
> z(1, b)
0.7761714954957284
```

Although `a` and `b` aren't the same objects, it gives the same results. It causes the tests fail.

szmarczak: @sindresorhus is it a bug? ^^^^^

timdp: Yeah, I just discovered that it JSON-stringifies the args. Darn. I suppose you could put a symbol on the cache to uniquely identify it.

szmarczak: Yeah, did already so! :)

szmarczak:
2.6x performance gain:

```
got x 4,719 ops/sec ±3.01% (79 runs sampled)
```

source/as-stream.js
```diff
@@ -12,41 +11,18 @@ module.exports = options => {
...
-		if (options.method === 'POST' || options.method === 'PUT' || options.method === 'PATCH') {
-			input.once('end', uploadComplete);
-			input.pipe(request);
-			return;
-		}
-
-		request.end(uploadComplete);
-	});
+	// Cancels the request
+	proxy._destroy = emitter.abort;
```
sindresorhus: Test for this?

source/index.js
```diff
@@ -24,10 +24,9 @@ const defaults = {
 				504
 			]
 		},
-		cache: false,
 		decompress: true,
-		useElectronNet: false,
```
sindresorhus: Not clear why you removed the defaults for `useElectronNet` and `cache`?

source/normalize-arguments.js
```diff
@@ -29,45 +23,62 @@ const preNormalize = options => {
...
+		options.gotTimeout = options.timeout;
+	}
+	delete options.timeout;
+
+	const {retry} = options;
+	if (defaults && retry !== false) {
+		options.retry = {...defaults.options.retry};
 	} else {
-		throw new TypeError(`Parameter \`hooks\` must be an object, not ${is(options.hooks)}`);
+		options.retry = {retries: 0, methods: [], statusCodes: []};
```
sindresorhus: Put the properties on separate lines.

source/request-as-event-emitter.js
```diff
@@ -6,38 +6,56 @@ const http = require('http');
...
+	if (!cachedCacheableRequests.has(request)) {
+		cachedCacheableRequests.set(request, new Map());
+	}
+
+	if (!cachedCacheableRequests.get(request).has(cache)) {
+		cachedCacheableRequests.get(request).set(cache, new CacheableRequest(request, cache));
+	}
+
+	return cachedCacheableRequests.get(request).get(cache);
+};
```
sindresorhus: Won't this cache all requests forever and cause a memory leak?

szmarczak:
Sure, will do! :)

source/index.js
szmarczak: It's not needed. There are many options which are `false` by default and those aren't here :) IMO we should only set defaults which point to `true`, it's more clear. Others are `false` implicitly.

source/request-as-event-emitter.js
szmarczak: I thought about it too. But this'd require to generate new cache per request.

Perhaps we could limit the size of the cache to 100 items?

source/index.js
sindresorhus: I think it's better to define all defaults upfront? What if the check is `foo == false`, then it will never match, since the default is not there, and the user would only pass in `true`.

source/request-as-event-emitter.js
szmarczak: > Why would it cache the requests?

What do you mean?

> I agree that memoization could be a feature of the package itself or one around it though.

+1

timdp: Cacheable-request exports a class (for some reason) that actually just gives you the cacheable factory method bound to the actual request factory (`request` in your code, which is a function, right?) and the cache. Calling the return value of `new CacheableRequest` is what actually produces the request, and there's no caching going on at that level. Hence, all you're doing is caching a function and some state, not the request itself. Therefore, v8 should GC it just fine. Correct me if I'm wrong.

szmarczak:
@github there's a bug: this PR won't update though I made some commits... that's the second time I've met this situation.

szmarczak:
If someone is interested to help with caching CacheableRequest instances, feel free: https://github.com/lukechilds/cacheable-request/compare/master...szmarczak:manager

sindresorhus:
Looks good. Nice work on the refactor. 🙌

timdp:
Indeed, very nice work! :+1:

For the caching of cacheable-request, how about introducing the lru-cache package? You'd have to do some magic to map arguments to keys though. (I wanted to mention this on a PR but it doesn't exist yet.)

sindresorhus:
@timdp Not really interested in adding an extra dependency just for this. And if we were, would probably be better to use https://github.com/sindresorhus/quick-lru, as it can use anything as keys, including objects.

timdp:
What you want is an LRU cache with multi-valued keys though, so a mix of LRU and multimap. I'm sure that exists somewhere.

PR #641, opened 2018-10-24
Add readme `query` note

### LINKED ISSUE #638: Arrays in query properties
## Problem

The documentation says that [`query`](https://github.com/sindresorhus/got#query) property can be represented as an object.

If the property is an array, the request path is built incorrectly.

```js
const res = await got('https://site.com', {
  query: {
    some: ['a', 'b']
  }
});
```

## Current behaviour

```
request => https://site.com/?some=a%2Cb
```

## Expected behaviour

```
request => https://site.com/?some=a&some=b
```

## Problem research

In the `got@8` we use `querystring`.

https://github.com/sindresorhus/got/blob/ad7b361dcb2490c3864b845b979b756f13f7d89b/index.js#L552

In the `got@9` we use `URLSearchParams`.

https://github.com/sindresorhus/got/blob/8d2e91171509bc2aec71c4483978cfd421ebcf1d/source/normalize-arguments.js#L118

```js
// Current behaviour
new URL.URLSearchParams({ some: ['a', 'b'] }).toString() // -> some=a%2Cb

// Correct way
new URL.URLSearchParams([["some", 'a'],["some", 'b']]).toString() // -> some=a&some=b
```

If I understand the description correctly, I cannot pass an object to this constructor:

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams

szmarczak:
Fixes #638

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -203,11 +203,27 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+Type: `string` `Array` `Object` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
 
 Query string object that will be added to the request URL. This will override the query string in `url`.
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is not what I proposed in https://github.com/sindresorhus/got/issues/638#issuecomment-432597514.

szmarczak:
What about now?

szmarczak:
now?

PR #643, opened 2018-10-26
Improve docs for the `query` option

### LINKED ISSUE #638: Arrays in query properties

sindresorhus:
Fixes #638
Closes #641

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -207,7 +207,7 @@ Type: `string` `Object<string, string|number>` [`URLSearchParams`](https://devel
 
 Query string that will be added to the request URL. This will override the query string in `url`.
 
-If you need to pass in an array, you can do it with a `URLSearchParams` instance:
+If you need to pass in an array, you can do it using a `URLSearchParams` instance:
 
 ```js
 const got = require('got');
--- a/readme.md
+++ b/readme.md
@@ -203,9 +203,36 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+Type: `string` `Object<string, string>` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
 
-Query string object that will be added to the request URL. This will override the query string in `url`.
+Query string that will be added to the request URL. This will override the query string in `url`.
+
+If you need to pass in an array, you can do it with a `URLSearchParams` instance:
+
+```js
+const got = require('got');
+
+const query = new URLSearchParams([['key', 'a'], ['key', 'b']]);
+
+got('https://example.com', {query});
+
+console.log(query.toString());
+//=> 'key=a&key=b'
+```
+
+And if you need a different array format, you could use the [`query-string`](https://github.com/sindresorhus/query-string) package:
+
+```js
+const got = require('got');
+const queryString = require('query-string');
+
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -203,9 +203,36 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
...
+Query string that will be added to the request URL. This will override the query string in `url`.
+
+If you need to pass in an array, you can do it with a `URLSearchParams` instance:
+
+```js
+const got = require('got');
+
+const query = new URLSearchParams([['key', 'a'], ['key', 'b']]);
+
+got('https://example.com', {query});
```
szmarczak: you don't have to create a new `URLSearchParams` instance... it'll be done in `normalize-arguments.js`.

```js
got('https://example.com', {query: [['key', 'a'], ['key', 'b']]});
```

sindresorhus: That's exactly what I want to remove in the next major version. Then it will be; either use an object with string values or specify a `URLSearchParams` instance.

sindresorhus: So either `query: {foo: '1'}` or `new URLSearchParams({foo: 1})`

szmarczak: okay :D

readme.md
```diff
@@ -203,9 +203,36 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+Type: `string` `Object<string, string>` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
 
-Query string object that will be added to the request URL. This will override the query string in `url`.
+Query string that will be added to the request URL. This will override the query string in `url`.
+
+If you need to pass in an array, you can do it with a `URLSearchParams` instance:
```
szmarczak: `with` => `using`

readme.md
```diff
@@ -203,9 +203,36 @@ Parse response body with `JSON.parse` and set `accept` header to `application/js
 
 ###### query
 
-Type: `string` `Object` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+Type: `string` `Object<string, string>` [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
```
szmarczak: `Object<string, string/number>`

PR #644, opened 2018-10-26
Mention `node-fetch` Node.js stream API in comparison table

### LINKED ISSUE #637: Comparison with node-fetch
I think there are some inaccuracies:

- Stream API: node-fetch has `res.body.pipe`.
- Custom defaults: node-fetch does merge `Headers` and `Request` options correctly per spec.

Not butthurt of anything, just bump into this table and think the comparison can be improved.

sindresorhus:
Fixes #637

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -986,7 +986,7 @@ const h2got = got.extend({request});
 | Browser support       |       ✖      |       ✖      |       ✔*     |       ✔      |
 | Electron support      |       ✔      |       ✖      |       ✖      |       ✖      |
 | Promise API           |       ✔      |       ✔      |       ✔      |       ✔      |
-| Stream API            |       ✔      |       ✔      |       ✖      |       ✖      |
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -986,7 +986,7 @@ const h2got = got.extend({request});
 | Browser support       |       ✖      |       ✖      |       ✔*     |       ✔      |
 | Electron support      |       ✔      |       ✖      |       ✖      |       ✖      |
 | Promise API           |       ✔      |       ✔      |       ✔      |       ✔      |
-| Stream API            |       ✔      |       ✔      |       ✖      |       ✖      |
+| Stream API            |       ✔      |       ✔      | Node.js only |       ✖      |
```
szmarczak: maybe `Node-specific`?

sindresorhus: I went with something that is consistent with the existing "Browser only".

PR #647, opened 2018-10-28
Change the way to retry using the afterResponse hook

szmarczak:


source/as-promise.js
```diff
@@ -27,28 +28,34 @@ module.exports = options => {
...
 						}
-						return;
-					}
+
+						return asPromise(mergeOptions(options, {
+							...updatedOptions,
+							retry: {
+								retries: () => 0
+							},
+							throwHttpErrors: false
```
sindresorhus: Why this? I think it's surprising behavior that it suddenly doesn't throw on errors.

source/as-promise.js
```diff
@@ -27,28 +28,34 @@ module.exports = options => {
...
+							throw new Error('Retry limit reached.');
 						}
-						return;
-					}
+
+						return asPromise(mergeOptions(options, {
+							...updatedOptions,
+							retry: {
+								retries: () => 0
+							},
```
sindresorhus: Why not just:

```js
retry: {
	retries: 0
},
```

source/as-promise.js
```diff
@@ -27,28 +28,34 @@ module.exports = options => {
...
 					// eslint-disable-next-line no-await-in-loop
-					response = await hook(response);
-
-					if (is.plainObject(response)) {
-						if (emitter.retry(response) === false) {
-							reject(new Error('Retry limit reached.'));
+					response = await hook(response, updatedOptions => {
+						if (emitter.retry(null) === false) {
+							throw new Error('Retry limit reached.');
 						}
```
sindresorhus: I don't fully understand this code, but this request should not be affected by the retry limit.

source/request-as-event-emitter.js
```diff
@@ -226,16 +226,6 @@ module.exports = (options, input) => {
...
-		if (is.plainObject(error)) {
-			// Retry with updated options
-			options = {
-				...options,
-				...error
-			};
-
-			error = null;
-		}
-
```
sindresorhus: Why is this being removed?

source/as-promise.js
szmarczak: These options are normalized. `retry.retries` is a function.

source/request-as-event-emitter.js
szmarczak: This code was used for retrying when `afterResponse` returned a plain object as the updated options.

source/as-promise.js
sindresorhus: But doesn't `mergeOptions` normalize? I'm just asking, because if it doesn't normalize the `retry` there, it doesn't normalize `updatedOptions` either, which is problematic.

source/as-promise.js
szmarczak: Then it'd be possible to have infinite loop. Assume it always returns 401, no matter what. Then it keeps retrying, and retrying... Keep in mind that the hooks are still present.

sindresorhus: So this only let it run the hook once?

source/as-promise.js
szmarczak: This is used only to get the response object (not an error). The original `asPromise` determines if it should throw or not.

test/hooks.js
```diff
@@ -259,14 +270,35 @@ test('throws on afterResponse when reached retry limit', async t => {
...
+								token: 'unicorn'
+							}
+						});
+					}
+
+					return response;
+				}
+			]
+		}
+	}), {instanceOf: got.HTTPError});
```
sindresorhus: Test the error message too. A lot of things could accidentally be HTTP errors.

source/as-promise.js
szmarczak: No, the hook is ran anyway, no matter if is it a normal request or a retry. But if it fails anyway, it loops. And to prevent the infinite loop we need to check the retries.

source/as-promise.js
szmarczak: yup, I forgot about it, sorry :p

readme.md
```diff
@@ -391,7 +391,7 @@ const got = require('got');
 const instance = got.extend({
```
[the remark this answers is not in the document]
szmarczak: forgot about it. fixin

source/as-promise.js
szmarczak: https://github.com/sindresorhus/got/pull/647/commits/98e51810797c8c93cc0b2fe55f5d0e6d2e7eb687#diff-0920de5c123084fabef750aa0967f8d4R309

source/as-promise.js
szmarczak: here's a test: https://github.com/sindresorhus/got/pull/647/commits/98e51810797c8c93cc0b2fe55f5d0e6d2e7eb687#diff-0920de5c123084fabef750aa0967f8d4R309

also you may set it to `true` and you'll see it fails

source/as-promise.js
sindresorhus: I think this behavior needs to be documented.

szmarczak: Or we can stop executing next hooks in the retry, then we won't need `emitter.retry(null) == false`. Example:

hook 1
hook 2 - retries
	hook 1
hook 3

Current behavior (very bad, it shouldn't be done like that):
hook 1
hook 2 - retries
	hook 1
	hook 2
	hook 3
hook3

source/as-promise.js
szmarczak: nah, you confused me! :P `mergeOptions` doesn't normalize. It just merges them.

source/as-promise.js
szmarczak: > Alright. I think this behavior needs to be documented though.

No need to have that behavior. Fixed! There's no infinity loop. See my previous comment :)

source/as-promise.js
sindresorhus: But the problem is still there. The users options needs to be normalized as if they called `got(options)`. That's what the user would expect.

readme.md
```diff
@@ -383,7 +383,9 @@ got('example.com', {
 Type: `Function[]`<br>
 Default: `[]`
 
-Called with [response object](#response). Each function should return the response or updated options. This is especially useful when you want to refresh an access token. Example:
+Called with [response object](#response) and a retry function.
+
+Each function should return the response or updated options. This is especially useful when you want to refresh an access token. Example:
```
sindresorhus: > return the response or updated options

They don't return the updated options.

source/as-promise.js
```diff
@@ -33,20 +33,19 @@ const asPromise = options => {
...
+						updatedOptions = mergeOptions(options, {
 							...updatedOptions,
 							retry: {
 								retries: () => 0
 							},
 							throwHttpErrors: false
-						}));
+						});
+
+						updatedOptions.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
```
sindresorhus: You are removing this hook and all hooks before it. Is that the intended behavior? Would be nice with a code comment then to explain the behavior.

szmarczak: Will do! :)

source/as-promise.js
szmarczak: Will be :)

PR #652, opened 2018-11-04
No new timeouts after an error

### LINKED ISSUE #631: Intermittent uncaught errors

szmarczak:
Fixes #631

### DIFF
--- a/source/utils/timed-out.js
+++ b/source/utils/timed-out.js
@@ -12,31 +12,7 @@ class TimeoutError extends Error {
 
 const reentry = Symbol('reentry');
 
-function addTimeout(delay, callback, ...args) {
-	// Event loop order is timers, poll, immediates.
-	// The timed event may emit during the current tick poll phase, so
-	// defer calling the handler until the poll phase completes.
-	let immediate;
-	const timeout = setTimeout(() => {
-		immediate = setImmediate(callback, delay, ...args);
-		/* istanbul ignore next: added in node v9.7.0 */
-		if (immediate.unref) {
-			immediate.unref();
-		}
-	}, delay);
-
-	/* istanbul ignore next: in order to support electron renderer */
-	if (timeout.unref) {
-		timeout.unref();
-	}
-
-	const cancel = () => {
-		clearTimeout(timeout);
-		clearImmediate(immediate);
-	};
-
-	return cancel;
-}
+const noop = () => {};
 
 module.exports = (request, delays, options) => {
 	/* istanbul ignore next: this makes sure timed-out isn't called twice */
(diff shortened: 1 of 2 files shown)


sindresorhus:
@timdp Can you try this out?

szmarczak:
> request.once('error', () => {}); // Ignore the `socket hung up` error made by request.abort()

Hmm... Is it a good idea to ignore all `socket hung up` errors? If users do `promise.cancel()`, they won't see them, unless they do `promise.on('request', request => setTimeout(() => request.abort(), n))`

timdp:
@szmarczak I ran

```bash
yarn add 'szmarczak/got#fix-timeouts'
```

and the test from #631 worked as expected:

```
caught: { RequestError: getaddrinfo ENOTFOUND 1541424862699.dev 1541424862699.dev:80
```

**However,** it noticeably takes a while to execute. I added `console.time()` and consistently got execution times of just over two seconds. In the readme, it says that the time between retries is

```
1000 * Math.pow(2, retry) + Math.random() * 100
```

which, with `retry = 0`, means that the maximum execution time should be 1,100 ms plus a few milliseconds for the timeout, right? I don't get where the extra second goes.

UltCombo:
@timdp I can confirm there's a bug in the default retry delay logic.

The [`retries`](https://github.com/sindresorhus/got/blob/50fdab303cb2a6b34383de13e5e0ece9ceaf80c9/source/normalize-arguments.js#L208) function gets `1` for the iteration argument in the first retry (see [here](https://github.com/sindresorhus/got/blob/50fdab303cb2a6b34383de13e5e0ece9ceaf80c9/source/request-as-event-emitter.js#L230)), so the [retry delay math](https://github.com/sindresorhus/got/blob/50fdab303cb2a6b34383de13e5e0ece9ceaf80c9/source/normalize-arguments.js#L239) is wrong (`1 << 1 === 2`).

timdp:
Also, I want to point out that the concept of `cancelers` is basically what RxJS used to call [`Disposable`](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/disposables/disposable.md) until version 4. We still have some code that depends on it at our company. Just another argument to support what I [originally said](https://github.com/sindresorhus/got/issues/630#issuecomment-428758237)  about async state, even if it's not very pragmatic. :-)

timdp:
Yeah, my original reproduction of the issue used `retry: 1`, so I kept that and discovered a second issue.

szmarczak:
Yup. It's a bug. From the docs:

> Delays between retries counts with function 1000 * Math.pow(2, retry) + Math.random() * 100, where retry is attempt number (starts from 0).

Let's merge this PR then. I'll make another one for that to keep things organized :)

Issue #639, opened 2018-10-24
Got v10 brainstorming


sindresorhus:
I would prefer not doing a breaking release for a long time, but that doesn't mean we can't brainstorm.

Feel free to suggest breaking changes here. Suggestions should include use-cases and why it's worth doing a breaking change.

- [x] When the `query` string is an object, but not `URLSearchParams`, we should enforce that all the keys are `string|number`. To avoid issues like https://github.com/sindresorhus/got/issues/638.
- [x] Rename the `query` option to `searchParams`.
- [x] Use `undefined` instead of `null` whenever possible.

sindresorhus:
@szmarczak What do you think about renaming the `query` option to `searchParams` to align with Ky and WHATWG URL terms. Would be more familiar for people coming from the browser.

szmarczak:
100% yes!

szmarczak:
options.retry.retries function should be called with the retry argument starting
from 1 instead of 0.

UltCombo:
> options.retry.retries function should be called with the retry argument starting
> from 1 instead of 0.

Looks like that's the current (undocumented) behavior.

szmarczak:
@UltCombo Yeah, you're right 😅 Sometimes I just think about one thing and do the second instead. Just a mistake. Edited :D

javierblancosp:
What about when the request body is a nodejs stream (A multipart request), options.json can not be true. In the case i want to parse the response body as json i have to do it manually on a multipart request.I know is not a big deal but it could be nice to allow this somehow 🏋️

szmarczak:
@javierblancosp why?

javierblancosp:
@szmarczak mm i think if i expect a response to be a json the request body should not modify this, in may way of thinking :P

szmarczak:
@javierblancosp Sorry for late response. Well, it's not a good idea. It has been discussed here: #174 #318 #383 #467 #511 #600 

Use this instead:

```js
const instance = got.extend({
    hooks: {
        beforeRequest: [
            options => {
                options.body = JSON.stringify(options.body);
            }
        ]
});

instance(url, options);
```

sindresorhus:
@szmarczak Would it be hard to support both `query` and `searchParams` at the same time in v9? That would be optimal as we could then change the docs to `searchParams` so new users use the correct option naming and we could then slowly phase it out. Maybe just log a warning in v10 and then remove it in v11.

szmarczak:
[...]

PR #653, opened 2018-11-06
Fix the retry math

szmarczak:
Fixes a bug where the retry delay is calculated incorrectly.

@timdp @UltCombo

### DIFF
--- a/source/normalize-arguments.js
+++ b/source/normalize-arguments.js
@@ -206,7 +206,7 @@ const normalize = (url, options, defaults) => {
 		const {retries} = options.retry;
 
 		options.retry.retries = (iteration, error) => {
-			if (iteration >= retries) {
+			if (iteration > retries) {
 				return 0;
 			}
 
@@ -236,7 +236,7 @@ const normalize = (url, options, defaults) => {
 			}
 
 			const noise = Math.random() * 100;
-			return (2 ** iteration) * 1000 + noise;
+			return (2 ** (iteration - 1)) * 1000 + noise;
 		};
 	}
 
--- a/source/normalize-arguments.js
+++ b/source/normalize-arguments.js
@@ -206,7 +206,7 @@ const normalize = (url, options, defaults) => {
 		const {retries} = options.retry;
 
 		options.retry.retries = (iteration, error) => {
-			if (iteration > retries) {
+			if (iteration >= retries) {
 				return 0;
 			}
 
@@ -236,7 +236,7 @@ const normalize = (url, options, defaults) => {
 			}
 
 			const noise = Math.random() * 100;
-			return ((1 << iteration) * 1000) + noise;
+			return (2 ** iteration) * 1000 + noise;
 		};
 	}
 
--- a/source/normalize-arguments.js
+++ b/source/normalize-arguments.js
@@ -236,7 +236,7 @@ const normalize = (url, options, defaults) => {
 			}
 
 			const noise = Math.random() * 100;
-			return ((2 ** (iteration - 1)) * 1000 ) + noise;
+			return ((2 ** (iteration - 1)) * 1000) + noise;
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


source/request-as-event-emitter.js
```diff
@@ -227,7 +226,7 @@ module.exports = (options, input) => {
 		let backoff;
 
 		try {
-			backoff = options.retry.retries(++retryTries, error);
+			backoff = options.retry.retries(retryCount, error);
```
FabricioMatteMassive: This is a breaking change:

The client can provide a custom `retries` function and it expects to receive `1` for the `iteration` argument in the first retry. This breaks the contract, it's a SEMVER-MAJOR change. I'd prefer to keep the old behavior.

UltCombo:
Oh I've submitted the review with my work account, oops. :sweat_smile:

source/normalize-arguments.js
```diff
@@ -236,7 +236,7 @@ const normalize = (url, options, defaults) => {
 			}
 
 			const noise = Math.random() * 100;
-			return ((1 << iteration) * 1000) + noise;
+			return ((2 ** iteration) * 1000) + noise;
```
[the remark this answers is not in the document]
sindresorhus: Yes, https://github.com/sindresorhus/got/blob/50fdab303cb2a6b34383de13e5e0ece9ceaf80c9/package.json#L9

sindresorhus:
@szmarczak Can you include a description in the PR body and add a test?

source/request-as-event-emitter.js
sindresorhus: > The client can provide a custom retries function and it expects to receive 1 for the iteration argument in the first retry.

@szmarczak We need a test for this.

sindresorhus: Yes, we need to preserve backwards compatibility. We can add it to https://github.com/sindresorhus/got/issues/639 if there's any benefit to it.

szmarczak: > The client can provide a custom retries function and it expects to receive 1 for the iteration argument in the first retry. This breaks the contract, it's a SEMVER-MAJOR change. I'd prefer to keep the old behavior.

Actually, no. It's a bug. According to the docs, it should start with 0:

> Delays between retries counts with function 1000 * Math.pow(2, retry) + Math.random() * 100, where retry is attempt number (starts from 0).

szmarczak: @sindresorhus 

> we need to preserve backwards compatibility

Compatibility with the docs or compatibility with the previous mechanism?

UltCombo: @szmarczak I see it as explaining the default logic, not the actual arguments. I'm using a custom `retries` function so it's a breaking change.

The documentation should also be updated to document what arguments the `retries` function gets.

szmarczak: In case we agree to save the previous mechanism, then yes.

@sindresorhus any thoughts?

sindresorhus: > @sindresorhus any thoughts?

Yes, like I said, backwards-compatibility must be preserved.

szmarczak: Right. Then it starts from 1.

sindresorhus: @szmarczak But if you think it should be changed, we can do so in Got 10. Add a comment https://github.com/sindresorhus/got/issues/639 if so.

UltCombo: I generally prefer 0-indexed collections, however in this specific case I believe it makes sense that `iteration === 1` in the first retry attempt. Having retry attempt as 0 is not intuitive.

Having the first retry attempt as 1 allows writing semantic code like:

```js
got('url', {
  retry: {
    // Retry up to 2 times.
    retries: retry => retry <= 2 ? getRetryDelay() : 0,
  },
});
```

@sindresorhus @szmarczak

PR #659, opened 2018-11-13
Clear `socket` timeout on error

### LINKED ISSUE #631: Intermittent uncaught errors

szmarczak:
Fixes #631

### DIFF
--- a/test/timeout.js
+++ b/test/timeout.js
@@ -470,5 +470,6 @@ test('socket timeout is canceled on error', async t => {
 	});
 
 	await t.throwsAsync(promise, {message});
+	// Wait a bit more to check if there are any unhandled errors
 	await delay(10);
 });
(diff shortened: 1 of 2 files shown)


test/timeout.js
```diff
@@ -458,3 +458,17 @@ test('no more timeouts after an error', async t => {
...
+
+	const promise = got(s.url, {
+		timeout: {socket: requestTimeout},
+		retry: 0
+	}).on('request', request => {
+		request.emit('error', new Error(message));
+	});
+
+	await t.throwsAsync(promise, {message});
+	await delay(10);
```
brandon93s: Consider copying the note here as well. Wasn't immediately clear why we were waiting. 

> // Wait a bit more to check if there are any unhandled errors

PR #660, opened 2018-11-13
Add ability to specify which network error codes to retry on

### LINKED ISSUE #654: Option to not retry on timeout errors
I'm configuring requests for AWS lambda environment, and case is that:
- I set timeout option to ensure it ends before allowed lambda invocation time ends (so I handle eventual timeout gently by rejecting lambda invocation with meaningful error), and automated retry after such timeout occurs is unwanted.
- I'd like to have retries applied for case of temporary network or server unavailability errors.

Is there an option to achieve that?

szmarczak:
Fixes #654


test/retry.js
```diff
@@ -151,6 +152,37 @@ test('custom retries', async t => {
 	t.true(tried);
 });
 
+test.only('custom errors', async t => {
```
brandon93s: Remove `.only`

advanced-creation.md
```diff
@@ -82,6 +82,16 @@ const defaults = {
 				502,
 				503,
 				504
+			],
+			errors: [
```
sindresorhus: I think we should name it `errorCodes`, since they're from `error.code`.

sindresorhus: Same in the other places you use "error".

szmarczak: I agree.

PR #663, opened 2018-11-14
Add Got options onto responses and errors

michaltk:
It would be helpful to have the options set on `response.req` to identify how the request was made.
For ex, in the `afterResponse` hook, in an event of a failed request, we would like to log the request options.
```javascript
afterResponse: [
      res => {
        if (res.statusCode !== 200) {
          logger.error(
            { res, options: res.req.options, message: res.body },
            'api error'
          );
        }
        return res;
      },
    ],

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -451,6 +451,12 @@ const instance = got.extend({
 
 The response object will typically be a [Node.js HTTP response stream](https://nodejs.org/api/http.html#http_class_http_incomingmessage), however, if returned from the cache it will be a [response-like object](https://github.com/lukechilds/responselike) which behaves in the same way.
 
+##### req
+
+Type: `Object`
+
+- `gotOptions` - The options that were set on this request.
+
 ##### body
 
 Type: `string` `Object` *(depending on `options.json`)*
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by sindresorhus** [comment]
It should be `request`, not `req`.

**Comment by michaltk** [comment]
@sindresorhus @szmarczak any chance of this getting merged soon?

**Comment by muliyul** [comment]
I'm also looking forward to that feature.

File: readme.md [hunk 1]
```diff
@@ -451,6 +451,12 @@ const instance = got.extend({
 
 The response object will typically be a [Node.js HTTP response stream](https://nodejs.org/api/http.html#http_class_http_incomingmessage), however, if returned from the cache it will be a [response-like object](https://github.com/lukechilds/responselike) which behaves in the same way.
 
+##### request
+
+Type: `Object`
+
```
szmarczak: It'd be nice to have a note here that it is *not* [`http.ClientRequest`](https://nodejs.org/api/http.html#http_class_http_clientrequest).

Rowno:
Doesn't this introduce a security issue people logging errors? Since the `body` option could easily contain sensitive data.

sindresorhus:
@Rowno https://github.com/sindresorhus/got/pull/622#issuecomment-472773011

sindresorhus:
@Rowno Next time, open a new issue instead of creating two separate discussions in old PRs about the same thing.

PR #670, opened 2018-12-09
Clarify error docs

### LINKED ISSUE #667: event property in TimeoutError

In the [source code](https://github.com/sindresorhus/got/blob/4a8722b8801851d60b8dbff0941e3f1714964df4/source/errors.js#L101) `TimeoutError` has `event` property, but it's missing in [documentation](https://github.com/sindresorhus/got#gottimeouterror). Can this property be considered as public?

Refs: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30790

#### Checklist

szmarczak:
Clarify some error docs to avoid misunderstanding.

Fixes #667

#### Checklist


sindresorhus:

readme.md
```diff
@@ -671,7 +671,7 @@ The default Got options.
 
 ## Errors
 
-Each error contains (if available) `body`, `statusCode`, `statusMessage`, `host`, `hostname`, `method`, `path`, `protocol`, `url`, and `gotOptions` properties to make debugging easier.
+Each error contains `host`, `hostname`, `method`, `path`, `protocol` and `url` properties to make debugging easier.
```
sindresorhus: I don't think you meant to delete `gotOptions`.

szmarczak: you're right, fixed :D

clocked0ne:
@szmarczak @sindresorhus  I don't see how this is clearer given that `body` is now missing as a property in all cases? Why not leave the original indication at the top of the section and add the relevant "includes" that appear in each type of error?

szmarczak:
> I don't see how this is clearer

Every single error has listed properties. There's no "guessing".

> body is now missing as a property in all cases

According to `body`: good spot, I just forgot to add it to docs :)
According to `property in all cases`: no it's not in all cases. Please check the source code.

clocked0ne:
Okay, so apparently `body` is only on `HTTPError`, but it's still wrong and would be great to see that fixed :)

szmarczak:
> it's still wrong and would be great to see that fixed :)

It is not wrong. It's always been like that. But if you want to request a feature, feel free to make an issue :)

PR #683, opened 2018-12-22
Add `init` hook

szmarczak:
Introduces a new hook: `init`. It gets called right before option normalization - very useful when you need to handle some things on your own.
For example, this fixes the [`jsonReplacer` example in `migration-guides.md`](https://github.com/sindresorhus/got/pull/683/files#diff-a54dbc53336e27b44b3bfab2b8c9b48f).

#### Checklist

### DIFF
--- a/migration-guides.md
+++ b/migration-guides.md
@@ -103,23 +103,11 @@ gotInstance(url, options);
 ```js
 const gotInstance = got.extend({
 	hooks: {
-		beforeRequest: [
+		init: [
 			options => {
 				if (options.json && options.jsonReplacer) {
-					// #1 solution (faster)
-					const newBody = {};
-					for (const [key, value] of Object.entries(options.body)) {
-						let newValue = value;
-						do {
-							newValue = options.jsonReplacer(key, newValue);
-						} while (typeof newValue === 'object');
-
-						newBody[key] = newValue;
-					}
-					options.body = newBody;
-
-					// #2 solution (slower)
-					options.body = JSON.parse(JSON.stringify(options.body, options.jsonReplacer));
+					options.body = JSON.stringify(options.body, options.jsonReplacer);
+					options.json = false; // We've handled that on our own
 				}
 			}
 		],
@@ -127,7 +115,8 @@ const gotInstance = got.extend({
 			response => {
 				const options = response.request.gotOptions;
 				if (options.json && options.jsonReviver) {
-					response.body = JSON.stringify(JSON.parse(response.body, options.jsonReviver));
+					response.body = JSON.parse(response.body, options.jsonReviver);
+					options.json = false; // We've handled that on our own
 				}
 
 				return response;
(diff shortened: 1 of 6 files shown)


readme.md
```diff
@@ -353,6 +353,17 @@ Type: `Object<string, Function[]>`
...
+###### hooks.init
+
+Type: `Function[]`<br>
+Default: `[]`
+
+Called with plain [request options](#options), right before their normalization. This is especially useful in conjunction with [`got.extend()`](#instances) and [`got.create()`](advanced-creation.md) when the input needs custom handling.
+
+See the [Request migration guide](migration-guides.md#breaking-changes) for an example.
+
+**Note**: this hook must be synchronous!
```
szmarczak: That doesn't sound good. It **should** be asynchronous.

szmarczak: It must be synchronous - otherwise `normalizeArguments` would be asynchronous, so errors thrown [here](https://github.com/sindresorhus/got/blob/master/source/create.js#L32) wouldn't be caught.

test/hooks.js
```diff
@@ -129,6 +133,33 @@ test('catches afterResponse errors', async t => {
 	}), errorString);
 });
 
+test('init', async t => {
```
alextes: Can this be more descriptive? Maybe something like 'init is called with options'?

szmarczak: Of course :D I'll rename other hook tests too.

source/normalize-arguments.js
```diff
@@ -125,6 +132,10 @@ const normalize = (url, options, defaults) => {
 	// Override both null/undefined with default protocol
 	options = merge({path: ''}, url, {protocol: url.protocol || 'https:'}, options);
 
+	for (const hook of options.hooks.init) {
+		hook(options);
```
sindresorhus: Let's throw a helpful error if the return value of `hook` is a Promise as it means someone accidentally made it async.

migration-guides.md
```diff
@@ -103,31 +103,23 @@ gotInstance(url, options);
 ```js
 const gotInstance = got.extend({
 	hooks: {
-		beforeRequest: [
+		init: [
 			options => {
+				// Save the original option, so we can look at it in the `afterResponse` hook
+				options._json = options.json;
```
sindresorhus:
```suggestion
				options.originalJson = options.json;
```

?

szmarczak: :+1:

source/normalize-arguments.js
```diff
@@ -125,6 +132,14 @@ const normalize = (url, options, defaults) => {
 	// Override both null/undefined with default protocol
 	options = merge({path: ''}, url, {protocol: url.protocol || 'https:'}, options);
 
+	for (const hook of options.hooks.init) {
+		const called = hook(options);
+
+		if (called instanceof Promise) {
```
sindresorhus: Use `is.promise()` here.

Issue #511, opened 2018-07-10
Cannot pass string as json body with json = true option


ikokostya:
**Versions**

`got`: 8.3.1

**Code example**

Try to call got with **string** as json body:

```js
got(url, {
  method: 'POST',
  json: true,
  body: 'str'
});
```

**Expected behavior**

No errors. The code above is similar to:

```js
got(url, {
   headers: {
       'content-type': 'applicaiton/json'
   },
   body: JSON.stringify('str')
})
```

Also I can pass any other primitive types (`null`, `number`) which are valid in JSON.

**Actual behavior**

I got the following error:

```
 TypeError: The `body` option must be a plain Object or Array when the `form` or `json` option is used
```

brandon93s:
The `json` option requires a JSON request body and a JSON response. When this is not the case, you can set the headers manually to support the endpoints you're interacting with.

More info: https://github.com/sindresorhus/got/issues/467#issuecomment-403189894

ikokostya:
@brandon93s What you mean under `JSON request body`? Body wrapped with `JSON.stringify()` is valid json body, no matter it's object (hash) or not.

brandon93s:
To clarify, the `json` option does the `stringify` and `parse` operations for you while also configuring headers to support JSON. Because of this, a plan object or an array are required so that got can handle the conversion to JSON. When passing a string, got cannot confirm that the input is valid JSON (without a parse anyways) and therefore does not want to make invalid assumptions in configuring headers on your behalf.

ikokostya:
> When passing a string, got cannot confirm that the input is valid JSON

Why? `got` converts body to string with `JSON.stringify`: https://github.com/sindresorhus/got/blob/ad7b361dcb2490c3864b845b979b756f13f7d89b/index.js#L583

`JSON.stringify()` takes any type:

```
> JSON.parse(JSON.stringify(null))
null
> JSON.parse(JSON.stringify(1))
1
> JSON.parse(JSON.stringify('str'))
'str
```

ikokostya:
@brandon93s  ?

ikokostya:
To be clear: `json = true` option is a convenience method to convert body to valid JSON string and setup corresponding request header. Current behavior is incorrect and unobvious. 

Unlike request, response body always parsed https://github.com/sindresorhus/got/blob/ad7b361dcb2490c3864b845b979b756f13f7d89b/index.js#L377 without checking its type, but it can be primitive type.

ikokostya:
[...]

PR #684, opened 2018-12-24
Consistent use of emphasis in readme

alextes:
Curious what you think @sindresorhus . It's small enough that I'd accept a _don't care_. Yet, your OSS experience is much more extensive so feel free to hit reject.

Improves #511

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1035,7 +1035,7 @@ const h2got = got.extend({request});
 
 |                       |     `got`    |   `request`  | `node-fetch` |    `axios`   |
 |-----------------------|:------------:|:------------:|:------------:|:------------:|
-| HTTP/2 support        |      ❔      |       ✖      |       ✖      |       ✖      |
+| HTTP/2 support        |       ❔      |       ✖      |       ✖      |       ✖      |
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -196,12 +196,12 @@ Default: `false`
 
 *If you use `got.stream()`, this option will be ignored.*
 
+*`body` must be a plain object or array and will be stringified.*
```
```suggestion
**Note**: *`body` must be a plain object or array and will be stringified.*
```

alextes:
@szmarczak you pointed out another inconsistency, I thought I'd might as well look at all emphasis use then.

* Prefixed lines warning about the effect of using an option with a `**Note**:`.
* Notes come after properties but before any explanation. This is the middle ground solution I came up with for #511 . It turns out this was already the case for some properties.
* Lines that used full line emphasis to suggest an alternative now use `**Tip**:`.
* I escaped an asterisk (what a word 😂) in the comparison table.

I in no way expect future PRs to follow the above. Wouldn't want to make doc changes hard. It's how people often get involved in projects. Don't want to be difficult there 😊 .

alextes:
@sindresorhus I felt the same 😂! I should've taken the opportunity 🙈.

PR #695, opened 2019-01-11
Ky-like body response transformations

### LINKED ISSUE #671: Ky-like response transformations

#174 #318 #383 #467 #511 #600

#### Describe the feature

Usage:

```js
got(url, options).json();     // a Promise returning JSON
got(url, options).buffer();   // a Promise returning Buffer
got(url, options).fromData(); // a Promise returning From Data
```

Let `options.json` be only responsible for the request. To parse the response you'll need to set `options.responseType` to `json` or just call `got(...).json()`

#### Checklist

szmarczak:
Fixes #671 

#### `options.responseType`

The option may be set to `json`, `buffer`, `text` or be empty (it defaults to `text`).

#### `options.resolveBody`

If set to `true`, the promise will return the Response body instead of the Response.

There are some shortcuts:

#### promise.json()

Automatically sets `options.responseType` to `json` and `options.resolveBody` to true.
Returns the promise.

#### promise.buffer()

Automatically sets `options.responseType` to `buffer` and `options.resolveBody` to true.
Returns the promise.

#### promise.text()

Automatically sets `options.responseType` to `text` and `options.resolveBody` to true.
Returns the promise.

#### `options.json`

The body will be `JSON.parse`d automatically when it's an object/array.

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -159,7 +159,7 @@ Type: `string` `Buffer` `stream.Readable` [`form-data` instance](https://github.
 
 **Note:** If you provide this option, `got.stream()` will be read-only.
 
-**Note:** if body is an object/array and `Content-Type` header is not set, it will be set to `application/json`.
+**Note:** If body is an object/array, it will be stringified. The `Content-Type` header will be set to `application/json` if it's not defined.
 
 The body that will be sent with a `POST` request.
 
@@ -172,7 +172,7 @@ The `content-length` header will be automatically set if `body` is a `string` / 
 Type: `string`<br>
 Default: `text`
 
-**Note**: when using streams, this option is ignored.
+**Note**: When using streams, this option is ignored.
 
 Parsing method used to retrieve the body from the response. Can be `text`, `json` or `buffer`. The promise has `.json()` and `.buffer()` functions which set this option automatically.
 
@@ -374,11 +374,9 @@ See the [Request migration guide](migration-guides.md#breaking-changes) for an e
 Type: `Function[]`<br>
 Default: `[]`
 
-Called with [normalized](source/normalize-arguments.js) [request options](#options). Got will make no further changes to the request before it is sent. This is especially useful in conjunction with [`got.extend()`](#instances) and [`got.create()`](advanced-creation.md) when you want to create an API client that, for example, uses HMAC-signing.
+Called with [normalized](source/normalize-arguments.js) [request options](#options). Got will make no further changes to the request before it is sent (except the body serialization). This is especially useful in conjunction with [`got.extend()`](#instances) and [`got.create()`](advanced-creation.md) when you want to create an API client that, for example, uses HMAC-signing.
 
 See the [AWS section](#aws) for an example.
-
-**Note:** If you modify the `body` you will need to modify the `content-length` header too, because it has already been computed and assigned.
 
 ###### hooks.beforeRedirect
 
@@ -1010,6 +1008,43 @@ const createTestServer = require('create-test-server');
 
 
 ## Tips
+
+### JSON mode
+
+By default, if you pass an object to the `body` option it will be stringified using `JSON.stringify`. Example:
+
+```js
+const got = require('got');
+
+(async () => {
+	const response = await got('httpbin.org/anything', {
+		body: {
+			hello: 'world'
+		},
+		responseType: 'json'
+	});
+
+	console.log(response.body.data);
+	//=> '{"hello":"world"}'
+})();
+```
+
+To receive a JSON body you can either set `responseType` option to `json` or use `promise.json()`. Example:
+
+```js
+const got = require('got');
+
+(async () => {
+	const {body} = await got('httpbin.org/anything', {
+		body: {
+			hello: 'world'
+		}
+	}).json();
+
+	console.log(body);
+	//=> {...}
+})();
+```
 
 ### User Agent
 
--- a/readme.md
+++ b/readme.md
@@ -35,7 +35,7 @@ Got is for Node.js. For browsers, we recommend [Ky](https://github.com/sindresor
 - [Handles gzip/deflate](#decompress)
 - [Timeout handling](#timeout)
 - [Errors with metadata](#errors)
-- [JSON mode](#json)
+- [JSON mode](#json-mode)
 - [WHATWG URL support](#url)
 - [Hooks](#hooks)
 - [Instances with custom defaults](#instances)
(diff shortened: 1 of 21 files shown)


sindresorhus:
> I'm still not sure if promise.json() should return a Response object or just the JSON...

Yeah. I'm conflicted too. I would really like it to be the same as `ky`, which does return the JSON directly, but then there's no shortcut to get the body as JSON, but also the response body. Well, there is, you could use the `responseType` option if you also need the response and not just the body.

sindresorhus:
> the json option was removed. The body will be JSON.parsed automatically when it's an object/array.

What's the argument for just handling automatically in the `body` option instead of what `ky` does, which is to have a `json` option that the user gives an object/array instead of the `body` option. I'm not sure which solution is the best, but I would like `got` and `ky` to agree on this either way. We should look at how other request libraries handles this (even not just in the Node.js ecosystem).

szmarczak:
> What's the argument for just handling automatically in the body option instead of what ky does, which is to have a json option that the user gives an object/array instead of the body option.

Actually I see no difference here. 🤔 The only thing I see is that `json` **must** be an object/array. While body handles the type automatically...

sholladay:
Generally speaking, I like the idea of `body` being automatically serialized to JSON where possible - smaller API surface and all that. It does limit the user a bit though, by only supporting arrays and objects at the top level. I'm sure that is fine for most REST API calls, but there may be some outliers where people want to send other types of values as JSON. `ky`'s `json` option makes this easy, because it is an alternative to `body`. Actually, it's best practice to [only send objects](https://stackoverflow.com/questions/3503102/what-are-top-level-json-arrays-and-why-are-they-a-security-risk) at the top level, for security reasons, but I don't know how many people are aware of this.

It also makes string bodies feel kind of ambiguous to me. For example, does `got(url, { body : 'hello world' })` send a request body of `'hello world'` or `'"hello world"'` (the latter being the result of `JSON.stringify('hello world')`)? I could imagine this tripping some people up, no matter which one we actually do. The semantics are a bit blurry unless you study the docs.

I guess overall I'm +0.5 on removing the `json` option. 👍 

I think `promise.json()` should only return the JSON, not the response. I almost never need the response when the response is 200 OK, I generally only need it when an error occurs. One thing that `got` does, which I think it awesome, and which we don't have in `ky`, is setting the `accept` header when `.json()` is called.

The `responseType` option is a bit more peculiar to me. While I can see the benefit for someone who _does_ need the response in addition to the JSON (or Buffer or whatever), I really don't think it's that common, and where it is needed, there may be better solutions. For example, the main case I can think of is pagination that requires reading response headers. But pagination should probably be solved with async iterators (as Sindre mentioned [here](https://github.com/sindresorhus/gh-got/issues/22#issuecomment-397380530)) or other means. And imitating `.json()` manually in userland is pretty trivial if someone _really_ needs it for some edge case.

sholladay:
> Oh, believe me, it's very common.

What do people need it for?

szmarczak:
> What do people need both the Response and the JSON for, on a successful response, other than pagination?

For example API rate limiting.

sholladay:
A client should generally keep making requests as normal until it receives a [429 Too Many Requests](https://httpstatuses.com/429), which is of course an error response, and `error.reponse` ought to be provided, for sure.

I don't find pre-emptive, client-side rate limiting based on successful API responses to be very useful. The client inherently has no way to know the state of rate limiting until it makes a request anyway. The response details about rate limiting are completely non-standardized and different providers handle it differently. And then even if you normalize the response details and keep track of them across requests, the data can get stale because you have no way of knowing whether there are other clients in use that are affecting the rate limit.

Maybe it's still worth it. I'm just skeptical.

sholladay:
But is it worth having `responseType` to make bots a little easier to write by, say, 4 or 5 lines of code? FWIW, any API that bans on or before the first 429 response is a **terrible** API.

szmarczak:
> But is it worth having responseType to make bots a little easier to write by, say, 4 or 5 lines of code?

I'm pretty sure there are another use cases too :)

> FWIW, any API that bans on or before the first 429 response is a terrible API.

Agreed.

sholladay:
Sure, maybe. If you want to keep `responseType`, I would still have `got(url, { responseType : 'json' })` return the Response and `got(url).json()` return the parsed object. In other words, `responseType` is about semantics of the request and response, whereas `.json()` and friends are additional sugar for getting the response body itself. I would strongly prefer not to return the whole Response from `.json()`.

szmarczak:
@sholladay Then `.text()` and `.buffer()` should return the body too I think. The behavior can't be ambiguous.

@sindresorhus What do you think about it? I'm still not sure about the `json`/`body` option. `json` is good if someone's forcing something to be JSON. `body` is good for objects/arrays (I don't think someone would pass a string to `json` though). I can't decide...

sindresorhus:
So this is decided:

- `responseType: 'json'` returns the response like normal
- `.json()`, `.text()` and `.buffer()` returns the body directly (no response).
	We should document that users need to use `responseType` instead if they want the full response. These methods are just shortcuts.

I'm still undecided about `json` vs `body`. I'll need a few more days to think about it.

readme.md
```diff
@@ -157,12 +157,36 @@ Type: `string` `Buffer` `stream.Readable` [`form-data` instance](https://github.
...
+
+Parsing method used to retrieve the body from the response. Can be `text`, `json` or `buffer`. The promise has `.json()` and `.buffer()` functions which set this option automatically.
+
+Example:
+
+```js
+const {body} = await got(url).json();
+```
+
+###### resolveBody
```
sholladay: Should this be a public option?

Would there be any reason for a user to set this to `true` instead of using one of the shortcut methods?

szmarczak: Custom instances:

```js
const instance = got.extend({
	resolveBody: true,
	responseType: 'json'
});

await instance(url) // returns the body
```

szmarczak: > With that example there's no way to get the whole response though

Nope:

```js
instance(url, {resolveBody: false}); // returns the whole response instead
```

> So I don't think this option should be public.

Then what if someone wants the body instead of the response for all requests?

szmarczak: It should be named `resolveBodyOnly`

szmarczak:
f u git

szmarczak:
`git` has just messed up. When I solved conflicts it just shouted out that I did not. Opening a new PR.

PR #696, opened 2019-01-12
Add `beforeError` hook

szmarczak:
Fixes https://github.com/sindresorhus/gh-got/blob/master/index.js

Instead using handler we could do:

```js
hooks: {
	beforeRequest: [
		options => {
			if (options.token) {
				options.headers.authorization = options.headers.authorization || `token ${options.token}`;
			}
		}
	],
	afterResponse: [
		response => {
			response.rateLimit = getRateLimit(response);
			return response;
		}
	],
	onError: [
		error => {
			const {response} = error;

			if (response && response.body) {
				error.name = 'GitHubError';
				error.message = `${response.body.message} (${error.statusCode})`;
			}

			if (response) {
				error.rateLimit = getRateLimit(response);
			}

			return error;
		}
	]
}
```

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -463,7 +463,9 @@ const instance = got.extend({
 Type: `Function[]`<br>
 Default: `[]`
 
-Called with an `Error` instance. The error is passed to the hook right before it's thrown. This is especially useful when you want to have more detailed errors. Example:
+Called with an `Error` instance. The error is passed to the hook right before it's thrown. This is especially useful when you want to have more detailed errors.
+
+**Note**: errors thrown while normalizing options are thrown directly.
 
 ```js	
 const got = require('got');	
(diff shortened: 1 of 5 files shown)


sindresorhus:
> Fixes sindresorhus/gh-got:index.js@master

I wouldn't say it fixes that issue. This just works around it. We still need to properly fix `handler`.

readme.md
```diff
@@ -447,6 +447,38 @@ const instance = got.extend({
...
+				const {response} = error;
+
+				if (response && response.body) {
+					error.name = 'GitHubError';
+					error.message = `${response.body.message} (${error.statusCode})`;
+				}
+
+				if (response) {
+					error.rateLimit = getRateLimit(response);
+				}
```
sindresorhus: I would drop this part. The example is already complete without this.

sindresorhus:
Does the hook catch absolutely *all* errors? If not, it should be documented what it doesn't catch. For example, I don't think it currently catches input validation errors (Neither should it IMHO).

sindresorhus:
Wouldn't `beforeError` be a more correct name for this?

sindresorhus:

szmarczak:
done

PR #698, opened 2019-01-12
Make the query values more strict

szmarczak:


source/normalize-arguments.js
```diff
@@ -136,6 +136,19 @@ const normalize = (url, options, defaults) => {
 	const {query} = options;
 	if (is.nonEmptyString(query) || is.nonEmptyObject(query) || query instanceof URLSearchParams) {
 		if (!is.string(query)) {
+			if (!(query instanceof URLSearchParams)) {
+				const verify = (value, type) => {
+					if (typeof value !== 'string' && typeof value !== 'number' && typeof value !== 'boolean' && value !== null) {
```
sindresorhus: Let's use the `.is` methods here.

source/normalize-arguments.js
```diff
@@ -136,6 +136,19 @@ const normalize = (url, options, defaults) => {
...
+						throw new TypeError(`The query ${type} '${value}' must be a string, number, boolean or null`);
+					}
+				};
+
+				for (const [key, value] of Object.entries(query)) {
+					verify(key, 'key');
+					verify(value, 'value');
+				}
+			}
```
sindresorhus: This is becoming quite nested. Can you extract the logic here into a top-level `validateSearchParams()` function?

Issue #700, opened 2019-01-13
Rewrite Got in TypeScript


sindresorhus:
We want typings available and keeping a type definition file in sync with the actual codebase is a lot of work in itself and it's easy for them to get out of sync. I also think the codebase could benefit from being written in TypeScript, as the code is quite complex and it's easy to make subtle mistakes.

gajus:
What is the reason for choosing TS of Flow?

sindresorhus:
It's mainly about familiarity and popularity. Popularity means it's easier to attract potential contributors. But I've also noticed that TS has better error messages, and the fact that TS is written in TS (meaning JS) is definitely a bonus. Very few knows OCaml.

gajus:
@sindresorhus Do you have an example of a specific error that TS produces that is more developer friendly than what Flow would produce?

The other two arguments make sense. OCaml codebase indeed makes it for a steep entry barrier to contribute to the codebase, and TS has gained a lot of popularity over the last 6 months.

gajus:
That comment is one year old.

Flow errors improved _a lot_ over the last 12 months.

Now errors show where the issue originates and walks you through all incompatibilities down the graph.

<img width="1092" alt="screenshot 2019-01-14 at 10 37 43" src="https://user-images.githubusercontent.com/973543/51107963-7720ca80-17e8-11e9-9e35-298916b35168.png">

This helps a lot with refactoring.

I am based of course as I hardly use TS. Flow feels like a second skin.

Either way, I think that TS is the right choice for Got for the other reasons you have suggested. I was more curious what is driving the decision rather than influencing the decision.

ferdaber:
This looks like an interesting project- I'd like to help! I can start with at least adding the TypeScript build/test steps.

medikoo:
Aren't you worried that complexity it'll add, will shadow the benefit?

`got` codebase is now pretty straightforward to get through, and library works very well (doesn't feel buggy in any part).

What real problem converting to TS aims to solve?

medikoo:
> Adding types will let us be more confident about doing changes

Do you remember any specific issues that wouldn't happen if it'll be TS already?

This move also makes more difficult to contribute for JS developers not interested or not familiar with TS. It narrows the audience.

sindresorhus:
> Can't this be solved simply with JSDoc? I haven't explored that deeply, but I think TS can read types of out it as well

No, JSDoc helps to some degree, but TS is much more powerful.

[...]

PR #705, opened 2019-01-19
convert progress.js to typescript

scttcper:
converted source/progress.js to progress.ts #700 

Thought I'd start with one file and get feedback.

transform callback takes undefined and not null, had to change that
```ts
callback(null, chunk);
callback(undefined, chunk);
```

not sure how you'd like to handle the two lint errors


source/progress.ts
```diff
@@ -1,96 +1,95 @@
...
-					emitter.emit('downloadProgress', {
-						percent,
-						transferred: downloaded,
-						total: downloadBodySize
-					});
-				}
+import { Transform } from 'stream';
+import { EventEmitter } from 'events';
+
+export function download(response: any, emitter: EventEmitter, downloadBodySize?: number) {
```
sindresorhus: Can you find a proper type for the `response` object? Should be in the NodeJS types.

source/progress.ts
```diff
@@ -1,96 +1,95 @@
...
-			clearInterval(progressInterval);
-		});
+			callback();
+		}
+	});
+}
 
-		request.once('response', () => {
-			clearInterval(progressInterval);
+export function upload(request: any, emitter: EventEmitter, uploadBodySize?: number) {
```
sindresorhus: Same as https://github.com/sindresorhus/got/pull/705/commits/c3c71b0ac63ea6c54efb35623fe9ae88285b1ab5#r249244253

source/progress.ts
```diff
@@ -1,96 +1,95 @@
...
+		emitter.emit('uploadProgress', {
+			percent: 1,
+			transferred: uploaded,
+			total: uploadBodySize
 		});
-	}
-};
+	});
+
+	request.once('socket', (socket: any) => {
```
sindresorhus: Same as https://github.com/sindresorhus/got/pull/705/commits/c3c71b0ac63ea6c54efb35623fe9ae88285b1ab5#r249244253

source/progress.ts
sindresorhus: You could use a default value for `downloadBodySize?: number`.

szmarczak: I believe it's `import {IncomingMessage} from 'http'`;

scttcper:
Thanks for the dope feedback.

@szmarczak @sindresorhus what would you like to do about https://nodejs.org/api/all.html#deprecations_dep0066_outgoingmessage_headers_outgoingmessage_headernames

getHeaders() doesn't return a buffer

szmarczak:
@scttcper I think it should be `(request as any)._header` then.

sindresorhus:
Looks good now. Thank you, @scttcper 🙌

PR #706, opened 2019-01-20
Add support for Brotli encoding

sindresorhus:
When using Node.js 11.7.0 or later.

// @kevva @silverwind

### DIFF
--- a/source/normalize-arguments.js
+++ b/source/normalize-arguments.js
@@ -5,6 +5,7 @@ const is = require('@sindresorhus/is');
 const lowercaseKeys = require('lowercase-keys');
 const urlToOptions = require('./utils/url-to-options').default;
 const validateSearchParams = require('./utils/validate-search-params');
+const supportsBrotli = require('./utils/supports-brotli').default;
 const merge = require('./merge');
 const knownHookEvents = require('./known-hook-events').default;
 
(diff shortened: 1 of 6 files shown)


source/normalize-arguments.js
```diff
@@ -5,6 +5,7 @@ const is = require('@sindresorhus/is');
 const lowercaseKeys = require('lowercase-keys');
 const urlToOptions = require('./utils/url-to-options').default;
 const validateSearchParams = require('./utils/validate-search-params');
+const supportsBrotli = require('./utils/supports-brotli');
```
szmarczak: no `.default` here?

source/utils/supports-brotli.ts
```diff
@@ -0,0 +1,3 @@
+import zlib from 'zlib';
+
+export default (): boolean => typeof (zlib as any).createBrotliDecompress === 'function';
```
szmarczak: Why is it a function?

source/normalize-arguments.js
```diff
@@ -193,7 +194,7 @@ const normalize = (url, options, defaults) => {
 	}
 
 	if (options.decompress && is.undefined(headers['accept-encoding'])) {
-		headers['accept-encoding'] = 'gzip, deflate';
+		headers['accept-encoding'] = supportsBrotli ? 'gzip, deflate, br' : 'gzip, deflate';
```
szmarczak: If `supportsBrotli` needs to be a function then it should be `supportsBrotli()`

source/utils/supports-brotli.ts
sindresorhus: Mistake

source/normalize-arguments.js
sindresorhus: Yes, mistake

PR #710, opened 2019-01-21
[BROKEN] Asynchronous handlers

### LINKED ISSUE #701: The `handler` method in `got.create()` should be able to be an async function
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/701)

This is currently not possible since the `next()` function can also return a stream.

Example of where this could have been useful: https://github.com/sindresorhus/gh-got/blob/bf13dc267a9dbac1e1a5be3542e448bf3c151781/index.js#L32-L51

> There is a $100.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/got/issues/701).

szmarczak:
Fixes #701 

#### Checklist

### DIFF
--- a/source/as-promise.js
+++ b/source/as-promise.js
@@ -1,6 +1,7 @@
 'use strict';
 const EventEmitter = require('events');
 const getStream = require('get-stream');
+const pEvent = require('p-event');
 const is = require('@sindresorhus/is');
 const PCancelable = require('p-cancelable');
 const requestAsEventEmitter = require('./request-as-event-emitter');
@@ -8,8 +9,9 @@ const {HTTPError, ParseError, ReadError} = require('./errors');
 const {mergeOptions} = require('./merge');
 const {reNormalize} = require('./normalize-arguments');
 
-const asPromise = options => {
+const asPromise = () => {
 	const proxy = new EventEmitter();
+	let options = {};
 
 	const parseBody = response => {
 		if (options.responseType === 'json') {
@@ -21,7 +23,23 @@ const asPromise = options => {
 		}
 	};
 
-	const promise = new PCancelable((resolve, reject, onCancel) => {
+	const promise = new PCancelable(async (resolve, reject, onCancel) => {
+		try {
+			const receivedOptions = await pEvent(proxy, 'options');
+
+			if (promise.isCanceled) {
+				return;
+			}
+
+			options = {
+				...receivedOptions,
+				...options
+			};
+		} catch (error) {
+			reject(error);
+			return;
+		}
+
 		const emitter = requestAsEventEmitter(options);
 
 		onCancel(emitter.abort);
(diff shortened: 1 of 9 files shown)


source/as-promise.js
```diff
@@ -2,110 +2,122 @@
...
 const requestAsEventEmitter = require('./request-as-event-emitter');
 const {HTTPError, ParseError, ReadError} = require('./errors');
 const {options: mergeOptions} = require('./merge');
 const {reNormalize} = require('./normalize-arguments');
 
-const asPromise = options => {
+const asPromise = () => {
+	let options = {};
 	const proxy = new EventEmitter();
+	const {promise, resolve, reject, onCancel} = detachedPCancelable();
```
sindresorhus: This pattern is called deferred promise and is generally an anti-pattern, as unlike the constructors, thrown errors won't be caught and won't result in promise rejection.

source/create.js
```diff
@@ -27,16 +26,23 @@ const create = defaults => {
 		defaults.handler = (options, next) => next(options);
 	}
 
+	if (defaults.handler.constructor.name !== 'AsyncPromise') {
```
sindresorhus: This is not a reliable check. What are you trying to do here?

source/create.js
```diff
@@ -27,16 +26,23 @@ const create = defaults => {
...
+		defaults.handler = util.promisify(defaults.handler);
+	}
+
 	function got(url, options) {
```
sindresorhus: Use arrow function.

source/create.js
```diff
@@ -27,16 +26,23 @@ const create = defaults => {
...
+		const promiseOrStream = options && options.stream ? asStream() : asPromise();
+
+		// eslint-disable-next-line promise/prefer-await-to-then
+		normalizeArguments(url, options, defaults).then(normalizedOptions => {
+			defaults.handler(normalizedOptions, promiseOrStream._sendRequest).catch(error => {
+				promiseOrStream._reject(error);
+			});
+		}).catch(error => {
+			promiseOrStream._reject(error);
+		});
```
sindresorhus: This could be simplified to:

```js
(async () => {
	try {
		const normalizedOptions = await normalizeArguments(url, options, defaults);
		await defaults.handler(normalizedOptions, promiseOrStream._sendRequest);
	} catch (error) {
		promiseOrStream._reject(error);
	}
})();
```

sindresorhus:
Can you also update the `got.create()` docs showing how the async handlers will be used?

source/create.js
szmarczak: Nevermind. I'll fix it :P

source/as-promise.js
szmarczak: > thrown errors won't be caught

Can you elaborate on this one? I'm not sure what do you mean by that...

szmarczak:
Reworked, it's much readable now :P The previous version of this PR was too bloated.

> Can you explain your changes?

In short, I added `await pEvent(proxy, 'options')` before `requestAsEventEmitter(options)`.
So normalizing options can be async. And if that can be async, then **all** hooks can be async too - including handlers ;)

> Can you also update the `got.create()` docs showing how the async handlers will be used?

What do you mean?

PR #712, opened 2019-01-23
Rewrite `errors.js` in TypeScript

ricardorojass:
#### Checklist

- [x] I have read the documentation.
- [x] I have included a pull request description of my changes.
- [ ] I have included some tests.
- [ ] If it's a new feature, I have included documentation updates.

#### Changes

* Rewrite `errors.js` in TypeScript
* Change requires in `as-promise.js, as-stream.js, create.js, request-as-event-emitter.js` to the format require('foo').default

#### Doubts

* What type I should use in `error` parameter?

Thank U :v:

PR #708, opened 2019-01-21
Rewrite `get-response.js` in TypeScript

ltciro:
#700 
#### Checklist

- [x] I have read the documentation.
- [x] I have included a pull request description of my changes.
- [ ] I have included some tests.
- [ ] If it's a new feature, I have included documentation updates.

#### Changes 
- Rewrite `get-response.js` in TypeScript
- Change  `require` in `request-as-event-emitter.js` to the format `require('foo').default`
- Add a type null in `progress.js` for be able to send `downloadBodySize` from `get-response.js`
#### Doubts 
- imports of decompress-response and mimic-respons, Should I create d.ts files, install @types or leave it as `require`?
- I don't know all properties of `options` and maybe the interface should be in some shared space?  

**Tests run well**

Thank You 😸

PR #714, opened 2019-01-27
Rewrite `merge.ts` in TypeScript

fnky:
#### Checklist

#### Changes

- Rewrite `merge.ts` in TypeScript.
- Add additional types for hooks for type completion in `known-hook-events.ts`.
- Changes the exported API slightly due to the way they are exported in TypeScript.


source/merge.ts
```diff
@@ -0,0 +1,93 @@
+import {URL} from 'url';
+import is from '@sindresorhus/is';
+// XO doesn't recognize `export type` for some reason.
+// eslint-disable-next-line import/named
+import knownHookEvents, {Hooks, HookType, HookEvent} from './known-hook-events';
+
+// TODO: Use the Got options-object types.
+interface Options {
```
fnky: The extranous `Options` interface is meant to be merged with the interface that was introduced in https://github.com/sindresorhus/got/pull/708 into a single shared interface.

source/merge.ts
```diff
@@ -0,0 +1,93 @@
+import {URL} from 'url';
+import is from '@sindresorhus/is';
+// XO doesn't recognize `export type` for some reason.
+// eslint-disable-next-line import/named
```
fnky: XO fails at linting for exported `type`s from `known-hook-events.ts`.

source/merge.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function mergeOptions(...sources: Options[]) {
+	sources = sources.map(source => source || {});
+	const merged = merge({}, ...sources);
+
+	// TODO: This is a funky situation. Even though we "know" that we're going to
+	//       populate the `hooks` object in the loop below, TypeScript want us to //       put them into the object upon initialization, because it cannot infer
+	//       that they are going to conform correctly in runtime.
+	// eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
+	const hooks = {} as { [Key in HookEvent]: HookType[] };
```
fnky: This one is quirky and probably not correct, but I have tried multiple ways to reflect that we _know_ that it's going to be populated later, in the loop underneath but we don't want to give it the values upfront.

source/merge.ts
```diff
@@ -0,0 +1,93 @@
...
+	merged.hooks = hooks as Hooks;
+
+	return merged;
+}
+
+type NextFunction = (error?: Error | string) => void;
+
+type IterateFunction = (options: Options) => void;
+
+type Method = 'GET' | 'PUT' | 'HEAD' | 'DELETE' | 'OPTIONS' | 'TRACE';
```
fnky: Maybe these types should be exported from another file to make them available across the codebase as well as the end-user API, if necessary.

source/known-hook-events.ts
```diff
@@ -1,8 +1,47 @@
-export default [
+import {RequestOptions} from 'https';
+import {IncomingMessage} from 'http';
+
+// TODO: This may be a bit overkill. It was done for different reasons:
```
fnky: I hope you understand the rationale here, even though it does look quite cumbersome.

source/known-hook-events.ts
```diff
@@ -1,8 +1,47 @@
...
+	/**
+	 * Called with plain request options, right before their normalization. This is especially useful in conjunction with got.extend() and got.create() when the input needs custom handling.
+	 *
+	 * @default []
+	 */
+	init: InitHook[];
+	beforeRequest: BeforeRequestHook[];
+	beforeRedirect: BeforeRedirectHook[];
+	beforeRetry: BeforeRetryHook[];
+	afterResponse: AfterResponseHook[];
```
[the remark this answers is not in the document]
sindresorhus: Should the docs be here or on the actual types? For example, line 11.

source/merge.ts
fnky: I wasn’t sure we wanted to do this across the project, but it makes it less confusing for future contributors.

source/known-hook-events.ts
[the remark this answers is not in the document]
fnky: This was on oversight on my part. I think I used this to check if you could document each property, and I think you can. I’ll suggest we do it on the properties as well as the types, since both can be consumed by different means.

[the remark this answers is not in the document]
fnky: Sorry I got it confused with on the type of `Hooks` vs. the properties that takes the list of hooks.

I tried to add the documentation at the type level, but TypeScript/VSCode apparently can't show proper intellisense when you type a function in the array. I think it makes sense to add them to the properties in that case.

fnky:
I have pushed a few changes regarding the review.

- Moved the `Options` interface into `utils/types.ts` so it can be reused.
- Added a new `CancelableRequest` type which extends from `PCancelable` with additional properties.
- Fixed some types that were incorrect.
- Added documentation for `Hooks` and its properties.

There are still some things that are needed to make it closer to what is documented:

##### `beforeError` error types

The error passed to the `beforeError` hook can be a number of different types of errors with different properties. I initially tried to make a `AnyError` type that was a union of these different errors, but I'll wait for #712 to be merged so we can discuss any concerns regarding what errors/properties are missing.

##### Extended Response object

There are multiple places where it is documented that you can retrieve for example `.body` from the response. Since `IncomingMessage` doesn't have this, we'll need to either make our own interface, based on [responselike](https://github.com/lukechilds/responselike) or use responselike's own `Response`.

source/merge.ts
```diff
@@ -0,0 +1,85 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
+	const hooks = {} as { [Key in HookEvent]: HookType[] };
+	for (const hook of knownHookEvents) {
+		hooks[hook] = [];
+	}
+
+	for (const source of sources) {
+		if (source.hooks) {
+			for (const hook of knownHookEvents) {
+				hooks[hook] = hooks[hook].concat(source.hooks[hook] || []);
```
fnky: I caught an issue here, where empty arrays passed by the user would otherwise end up being an Array with an `undefined` item. Kudos to TypeScript.

source/known-hook-events.ts
```diff
@@ -1,35 +1,67 @@
...
+	 * Called with normalized [request options](https://github.com/sindresorhus/got#options), the error and the retry count. Got will make no further changes to the request. This is especially useful when some extra work is required before the next try.
+	 *
+	 * @default []
+	 */
 	beforeRetry: BeforeRetryHook[];
+
+	/**
+	 * Called with an `Error` instance. The error is passed to the hook right before it's thrown. This is especially useful when you want to have more detailed errors.
+	 *
+	 * @note Errors thrown while normalizing input options are thrown directly and not part of this hook.
```
sindresorhus: I don't think there's any `@note` tag. I usually just use `Note: `.

source/merge.ts
```diff
@@ -60,16 +61,14 @@ type NextFunction = (error?: Error | string) => void;
 
 type IterateFunction = (options: Options) => void;
 
-type Method = 'GET' | 'PUT' | 'HEAD' | 'DELETE' | 'OPTIONS' | 'TRACE';
```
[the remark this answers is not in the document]
fnky: What do you think to allow lowercase too, e.g.

```ts
type Method = 'GET' | 'PUT' | 'HEAD' ... | 'get' | 'put' | 'head' ...
```

source/known-hook-events.ts
```diff
@@ -1,8 +1,81 @@
-export default [
+import {IncomingMessage} from 'http';
+import {Options, CancelableRequest} from './utils/types';
+import {HTTPError} from './errors';
+
+export type InitHook = (options: Options) => void;
+// TODO: The `Error` type should confirm to any possible extended error type that can be thrown. See https://github.com/sindresorhus/got#hooksbeforeerror
+export type BeforeErrorHook = (error: Error) => Error | void | Promise<Error> | Promise<void>;
```
szmarczak: I'm not sure why the `void` type is here... Shouldn't it be just `Error | Promise<Error>`?

source/known-hook-events.ts
```diff
@@ -1,8 +1,81 @@
...
+export type BeforeRequestHook = (options: Options) => void | Promise<void>
+export type BeforeRedirectHook = (options: Options) => void | Promise<void>
+export type BeforeRetryHook = (options: Options, error: HTTPError, retryCount: number) => void | Promise<void>
+export type AfterResponseHook = (response: IncomingMessage, retryWithMergedOptions: (options: Options) => IncomingMessage | CancelableRequest<IncomingMessage>) => IncomingMessage | CancelableRequest<IncomingMessage>
```
szmarczak: I think it should be

```diff
-retryWithMergedOptions: (options: Options) => IncomingMessage | CancelableRequest<IncomingMessage>
+retryWithMergedOptions: (options: Options) => CancelableRequest<IncomingMessage> 
```

source/known-hook-events.ts
```diff
@@ -1,8 +1,81 @@
...
+
+export type HookType = BeforeErrorHook | InitHook | BeforeRequestHook | BeforeRedirectHook | BeforeRetryHook | AfterResponseHook;
+
+/**
+ * Hooks allow modifications during the request lifecycle. Hook functions may be async and are run serially.
+ */
```
szmarczak: @sindresorhus Do we need to double the documentation?

source/merge.ts
szmarczak: I'm not a TypeScript master either :P

Is there maybe another way? For example the [Hooks interface](https://github.com/sindresorhus/got/pull/714/files#diff-8cbb1059c76170f4b0419c6df7f197baR18):

```ts
const hooks: Hooks = {}
```

source/merge.ts
```diff
@@ -0,0 +1,85 @@
...
+			}
+		}
+	}
+
+	return target;
+}
+
+export function mergeOptions(...sources: Options[]) {
+	sources = sources.map(source => source || {});
+	const merged = merge({}, ...sources);
```
szmarczak: The type of `merged` should be `Options`

source/merge.ts
```diff
@@ -0,0 +1,93 @@
...
+
+	for (const source of sources) {
+		if (source.hooks) {
+			for (const hook of knownHookEvents) {
+				hooks[hook] = hooks[hook].concat(source.hooks[hook] || []);
+			}
+		}
+	}
+
+	merged.hooks = hooks as Hooks;
```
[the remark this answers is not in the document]
szmarczak: [This would fix that](https://github.com/sindresorhus/got/pull/714/files#r251524354)

source/utils/types.ts
```diff
@@ -0,0 +1,21 @@
...
+import {Hooks} from '../known-hook-events';
+
+export type Method = 'GET' | 'PUT' | 'HEAD' | 'DELETE' | 'OPTIONS' | 'TRACE' | 'get' | 'put' | 'head' | 'delete' | 'options' | 'trace';
+
+export interface Options extends RequestOptions {
+	hooks?: Partial<Hooks>;
+	decompress?: boolean;
+	encoding?: BufferEncoding | null;
+	method?: Method;
+	[ key: string ]: unknown | Options;
```
szmarczak: Is the `Options` type needed here?

source/known-hook-events.ts
sindresorhus: https://github.com/sindresorhus/got/pull/714#discussion_r251441415

source/utils/types.ts
sindresorhus: We should also add a comment that the line should be removed in the future when the TS migration is complete.

source/known-hook-events.ts
szmarczak: Agreed on that thing. Sorry I didn't shape my question correctly: I mean, generally, do we need to redocument *everything* in the code, isn't the README just enough?

source/known-hook-events.ts
fnky: I think this may be a confusion with the line as it is currently:

```ts
retryWithMergedOptions: (options: Options) => CancelableRequest<IncomingMessage>
```

I'll make a type for `RetryFunction` instead to make it more readable.

source/known-hook-events.ts
fnky: I think the issue with using the README is that you quickly end up with out of date documentation.

For example in the documentation for [`beforeError`](https://github.com/sindresorhus/got/#hooksbeforeerror) hook, the example uses `onError` as the property name—this could be a typo or a change that didn't get updated, but nevertheless it makes it more difficult.

If we redocument everything, we can later generate the documentation.

szmarczak: > If we redocument everything, we can later generate the documentation.

Looks promising! Let's do that way then :)

source/merge.ts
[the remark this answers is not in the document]
fnky: We could, but this then requires TypeScript users to import it every time the need to update `method`. Do you think that would be too cumbersome for this?

source/merge.ts
fnky: Added explicit return type for `merged` function.

source/known-hook-events.ts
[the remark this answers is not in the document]
fnky: So the consensus is to add the same documentation for both types and the properties in Hooks?

fnky:
I have addressed some of the comments :)

source/merge.ts
szmarczak: ping?

fnky: I have tried that and also with `as Hooks`.

```ts
hooks[hook] = hooks[hook].concat(source.hooks[hook] || []);
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         [ts] Cannot invoke an expression whose type lacks a call signature.
                         Type '{ (...items: ConcatArray<InitHook>[]): InitHook[]; (...items: (InitHook | ConcatArray<InitHook>)[]): InitHook[]; } | { (...items: ConcatArray<BeforeRequestHook>[]): BeforeRequestHook[]; (...items: (BeforeRequestHook | ConcatArray<...>)[]): BeforeRequestHook[]; } | { ...; } | { ...; } | { ...; } | { ...; }'
                         has no compatible call signatures. [2349]
```

Seen as this line is temporary, I don't think its much of an issue just an annoyance.

source/known-hook-events.ts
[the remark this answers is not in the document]
sindresorhus: Yup

source/merge.ts
szmarczak: Oh, thanks for the explaination, it's clear now. I'll try to think of something...

sindresorhus:
Thank you, @fnky! This is excellent 🙌

PR #720, opened 2019-02-02
Rewrite `as-stream.js` in TypeScript

Magellol:
Hey guys 👋 

According to #700, I have started to tackle `as-stream`. I've made good progress already but I have one blocker in my build I need to get confirmation for. My typing doesn't seem to be quite accurate based on the usage of the lib, although I wasn't entirely sure what it should be.

I've also left a few comments throughout the code and I'd like to quickly talk about them so I can remove the `todo`s. I can also document exported interfaces once I'm 100% sure my typings are accurate.

#### Checklist


source/as-stream.ts
```diff
@@ -1,16 +1,21 @@
...
 	const input = new PassThrough();
 	const output = new PassThrough();
 	const proxy = duplexer3(input, output);
 	const piped = new Set();
 	let isFinished = false;
 
+	// @todo
+	// RetryDescriptor#retries looks like it could be either a number (according to `defaults`) or a function
+	// that returns a number. Once confirmed with maintainers, this could be a bug or a bad typing on my end.
 	options.retry.retries = () => 0;
```
Magellol: This is the piece of code that fails the build. According to the [doc](https://github.com/sindresorhus/got#retry) `retries` can be `number | () => number`.

TS complains because in `defaults` we set this value to be `2` but we quickly override it here regardless of it's initial value. Once I understand more how this is used, maybe I'll be able to fix the typings.

source/as-stream.ts
```diff
@@ -19,12 +24,12 @@ module.exports = options => {
 		};
 	}
 
-	const emitter = requestAsEventEmitter(options, input);
+	const emitter = requestAsEventEmitter(options, input) as RequestEmitter;
```
Magellol: Weirdly enough, I had to type cast this value, as opposed to simply typing it.
`new EventEmitter` from `node#events` seems to return a type of `internal` as opposed to return a `EventEmitter`. I'm a bit confused, it might be an issue with the node typings?

source/get-response.ts
```diff
@@ -15,7 +15,11 @@ export default (response: IncomingMessage, options: Options, emitter: EventEmitt
 
 	mimicResponse(response, progressStream);
 
-	const newResponse: PassThrough | Transform = options.decompress === true &&
+	// @todo
+	// I believe this typings was slightly wrong because the way `response` is used in 
+	// `as-stream.ts::52` it requires a `statusCode` which `IncomingMessage` from node has.
+	// Maybe a maintainer can guide me a bit here.
+	const newResponse: Response = options.decompress === true &&
```
Magellol: I'd like to confirm this change makes sense. I'm not super knowledgeable about what the actual code is doing. If someone could clarify this for me.

source/index.js
```diff
@@ -2,7 +2,12 @@
 const pkg = require('../package.json');
 const create = require('./create');
 
-const defaults = {
+// @todo
+// I exported this default to be able to infer typings from it.
+// I'm guessing this `defaults` is being merged with user options
+// so `as-stream` receives an object that has merged props.
+// Let's get a confirmation from a maintainer but I can also create an explicit typings for it.
+export const defaults = {
```
Magellol: Like the comment says, I can also create an explicit typing for the `defaults` value. I use this value as a type directly using `typeof` operator.

I just wanted to make sure:
A) Are you guys okay with this, or would you prefer an explicit Default typing
B) Am I correct to assumre that `as-stream` receives a merged object from default and passed in options?

source/utils/types.ts
```diff
@@ -47,8 +52,16 @@ export interface InterfaceWithDefaults extends Instance {
 	};
 }
 
+interface RetryDescriptor {
+	// @todo Unsure of which types `retry` can be.
+	// Used `unknown` so anytime it'll be used, TS will complain and require a type assertion
+	// Maybe a maintainer can guide me here.
+	retries: (retry: unknown, error: Error) => number | number;
```
Magellol: My first guess is `retry` would be a `number` but I just needed some clarifications.

sindresorhus: You can just add a console.log to the code and check yourself whenever you wonder what the actual type is ;)

source/utils/types.ts
```diff
@@ -1,14 +1,19 @@
 import {IncomingMessage} from 'http';
 import {RequestOptions} from 'https';
+import {Readable} from 'stream';
```
sindresorhus: `import {Readable as ReadableStream} from 'stream';` to improve code readability.

source/utils/types.ts
sindresorhus: `number | number;`?

source/get-response.ts
sindresorhus: Correct. Good catch.

source/utils/types.ts
Magellol: I've removed the comment and fixed the `retry` type to be `number`.
The whole type is now `() => number | number` which I believe is accurate now? Let me know if it's not.

Magellol:
Build is now passing at this commit https://github.com/sindresorhus/got/pull/720/commits/d6600c7a8903a4772affe25bdc0754c9c203d211

My mistake was the intersection type between `Defaults` and `Options`. Instead of inferring the type from the actual value, I've created a dedicated type. Let me know if all is good 👍

source/as-stream.ts
sindresorhus: Yup: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/events.d.ts

Can you check if there's an issue in https://github.com/DefinitelyTyped/DefinitelyTyped, and if not, open one?

source/get-response.ts
```diff
@@ -1,8 +1,8 @@
 import {IncomingMessage} from 'http';
 import EventEmitter from 'events';
-import {Transform, PassThrough} from 'stream';
+import {Transform} from 'stream';
```
sindresorhus: `Transform` => `Transform as TransformStream`

source/utils/types.ts
```diff
@@ -59,10 +73,16 @@ export interface Options extends RequestOptions {
 	decompress?: boolean;
 	encoding?: BufferEncoding | null;
 	method?: Method;
+	retry?: RetryDescriptor;
+	throwHttpErrors?: boolean;
 	// TODO: Remove this once TS migration is complete and all options are defined.
 	[key: string]: unknown;
 }
 
+export interface RequestEmitter extends EventEmitter {
```
sindresorhus: Isn't there a built-in type we can use for this?

source/utils/types.ts
```diff
@@ -59,10 +73,16 @@ export interface Options extends RequestOptions {
 	decompress?: boolean;
 	encoding?: BufferEncoding | null;
 	method?: Method;
+	retry?: RetryDescriptor;
```
```suggestion
	retry?: number | RetryDescriptor;
```

https://github.com/sindresorhus/got#retry

source/utils/types.ts
```diff
@@ -47,8 +49,20 @@ export interface InterfaceWithDefaults extends Instance {
 	};
 }
 
+interface RetryDescriptor {
+	retries: ((retry: number, error: Error) => number) | number;
+	methods: Method[];
+	statusCodes: number[];
+	errorCodes: string[];
+}
```
sindresorhus: Make sure it has everything from the docs: https://github.com/sindresorhus/got#retry

source/utils/types.ts
```diff
@@ -47,8 +49,20 @@ export interface InterfaceWithDefaults extends Instance {
 	};
 }
 
+interface RetryDescriptor {
+	retries: ((retry: number, error: Error) => number) | number;
+	methods: Method[];
+	statusCodes: number[];
```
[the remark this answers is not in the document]
sindresorhus: Example: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4a98006b062812c2cf10d267b1a3769c2e0c23e5/types/got/index.d.ts#L247

Magellol:
Thanks for the feedback. Will make changes as soon as I can.

source/utils/types.ts
Magellol: You're right, there is. See https://github.com/sindresorhus/got/pull/720/commits/775684bb9fe153bbf641d8a90993cede216f7c0e for the change.

Magellol:
@sindresorhus I believe I have adressed all requested changes. I will look into https://github.com/sindresorhus/got/pull/720#discussion_r253270974 asap.

source/as-stream.ts
```diff
@@ -1,10 +1,13 @@
 'use strict';
```
szmarczak: remove this

Magellol: Done. See 3e88c36 👍

sindresorhus:
Don't forget to do https://github.com/sindresorhus/got/pull/720#discussion_r253345815.

sindresorhus:
Nice work, @Magellol 🙌

Magellol:
@sindresorhus Sounds good. Thanks!
Yeah, it slipped my mind. I haven't seen an issue _yet_ but I just glanced, I'll look more deeply.

PR #721, opened 2019-02-02
Rewrite `create.js` in TypeScript

michaeldera:
#### After npm run build
I have one issue I encountered I am not really sure how to solve on line 67 and 68 of [create.ts](source/create.ts). It is generating the following errors:

- Element implicitly has 'any' type because type '{ (url: URL, options: Options): any; create: (defaults: any ) => ...; has no index signature, and 
- Element implicitly has 'any' type because type '( url: URL, options: Options) => any' has no index signature.

I am not sure how to approach the issue... however if, instead of using the `for (const method of aliases){` on line 66 and I assign each alias individually, i.e `got.put = (url: Url ....` the problem goes away but I presume there is a better way

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -3,7 +3,7 @@ import {URL} from 'url';
 import errors from './errors';
 import asStream from './as-stream';
 import asPromise from './as-promise';
-import {normalize, preNormalize} from './normalize-arguments';
+import { normalize, preNormalize } from './normalize-arguments';
 import merge, { mergeOptions, mergeInstances} from './merge';
 import deepFreeze from './utils/deep-freeze';
 import { InterfaceWithDefaults, Method, Options, NextFunction } from './utils/types';
--- a/source/create.ts
+++ b/source/create.ts
@@ -1,14 +1,16 @@
 'use strict';
+import {URL} from 'url';
 import * as errors from './errors';
 import * as  asStream  from './as-stream';
 import * as asPromise from './as-promise';
 import * as normalizeArguments from './normalize-arguments';
 import merge, { mergeOptions, mergeInstances} from './merge';
 import deepFreeze from './utils/deep-freeze';
+import { InterfaceWithDefaults, Method, Options, NextFunction } from './utils/types';
 
 const getPromiseOrStream = ( options : any ) => options.stream ? asStream(options) : asPromise(options);
 
-const aliases :  = [
+const aliases : Method[]  = [
 	'get',
 	'post',
 	'put',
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you do the first commit as just renaming `.js` to `.ts` using `git mv`? That way GitHub can show an actual diff between the files. Right now it's impossible to review the differences.

source/create.ts
```diff
@@ -0,0 +1,82 @@
+'use strict';
+import {URL} from 'url';
```
sindresorhus: Don't import `URL`. We use the `URL` global.

source/create.ts
```diff
@@ -0,0 +1,82 @@
...
+import errors from './errors';
+import asStream from './as-stream';
+import asPromise from './as-promise';
+import { normalize, preNormalize } from './normalize-arguments';
+import merge, { mergeOptions, mergeInstances} from './merge';
+import deepFreeze from './utils/deep-freeze';
+import { InterfaceWithDefaults, Method, Options, NextFunction } from './utils/types';
+
+const getPromiseOrStream = ( options : any ) => options.stream ? asStream(options) : asPromise(options);
```
sindresorhus: Don't use `any`. Type it properly.

source/create.ts
```diff
@@ -0,0 +1,82 @@
...
+	defaults = merge({}, defaults);
+	preNormalize(defaults.options);
+
+	if (!defaults.handler) {
+		// This can't be getPromiseOrStream, because when merging
+		// the chain would stop at this point and no further handlers would be called.
+		defaults.handler = (options: any, next: NextFunction) => next(options);
+	}
+
+	function got(url: URL, options: Options) {
```
sindresorhus: Got also supports a `string` in the `url` parameter.

michaeldera:
Alright. Let me redo the PR.

sindresorhus:
There's no need to close or open a new PR. You can just push additional commits to this PR ;)

source/create.ts
szmarczak: @sindresorhus Does TypeScript resolve it automatically if using Node 8?

michaeldera:
> 
> 
> There's no need to close or open a new PR. You can just push additional commits to this PR ;)

Please forgive my ignorance... the first commit, where I am supposed to do `git mv` so it is possible to see the diff, that is the part I cannot get around adding that commit... tried rebasing but not quite getting that bit right

szmarczak:
@michaeldera Ping. Are you finishing this PR? Otherwise, I could work on it :)

michaeldera:
> @michaeldera Ping. Are you finishing this PR? Otherwise, I could work on it :)

If it is not too far off I can work on it over the weekend but you could work on it if you would like.

szmarczak:
> If it is not too far off I can work on it over the weekend but you could work on it if you would like.

Sure! I'm glad I can help :) At the moment I'm focused on my other Got PRs - gotta fix them.

michaeldera:
Great! That works! If I am able to do it before I certainly will :)

source/create.ts
sindresorhus: Never mind. I got confused with Ky.

sindresorhus:
> Please forgive my ignorance... the first commit, where I am supposed to do `git mv` so it is possible to see the diff, that is the part I cannot get around adding that commit... tried rebasing but not quite getting that bit right

Copy-paste your changes somewhere as a backup, create a new branch from `master`, rename the `.js` file to `.ts`, commit this change, then copy-paste back your changes to the file, now commit these changes as a new commit, then force push to this remote PR branch.

michaeldera:
Doing this over now

sindresorhus:
No need to create a new PR. Force push is your friend.

sindresorhus:
@michaeldera Still interested in finishing this?

michaeldera:
I am sorry I will not be able to right now.

PR #729, opened 2019-02-10
Fix `port` not being reset on redirect

### LINKED ISSUE #719: Reset of request options on redirect

I have a local server running at `localhost:5002` which then does a redirect to an external site. However, when I use `got.stream` to make the request to `localhost:5002` and follow redirects, it also tries to use port 5002 for the external site, which is incorrect and results in a connection timeout. 

This problem can be easily worked around using the `on('redirect')` event handler for `got.stream`:

```
      await got.stream('http://localhost:5002/test')
        .on('redirect', (_, nextOptions) => {
          nextOptions.port = null;
        })
```

I think the `beforeRedirect` hooks would be an equally effective solution in this case as well.

Perhaps this is how `got` is intended to be used in this case, but I didn't see any documentation mentioning it. Also, it was not immediately obvious to me that this was happening (I may just be dense) until I used a debugger to step through the `got` code. 

I see that `got` already calls [`urlToOptions`](https://github.com/sindresorhus/got/blob/master/source/utils/url-to-options.ts) to update the request options before making the redirect request. However, in my case, the `port` value from the first request does not get overwritten because the redirect URL does not specify a port, thus there is no `port` property extracted by the spread operation on `urlToOptions(redirectURL)`. 

It seems like it might be worth making a small change to reset properties like `port` that may not exist on the redirect URL to `null` when building the redirect options, since it seems like the code is already trying to make following redirects transparent to the end user. If that sounds good, I can put together a PR.

Another option would just be to add documentation regarding this possible issue and how to work around it.

#### Checklist

szmarczak:
Fixes #719

### DIFF
--- a/test/redirects.js
+++ b/test/redirects.js
@@ -38,6 +38,13 @@ test.before('setup', async () => {
 		response.end();
 	});
 
+	http.on('/to-port-80', (request, response) => {
+		response.writeHead(307, {
+			location: 'http://localhost'
+		});
+		response.end();
+	});
+
 	http.on('/utf8-url-áé', reached);
 	http.on('/?test=it’s+ok', reached);
 
@@ -218,4 +225,17 @@ test('throws on malformed redirect URI', async t => {
 test('throws on invalid redirect URL', async t => {
 	const error = await t.throwsAsync(got(`${http.url}/invalidRedirect`));
 	t.is(error.code, 'ERR_INVALID_URL');
+});
+
+test('port is reset on redirect', async t => {
+	const s = await createServer();
+	s.on('/', (request, response) => {
+		response.end('ok');
+	});
+	await s.listen(80);
+
+	const {body} = await got(`${http.url}/to-port-80`);
+	t.is(body, 'ok');
+
+	await s.close();
 });
--- a/test/redirects.js
+++ b/test/redirects.js
@@ -1,5 +1,6 @@
 import {URL} from 'url';
 import test from 'ava';
+import nock from 'nock';
 import got from '../source';
 import {createServer, createSSLServer} from './helpers/server';
 
@@ -34,13 +35,6 @@ test.before('setup', async () => {
 	http.on('/finite', (request, response) => {
 		response.writeHead(302, {
 			location: `${http.url}/`
-		});
-		response.end();
-	});
-
-	http.on('/to-port-80', (request, response) => {
-		response.writeHead(307, {
-			location: 'http://localhost'
 		});
 		response.end();
 	});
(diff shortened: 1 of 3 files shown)


szmarczak:
Why does Travis need port 80? 🤔

sindresorhus:
> If you need web server to be listening on port 80, remember to use sudo (Linux will not allow non-privileged process to bind to port 80). For ports greater than 1024, using sudo is not necessary (and not recommended).
> \- https://docs.travis-ci.com/user/gui-and-headless-browsers/#starting-a-web-server

test/redirects.js
```diff
@@ -38,6 +38,13 @@ test.before('setup', async () => {
 		response.end();
 	});
 
+	http.on('/to-port-80', (request, response) => {
```
sindresorhus:
```suggestion
	http.on('/redirect-to-port-80', (request, response) => {
```

sindresorhus: And why not keep this endpoint inside the test itself?

szmarczak: Well, if we want to move the route into the specific test, then we should do it everywhere, not only here.

Also, that would result in having two servers in one test function.

szmarczak:
> > If you need web server to be listening on port 80, remember to use sudo (Linux will not allow non-privileged process to bind to port 80). For ports greater than 1024, using sudo is not necessary (and not recommended).
> > - [docs.travis-ci.com/user/gui-and-headless-browsers/#starting-a-web-server](https://docs.travis-ci.com/user/gui-and-headless-browsers/#starting-a-web-server)

So I guess it should be `sudo nyc ava`?

test/redirects.js
sindresorhus: > Well, if we want to move the route into the specific test, then we should do it everywhere, not only here.

Yes, but that's an insane amount of work. I think we should at least do so for new tests. Better to do it iteratively.

> Also, that would result in having two servers in one test function.

Why is that a problem?

sindresorhus:
Can't you pick a different port than `80`?

sindresorhus:
> So I guess it should be `sudo nyc ava`?

Then we would have to have `sudo` enabled on Travis too, which means we couldn't use their fast Docker images, which would make Travis much slower...

test/redirects.js
szmarczak: > Why is that a problem?

The code becomes less readable.

> Yes, but that's an insane amount of work. I think we should at least do so for new tests. Better to do it iteratively.

Ok so.

szmarczak:
> Can't you pick a different port than 80?

Yeah, I can. `443`. The port has to be default one.

Or... we could check the error if there's nothing listening on 80...

szmarczak:
> Then we would have to have sudo enabled on Travis too, which means we couldn't use their fast Docker images, which would make Travis much slower...

IMO using sudo shouldn't affect the performance... But I believe you're right :)

sindresorhus:
> IMO using sudo shouldn't affect the performance...

It's not "using sudo" that makes it slow. It's just that the Docker images on Travis don't support sudo, so when you enable sudo support there it will use the old non-Docker images, which are slower.

test/redirects.js
sindresorhus: > The code becomes less readable.

I would actually argue for the opposite. By having all the related code in the test you can fully understand what it does. If the test calls the server code, you need to scroll up or look in another file to see what the server part of the test does. Also, having shared tests makes it very easy to reuse endpoints for different tests, which means if you want to change the logic of one test and endpoint, you need to change many tests.

szmarczak:
> when you enable sudo support there it will use the old non-Docker images, which are slower.

Thanks for explaination, makes sense.

szmarczak:
We could also use `nock` there...

sindresorhus:
If `443` works, I would prefer that. Otherwise, `nock` is ok.

szmarczak:
> If `443` works, I would prefer that.

But `443` is below `1080`, so it won't work.

> Otherwise, `nock` is ok.

Ok, I'll do it :)

PR #731, opened 2019-02-19
Add DNS cache

### LINKED ISSUE #661: DNS cache
I do understand that mabe it's not the purpose of this lib (feel free to close in case).
But would be useful adding a dns cache layer.
It's one of the common problem nowadays with huge number of requests. It doesn't cache dns lookups or anything. Maybe it's worth just mentioning this in Readme.

What's the cleanest way to deal with it?

szmarczak:
Fixes #661 

#### Checklist

### DIFF
--- a/source/as-stream.ts
+++ b/source/as-stream.ts
@@ -76,13 +76,13 @@ export default function asStream(options: MergedOptions) {
 			throw new Error('Failed to pipe. The response has been emitted already.');
 		}
 
-		const result = pipe(destination, options);
+		pipe(destination, options);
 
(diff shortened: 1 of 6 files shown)


source/as-stream.ts
```diff
@@ -76,13 +76,13 @@ export default function asStream(options: MergedOptions) {
 			throw new Error('Failed to pipe. The response has been emitted already.');
 		}
 
-		const result = pipe(destination, options);
+		pipe(destination, options);
```
szmarczak: The behavior is the same, this just fixes some some linting error :P

PR #738, opened 2019-02-22
Rewrite `as-promise.js` in TypeScript

kevcenteno:
#### Changes
Rewrites as-promise.js in TypeScript

#700 

#### Checklist

### DIFF
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -1,35 +1,41 @@
-'use strict';
-const EventEmitter = require('events');
-const getStream = require('get-stream');
-const is = require('@sindresorhus/is');
-const PCancelable = require('p-cancelable');
-const requestAsEventEmitter = require('./request-as-event-emitter');
-const {HTTPError, ParseError, ReadError} = require('./errors');
-const {mergeOptions} = require('./merge');
-const {reNormalize} = require('./normalize-arguments');
+import {ClientRequest, IncomingMessage} from 'http';
+import EventEmitter from 'events';
+import getStream from 'get-stream';
+import is from '@sindresorhus/is';
+import PCancelable from 'p-cancelable';
+import requestAsEventEmitter from './request-as-event-emitter';
+import {HTTPError, ParseError, ReadError} from './errors';
+import {mergeOptions} from './merge';
+import {reNormalize} from './normalize-arguments';
+import {CancelableRequest, Options, Response} from './utils/types';
 
-const asPromise = options => {
+// TODO: Remove once request-as-event-emitter is converted to TypeScript
+interface RequestAsEventEmitter extends ClientRequest {
+	retry: (error: Error) => boolean;
+}
+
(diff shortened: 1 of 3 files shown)


source/as-promise.ts
```diff
@@ -44,21 +51,26 @@ const asPromise = options => {
...
-						}));
-
-						// Remove any further hooks for that request, because we we'll call them anyway.
-						// The loop continues. We don't want duplicates (asPromise recursion).
-						updatedOptions.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
-
-						return asPromise(updatedOptions);
-					});
+				if (options.hooks && options.hooks.afterResponse) {
+					for (const [index, value] of Object.entries(options.hooks.afterResponse).entries()) {
```
sindresorhus: You're right that the statement was incorrect previously, but yours doesn't look correct either. I think it should be:

```suggestion
					for (const [index, hook] of options.hooks.afterResponse.entries()) {
```

source/as-promise.ts
```diff
@@ -44,21 +51,26 @@ const asPromise = options => {
...
-							throwHttpErrors: false
-						}));
-
-						// Remove any further hooks for that request, because we we'll call them anyway.
-						// The loop continues. We don't want duplicates (asPromise recursion).
-						updatedOptions.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
-
-						return asPromise(updatedOptions);
-					});
+				if (options.hooks && options.hooks.afterResponse) {
```
sindresorhus: I would prefer using `!` (non-null assertion operator) instead of guarding it here, as we already guarantee its existence, so it's just the TS compiler that is struggling.

source/as-promise.ts
```diff
@@ -44,21 +51,26 @@ const asPromise = options => {
...
+						response = await hook(response, updatedOptions => {
+							updatedOptions = reNormalize(mergeOptions(options, {
+								...updatedOptions,
+								retry: 0,
+								throwHttpErrors: false
+							}));
+
+							// Remove any further hooks for that request, because we we'll call them anyway.
+							// The loop continues. We don't want duplicates (asPromise recursion).
+							if (updatedOptions.hooks && options.hooks && options.hooks.afterResponse) {
```
sindresorhus: Same as https://github.com/sindresorhus/got/pull/738/files#r259592816

source/as-promise.ts
```diff
@@ -1,35 +1,42 @@
 'use strict';
```
sindresorhus: This is not needed anymore.

kevcenteno:
@sindresorhus Force pushed the changes

sindresorhus:
No need to force push. It's easier to review when we can see what's changed. We squash on merge anyway.

sindresorhus:
Thank you for helping out, @kevcenteno 🙌

PR #756, opened 2019-03-15
Don't infer `POST` automatically when specifying `body`

szmarczak:
Throws instead of defaulting the method to POST when specified body without method.

#### Checklist


sindresorhus:
There's some docs that are not updated: https://github.com/sindresorhus/got#json-mode

Try grepping for `got(` and find the places using `json` or `body`. There might be more places in the code and other docs.

szmarczak:
Ah, forgot to update readme examples... Will do.

Issue #758, opened 2019-03-21
Make the TypeScript types strict and better


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24180%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/got/issues/758)

We have successfully transitioned to TypeScript, but we have not yet enabled strict mode. Although some of the files have strict types, there's still a lot of work to properly type everything.

**If you want to take on this issue, it's expected that you have deep TypeScript understanding and experience. It's not an easy issue.** Comment if you decide to start on this to prevent duplicate work.

I suggest the following order of fixing things (can be multiple PRs):

- [x] Remove `"strict": false,` in tsconfig.json and properly type what's required just to get it to compile (`$ npm test` should pass).
- [x] Improve existing typing and ensure everything has as specific and correct types are possible. (Some progress has been made: https://github.com/sindresorhus/got/pull/760)
- [x] Remove `noImplicitReturns` from tsconfig.json and fix any errors.
- [x] Remove `noUnusedParameters` from tsconfig.json and fix any errors.
- [x] Remove [these lines](https://github.com/sindresorhus/got/blob/0a79ccbe91cc6da1e2e0e41157173e32a5d18a16/package.json#L113-L115) in package.json and fix errors.
- [x] Fix type-related TODO comments.
- [x] Remove `// @ts-ignore` comments and fix the errors.
- [ ] Add doc comments to *public* (meaning exported) functions/types/interfaces. Borrow text from readme. This especially applies to the options and the main Got interface. Follow this style: https://github.com/sindresorhus/typescript-definition-style-guide#documentation
- [ ] Use `readonly` and const assertions wherever possible: https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $180.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($180.00)

### Submitted pull Requests
- [#797 src: Work on strict ts migrations](https://issuehunt.io/r/sindresorhus/got/pull/797)
- [#928 Make TypeScript types conforms to strict mode](https://issuehunt.io/r/sindresorhus/got/pull/928)
---

### Tips

[...]

PR #760, opened 2019-03-22
Improve typings relating to requestAsEventEmitter

paulmelnikow:
This improves several of the types in `requestAsEventEmitter`, including its own return type and its dependencies. It also clears several TypeScript warnings.

There are several changes which convert

```ts
import http from 'http'
```

to

```ts
import * as fs from 'fs'
```

This is to clear warnings which look like this:

```
source/utils/get-body-size.ts:1:8 - error TS1192: Module '"fs"' has no default export.

1 import fs from 'fs';
```

My workflow has primarily to run `./node_modules/.bin/tsc --noEmit source/request-as-event-emitter.ts` and then the tests. I'm not sure if there is a better way to evaluate the changes to the types.

This is my first PR on got. I'm not an expert in TypeScript and there are definitely some good challenges here! Looking forward to feedback.

#### Checklist

Ref: #758

### DIFF
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -21,7 +21,7 @@ const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308])
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);
 
 export interface RequestAsEventEmitter extends EventEmitter {
-	retry: (error: Error) => (boolean | undefined);
+	retry: (error: Error) => boolean;
 	abort: () => void;
 }
 
@@ -252,7 +252,7 @@ export default (options, input?: TransformStream) => {
 		}
 	};
 
-	emitter.retry = (error: Error): (boolean | undefined) => {
+	emitter.retry = (error: Error): boolean => {
 		let backoff: number;
 
 		try {
@@ -260,7 +260,7 @@ export default (options, input?: TransformStream) => {
 			backoff = ((options.retry as RetryOption).retries as RetryFunction)(++retryCount, error);
 		} catch (error2) {
 			emitError(error2);
-			return;
+			return false;
 		}
 
 		if (backoff) {
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -1,10 +1,9 @@
-import {format as urlFormat, URL, URLSearchParams} from 'url'; // TODO: Use the `URL` global when targeting Node.js 10
-import * as util from 'util';
-import {EventEmitter} from 'events';
+import urlLib, {URL, URLSearchParams} from 'url'; // TODO: Use the `URL` global when targeting Node.js 10
+import util from 'util';
+import EventEmitter from 'events';
 import {Transform as TransformStream} from 'stream';
-import * as http from 'http';
-import * as https from 'https';
-// TODO: This causs a TypeScript warning, but using `import * as CacheableRequest` produces a runtime error.
+import http from 'http';
+import https from 'https';
 import CacheableRequest from 'cacheable-request';
 import toReadableStream from 'to-readable-stream';
 import is from '@sindresorhus/is';
@@ -20,11 +19,6 @@ import {RequestFn, Options, Delays, RetryFn, RetryOption} from './utils/types';
 
 const getMethodRedirectCodes = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
 const allMethodRedirectCodes = new Set([300, 303, 307, 308]);
-
-type ProcessVersions = typeof process.versions
-interface ProcessVersionsWithElectron extends ProcessVersions {
-	electron: string;
-}
 
 export interface RequestAsEventEmitter extends EventEmitter {
 	retry: (error: Error) => (boolean | undefined);
@@ -80,7 +74,7 @@ export default (options, input?: TransformStream) => {
 		}
 
 		/* istanbul ignore next: electron.net is broken */
-		if (options.useElectronNet && (process.versions as ProcessVersionsWithElectron).electron) {
+		if (options.useElectronNet && (process.versions as any).electron) {
 			// @ts-ignore
 			const r = ({x: require})['yx'.slice(1)]; // Trick webpack
 			const electron = r('electron');
@@ -352,7 +346,7 @@ export default (options, input?: TransformStream) => {
 				options.headers.accept = 'application/json';
 			}
 
-			requestUrl = options.href || (new URL(options.path, urlFormat(options))).toString();
+			requestUrl = options.href || (new URL(options.path, urlLib.format(options))).toString();
 
 			await get(options);
 		} catch (error) {
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -74,6 +74,8 @@ export default (options, input?: TransformStream) => {
 		}
 
 		/* istanbul ignore next: electron.net is broken */
+		// No point in typing process.versions correctly, as
+		// process.version.electron is used only once, right here.
 		if (options.useElectronNet && (process.versions as any).electron) {
 			// @ts-ignore
 			const r = ({x: require})['yx'.slice(1)]; // Trick webpack
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -15,7 +15,7 @@ import getResponse from './get-response';
 import {uploadProgress} from './progress';
 import {CacheError, UnsupportedProtocolError, MaxRedirectsError, RequestError, TimeoutError} from './errors';
 import urlToOptions from './utils/url-to-options';
-import {RequestFn, Options, Delays, RetryFn, RetryOption} from './utils/types';
+import {RequestFunction, Options, Delays, RetryFunction, RetryOption} from './utils/types';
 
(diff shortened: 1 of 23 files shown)


source/utils/types.ts
```diff
@@ -50,19 +51,34 @@ export interface InterfaceWithDefaults extends Instance {
...
+	connect?: number;
+	secureConnect?: number;
+	socket?: number;
+	response?: number;
+	send?: number;
+	request?: number;
+}
+
+// The library overrides the type definition of `agent`.
+export interface Options extends Pick<https.RequestOptions, Exclude<keyof https.RequestOptions, 'agent'>> {
```
paulmelnikow: `Options` is a particularly tricky case.

One challenge is that got supports options which have slightly different types than `https.RequestOptions`, which causes it to be type-wise incompatible. Using Pick/Exclude here seems like a partial solution for the first issue, though it requires coercion back to `https.RequestOptions` in `request-as-event-emitter.ts`.

Another is that `preNormalizeArguments` and `normalizeOptions` narrow down the option types. I haven't addressed that yet.

source/utils/types.ts
```diff
@@ -77,6 +93,10 @@ export interface Options extends RequestOptions {
 	method?: Method;
 	retry?: number | RetryOption;
 	throwHttpErrors?: boolean;
+	cookieJar?: CookieJar;
+	request?: RequestFn;
+	agent: https.Agent | boolean | { [key: string]: https.Agent };
+	gotTimeout?: number | Delays;
```
paulmelnikow: I tried adding `cache?: Store<any>` here, though the linter then required pulling in `keyv` as a dependency. I think a better solution would be to export Store as a typealias from the typings for CacheableRequest.

There's another issue with the typings for CacheableRequest, related to whether or not it's a default export. I added a todo comment. I suspect a change is needed to the typings, though I'm afraid it's a bit over my head.

sindresorhus:
> There are several changes which convert

You should not need to do changes like these. We have the `allowSyntheticDefaultImports` config enabled. Seems you're not picking up our tsconfig for some reason.

source/utils/types.ts
sindresorhus: Would be better to use `Omit` from https://github.com/sindresorhus/type-fest#utilities

source/utils/types.ts
sindresorhus: > I think a better solution would be to export Store as a typealias from the typings for CacheableRequest.

👍

source/request-as-event-emitter.ts
```diff
@@ -68,11 +80,11 @@ export default (options, input?: TransformStream) => {
 		}
 
 		/* istanbul ignore next: electron.net is broken */
-		if (options.useElectronNet && (process.versions as any).electron) {
+		if (options.useElectronNet && (process.versions as ProcessVersionsWithElectron).electron) {
```
sindresorhus: No point in typing this properly. It's only used once internally here. I think in this case using `any` is ok.

paulmelnikow:
> You should not need to do changes like these. We have the `allowSyntheticDefaultImports` config enabled. Seems you're not picking up our tsconfig for some reason.

That makes sense; I'll revert them. Though it's odd that `tsc` isn't picking up the tsconfig correctly.

source/utils/types.ts
paulmelnikow: I opened DefinitelyTyped/DefinitelyTyped#34155 though it's not blocking this.

source/request-as-event-emitter.ts
```diff
@@ -231,21 +242,22 @@ export default (options, input?: TransformStream) => {
...
 			} catch (error) {
 				emitError(new RequestError(error, options));
 			}
 		}
 	};
 
-	// TODO: Properly type this
-	(emitter as any).retry = error => {
-		let backoff;
+	emitter.retry = (error: Error): (boolean | undefined) => {
```
sindresorhus: Can we get this to only return a boolean?

source/utils/types.ts
```diff
@@ -50,19 +52,34 @@ export interface InterfaceWithDefaults extends Instance {
 	};
 }
 
-interface RetryOption {
-	retries?: ((retry: number, error: Error) => number) | number;
+export type RetryFn = (retry: number, error: Error) => number;
```
sindresorhus: `RetryFn` => `RetryFunction`

source/utils/types.ts
```diff
@@ -50,19 +52,34 @@ export interface InterfaceWithDefaults extends Instance {
...
+
+export interface RetryOption {
+	retries?: RetryFn | number;
 	methods?: Method[];
 	statusCodes?: StatusCode[];
 	maxRetryAfter?: number;
 	errorCodes?: ErrorCode[];
 }
 
+export type RequestFn = typeof https.request;
```
sindresorhus: `RequestFn` => `RequestFunction`

source/utils/types.ts
```diff
@@ -50,19 +52,34 @@ export interface InterfaceWithDefaults extends Instance {
...
+	connect?: number;
+	secureConnect?: number;
+	socket?: number;
+	response?: number;
+	send?: number;
+	request?: number;
+}
+
+// The library overrides the type definition of `agent`.
+export interface Options extends Omit<https.RequestOptions, 'agent'> {
```
sindresorhus: Do you really have to omit `agent` here? I thought extended interfaces could override the parent one.

source/utils/types.ts
```diff
@@ -77,6 +94,10 @@ export interface Options extends RequestOptions {
 	method?: Method;
 	retry?: number | RetryOption;
 	throwHttpErrors?: boolean;
+	cookieJar?: CookieJar;
+	request?: RequestFn;
+	agent: https.Agent | boolean | { [key: string]: https.Agent };
```
sindresorhus: What about `http.Agent`? Or is it exactly the same?

Instead of `[key: string]`, you can just explicitly list the keys. There are only two.

source/utils/types.ts
paulmelnikow: I think that would work if `agent` were being overridden with a type that was more specific / restrictive than `https.RequestOptions`. And then, `Options` would still be a subtype of `http.RequestOptions`.

However, the override is more permissive, so I have to drop the old `agent` and define a new one. Instances of the new `Options` type no longer conform to `https.RequestOptions`.

source/utils/types.ts
paulmelnikow: > What about `http.Agent`? Or is it exactly the same?

`https.Agent` extends `http.Agent`. Its constructor takes a few additional options.

> Instead of `[key: string]`, you can just explicitly list the keys. There are only two.

Ah, sure, I'll update that. I wasn't sure if there were other protocols the library might support.

source/utils/types.ts
sindresorhus: Makes sense. Thanks for explaining.

source/utils/types.ts
paulmelnikow: Merged and updated.

source/request-as-event-emitter.ts
```diff
@@ -233,24 +243,25 @@ export default (options, input?: TransformStream) => {
 
 			cacheRequest.once('request', handleRequest);
 		} else {
-			// Catches errors thrown by calling fn.request(...)
+			// Catches errors thrown by calling requestFn(...)
```
szmarczak: It's `requestFn.request` if you do it this way. IMO it should be `fn.request` as before, no need to change.

source/request-as-event-emitter.ts
```diff
@@ -70,11 +76,13 @@ export default (options, input?: TransformStream) => {
...
 
 		/* istanbul ignore next: electron.net is broken */
+		// No point in typing process.versions correctly, as
+		// process.version.electron is used only once, right here.
 		if (options.useElectronNet && (process.versions as any).electron) {
 			// @ts-ignore
 			const r = ({x: require})['yx'.slice(1)]; // Trick webpack
 			const electron = r('electron');
-			fn = electron.net || electron.remote.net;
+			requestFn = electron.net || electron.remote.net;
```
paulmelnikow: Are there tests for this code? I realize this was not accessing `electron.net.request` as it should.

source/request-as-event-emitter.ts
paulmelnikow: While the old `fn` was an object with a `request` property, the new `requestFn` is the request function itself. Typing the `fn` object would look like

```ts
interface ObjWithRequestFunction {
  request: RequestFunction
}
```

and would still require creating a duck-typed object with a `request` property:

https://github.com/sindresorhus/got/pull/760/files#diff-267e4380ccf0cfebc8b885a948acf226L62

All of that seems unnecessarily indirect.

The diff is a bit confusing, mostly because `fn` was a confusing name, though the new code seems clearer and more direct.

szmarczak: Ah. After that commit it makes sense now

source/request-as-event-emitter.ts
sindresorhus: No, too much overhead to include Electron testing to be worth it. We do manual test after doing changes to that code.

sindresorhus:
@paulmelnikow Is this good to merge or do you intend to do more changes?

paulmelnikow:
Good to merge!

sindresorhus:
Thank you! 🙏

paulmelnikow:
Thanks for all the comments!

PR #763, opened 2019-03-26
Enable the `noImplicitReturns` TypeScript setting

paulmelnikow:
This removes the two remaining implicit returns and removes `noImplicitReturns: false` from `tsconfig.json`. That checks off one of the items in #758.

Merging is blocked on #760, which removes another implicit return.

#### Checklist

### DIFF
--- a/source/utils/timed-out.ts
+++ b/source/utils/timed-out.ts
@@ -31,7 +31,7 @@ export interface Delays {
 export default (request: ClientRequest, delays: Delays, options: any) => {
 	/* istanbul ignore next: this makes sure timed-out isn't called twice */
 	if (Reflect.has(request, reentry)) {
-		return;
+		return noop;
 	}
 
 	(request as any)[reentry] = true;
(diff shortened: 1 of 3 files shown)


source/utils/timed-out.ts
```diff
@@ -31,7 +31,7 @@ export interface Delays {
 export default (request: ClientRequest, delays: Delays, options: any) => {
 	/* istanbul ignore next: this makes sure timed-out isn't called twice */
 	if (Reflect.has(request, reentry)) {
-		return;
+		return noop;
```
paulmelnikow: Returning a no-op function certainly is safer than returning something that is not callable! 😀 

It might be even safer to find a way to return a second cancellation handle for the same timeout. I'm not sure it's worth it, or for that matter really clear on the circumstances that could lead here.

sindresorhus: Returning a no-op function is fine.

PR #764, opened 2019-03-28
Enable the 'noUnusedParameters' TypeScript setting

paulmelnikow:
This removes `noUnusedParameters: false`, one of the tasks in the checklist at #758.

After reading these documents, it seems like either `_` or e.g. `_unusedRequest` is a good replacement name which silences the errors for these parameters

- https://github.com/Microsoft/TypeScript/issues/24249
- https://github.com/Microsoft/TypeScript/issues/9458

Looking forward to comments, and curious if you have another way in mind.

#### Checklist

### DIFF
--- a/test/agent.ts
+++ b/test/agent.ts
@@ -14,11 +14,11 @@ test.before('setup', async () => {
 
 	// HTTPS Handlers
 
-	https.on('/', (request, response) => {
+	https.on('/', (_unusedRequest, response) => {
 		response.end('https');
 	});
 
-	https.on('/httpsToHttp', (request, response) => {
+	https.on('/httpsToHttp', (_unusedRequest, response) => {
 		response.writeHead(302, {
 			location: http.url
 		});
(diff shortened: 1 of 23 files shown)


test/agent.ts
```diff
@@ -14,11 +14,11 @@ test.before('setup', async () => {
 
 	// HTTPS Handlers
 
-	https.on('/', (request, response) => {
+	https.on('/', (_unusedRequest, response) => {
```
sindresorhus: I think just `_request` would be better.

paulmelnikow: Sure, I'll change it if you'd like. I went with this because with `_request`, it seems likely that if in the future the code were changed and _did_ start to use the parameter, the programmer likely would not know to remove the underscore. Whereas if you start calling things on `_unusedRequest` it screams that the name should be changed.

sindresorhus:
Sorry, this was made moot by #765.

PR #765, opened 2019-03-30
Use preconfigured Got instance in tests

szmarczak:
#### TODO

- [x] Remove duplicated tests
- [x] Rename some tests 
- [x] Rebase tests to use newest features of AVA (e.g. using strings instead of RegExps)
- [x] Use preconfigured Got instance in tests
- [x] Changed routes to `/` for better readability (no weird names)
- [x] Prefix unused variables with an underscore

#### Checklist

*Sorry for force-pushes. It would be +100 commits if I didn't do that.*


source/normalize-arguments.ts
```diff
@@ -105,7 +105,7 @@ export const preNormalizeArguments = (options: any, defaults?: any) => {
 export const normalizeArguments = (url, options, defaults?: any) => {
 	if (is.plainObject(url)) {
 		options = {...url, ...options};
-		url = options.url || {};
+		url = options.url || '';
```
szmarczak: This needs testing. `''` if has `baseUrl`, otherwise `{}`.

szmarczak: It's an empty string, because merging URLs require two strings (`baseUrl` feature). So you can `customInstance({headers: ...})` without providing the `url` - it will use `baseUrl`. It would throw if the `url` was `{}`.

szmarczak: Tested :D

test/cache.ts
```diff
@@ -127,26 +99,39 @@ test('Cached response should have got options', async t => {
...
+	const {statusCode, body} = await got({cache, throwHttpErrors: false});
 	t.is(statusCode, 200);
 	t.deepEqual(body, 'ok');
 });
 
-test('DNS cache works', async t => {
+test('DNS cache works', withServer, async (t, _server, got) => {
 	const map = new Map();
-	await t.notThrowsAsync(got('https://example.com', {dnsCache: map}));
+	await t.notThrowsAsync(() => got('https://example.com', {dnsCache: map}));
```
sindresorhus: While AVA does support this, there's no point in doing it this way, as you can just pass the promise directly, like it did previously. Applies to all the places you changed this.

szmarczak: But if there's something wrong then `got()` could throw. But the test would fail anyway I think...

sindresorhus: `got()` should never throw. If it throws, it's a bug we should fix. It should **always** reject the promise. Promise-returning functions should never throw. `got.stream()` however, can throw.

szmarczak: Okay, doing that now :D

PR #767, opened 2019-04-02
Throw if request canceled and received incomplete response

### LINKED ISSUE #748: Got returning response before retries
#### Describe the bug

 - Node.js version:  v11.3.0
 - OS & version: MacOs v11.3.0

#### Actual behavior

got return response before retry

```
[Downloaded] http://www.alice-in-wonderland.net/wp-content/uploads/1book1.jpg
[Retry] http://www.alice-in-wonderland.net/wp-content/uploads/1book1.jpg
[Retry] http://www.alice-in-wonderland.net/wp-content/uploads/1book1.jpg
[Retry] http://www.alice-in-wonderland.net/wp-content/uploads/1book1.jpg
```

#### Expected behavior

got should do all retries before returning response

#### Code to reproduce

```
const got = require('got');
const fs = require('fs');

async function test() {
    const url = 'http://www.alice-in-wonderland.net/wp-content/uploads/1book1.jpg';
    const resp = await got(url, {
        timeout: {
            request: 1e3,
        },
        retry: 3,
        hooks: {
            beforeRetry: [(opt, err, retryN) => {
                console.log("[Retry]", opt.href)
            }]
        }
    });

    console.log('[Downloaded]', url);

    fs.createWriteStream('img').write(resp.body)
}

test();
```

#### Checklist

szmarczak:
Fixes #748

### DIFF
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -4,7 +4,7 @@ import getStream from 'get-stream';
 import is from '@sindresorhus/is';
 import PCancelable from 'p-cancelable';
 import requestAsEventEmitter from './request-as-event-emitter';
-import {HTTPError, ParseError, ReadError} from './errors';
+import {HTTPError, ParseError, ReadError, IncompleteResponseError} from './errors';
 import {mergeOptions} from './merge';
 import {reNormalizeArguments} from './normalize-arguments';
 import {CancelableRequest, Options, Response} from './utils/types';
@@ -38,6 +38,15 @@ export default function asPromise(options: Options) {
 			} catch (error) {
 				reject(new ReadError(error, options));
 				return;
+			}
+
+			if (!response.complete) {
+				if (response.req.aborted) {
+					// Canceled while downloading - will throw a CancelError or TimeoutError
+					return;
+				}
+
+				reject(new IncompleteResponseError(response, options));
 			}
 
 			const limitStatusCode = options.followRedirect ? 299 : 399;
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -4,7 +4,7 @@ import getStream from 'get-stream';
 import is from '@sindresorhus/is';
 import PCancelable from 'p-cancelable';
 import requestAsEventEmitter from './request-as-event-emitter';
-import {HTTPError, ParseError, ReadError, IncompleteResponseError} from './errors';
+import {HTTPError, ParseError, ReadError} from './errors';
 import {mergeOptions} from './merge';
 import {reNormalizeArguments} from './normalize-arguments';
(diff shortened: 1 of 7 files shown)


source/request-as-event-emitter.ts
```diff
@@ -120,6 +124,21 @@ export default (options, input?: TransformStream) => {
 					gotOptions: options
 				};
 
+				// Cached requests don't have the request object (so we can't do `request.aborted`)
+				// so we need to create our own `canceled` property.
```
szmarczak: Are cached requests cancelable? :confused:

source/as-promise.ts
```diff
@@ -40,6 +40,11 @@ export default function asPromise(options: Options) {
 				return;
 			}
 
+			if (response.canceled()) {
```
szmarczak: If the cached requests are not cancelable we could simply do `if (response.req && response.req.aborted) {`

source/as-stream.ts
```diff
@@ -23,7 +23,15 @@ export default function asStream(options: MergedOptions) {
 	const emitter = requestAsEventEmitter(options, input);
 
 	// Cancels the request
-	proxy._destroy = emitter.abort;
+	proxy._destroy = (error, callback) => {
+		// Stop transmitting data,
+		// so the stream won't have an `end` event.
+		output.destroy();
```
szmarczak: The more elegant way would be to unpipe the response.

szmarczak: We don't have to care about it. It will end if no error is provided, otherwise it will throw before ending.

source/as-promise.ts
```diff
@@ -40,6 +40,11 @@ export default function asPromise(options: Options) {
 				return;
 			}
 
+			if (response.req && response.req.aborted) {
+				// Canceled while downloading - will throw a CancelError or TimeoutError
```
tobenna:
```suggestion
				// Cancelled while downloading - will throw a CancelError or TimeoutError
```

test/helpers/slow-data-stream.ts
```diff
@@ -0,0 +1,18 @@
+import {PassThrough} from 'stream';
+
+export default () => {
```
tobenna: It'll be nice to add an explicit return type here so it's once less thing we have to add when making the typescript types better.
```suggestion
export default (): PassThrough => {
```

source/as-promise.ts
szmarczak: https://github.com/sindresorhus/p-cancelable/blob/master/readme.md#cancelable-vs-cancellable :smiley:

szmarczak:
Thanks for the review @tobenna 🎉

It's very nice that you've had some time to take a look at this :)

PR #768, opened 2019-04-02
Add `.isFromCache` property to the stream API and rename the Promise API property from `.fromCache` to `.isFromCache`

### LINKED ISSUE #747: Stream `fromCache` property
I need to `pipe()` the downloaded response for further processing (unzipping), but only when the response is not from cache. However, when using `got.stream()`, I can't figure out how to check where the response comes from - origin or cache.

```js
...
const response = got.stream(originUrl, { cache }); 
const fromCache = ...; // How to set this
if (fromCache) {
  response.pipe(unzip). ...
}
```
I've also tried
```js
const response = await got(originUrl, { cache });

if (response.fromCache) { // Here cache check is ok
  response.pipe(unzip). ...
}
```
but in this case response stream is already consumed.

Thanks

#### Checklist

szmarczak:
Fixes #747 

TODO: docs

### DIFF
--- a/source/as-stream.ts
+++ b/source/as-stream.ts
@@ -1,13 +1,17 @@
-import {PassThrough as PassThroughStream} from 'stream';
+import {PassThrough as PassThroughStream, Duplex as DuplexStream} from 'stream';
 import duplexer3 from 'duplexer3';
 import requestAsEventEmitter from './request-as-event-emitter';
 import {HTTPError, ReadError} from './errors';
 import {MergedOptions, Response} from './utils/types';
 
+export class ProxyStream extends DuplexStream {
+	fromCache?: boolean;
+}
+
 export default function asStream(options: MergedOptions) {
 	const input = new PassThroughStream();
 	const output = new PassThroughStream();
-	const proxy = duplexer3(input, output);
+	const proxy = duplexer3(input, output) as ProxyStream;
 	const piped = new Set();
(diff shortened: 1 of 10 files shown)


source/as-stream.ts
```diff
@@ -37,7 +42,14 @@ export default function asStream(options: MergedOptions) {
...
 
-		isFinished = true;
+		{
+			const read = proxy._read.bind(proxy);
+			proxy._read = (...args) => {
+				isFinished = true;
+
+				return read(...args);
+			};
+		}
```
szmarczak: It's a pity there is no `read` event...

szmarczak: After the first chunk is sent further `.pipe(...)` calls should throw.

Currently it throws if the `response` event was emitted.

sindresorhus:
I would name it `.isFromCache`.

Is there any point in making it `undefined` before the `request` event? Can't we just make it `false` then?

szmarczak:
> I would name it .isFromCache.

Sure. Should it be changed for Promises too?

> Is there any point in making it undefined before the request event? Can't we just make it false then?

It's undefined because we don't know is it true or false. I don't wanna make users mistaken.

sindresorhus:
Yes, let’s change the Promise one too (forgot we had it there).

sindresorhus:
👌

PR #772, opened 2019-04-10
Proxy headers from server request to Got

### LINKED ISSUE #743: Piping server request to Got should pipe headers too
In the [More about streams](https://github.com/sindresorhus/got/blob/master/migration-guides.md#more-about-streams) section, the `request` lib can also pipe the request headers, but looks like `got` won't do that, so the words could be more specific that got will only proxy response headers(at least for now) instead of both request and response headers.

szmarczak:
Fixes #743

### DIFF
--- a/source/as-stream.ts
+++ b/source/as-stream.ts
@@ -109,8 +109,8 @@ export default function asStream(options: MergedOptions) {
 	proxy.on('pipe', source => {
 		if (source instanceof IncomingMessage) {
 			options.headers = {
-				...options.headers,
-				...source.headers
+				...source.headers,
+				...options.headers
 			};
 		}
 	});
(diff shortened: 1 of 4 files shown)


source/as-stream.ts
```diff
@@ -105,6 +106,15 @@ export default function asStream(options: MergedOptions) {
 		return unpipe(stream);
 	};
 
+	proxy.on('pipe', source => {
+		if (source instanceof IncomingMessage) {
+			options.headers = {
+				...options.headers,
+				...source.headers
```
JasonBoy: I think the options.headers should take precedence over the default headers from incomingMessage

szmarczak: Good spot!

PR #776, opened 2019-04-11
Fix Webpack warnings

### LINKED ISSUE #742: Tricking webpack makes webpack spit out ugly warnings
I use `got 9.6.0` with `webpack 4.29.6` and target a node application. Webpack throws this ugly warning 
```
WARNING in ./node_modules/got/source/request-as-event-emitter.js 72:18-25
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```
The reason for this is the following tricky code in request-as-event-emitter.js line 42
``` js
  const r = ({x: require})['yx'.slice(1)]; // Trick webpack
  const electron = r('electron');
```
If I replace those lines with
``` js
  const electron = require('electron');
```
and configure webpack to treat 'electron' as an external dependency everything is fine without warnings.
Maybe this hack was necessary once, but it would nice to use something more standard that works out of the box.

szmarczak:
Fixes #742

### DIFF
--- a/source/utils/dynamic-require.ts
+++ b/source/utils/dynamic-require.ts
@@ -1,3 +1,3 @@
-export default (moduleObject: NodeModule, request: moduleId): unknown => {
+export default (moduleId: string, moduleObject: NodeModule = module): unknown => {
 	return moduleObject.require(moduleId);
 };
--- a/source/utils/dynamic-require.ts
+++ b/source/utils/dynamic-require.ts
@@ -1,0 +1,3 @@
+export default (mod: NodeModule, request: string): any => {
+	return mod.require(request);
+}
--- a/source/utils/dynamic-require.ts
+++ b/source/utils/dynamic-require.ts
@@ -1,3 +1,3 @@
-export default (mod: NodeModule, request: string): any => {
-	return mod.require(request);
+export default (moduleObject: NodeModule, request: moduleId): unknown => {
+	return moduleObject.require(moduleId);
 };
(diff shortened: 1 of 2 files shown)


sindresorhus:
mod => moduleObject

request => moduleId

any => unknown

sindresorhus:
Have you tested that this doesn't introduce any warnings when using Got with Webpack?

source/utils/dynamic-require.ts
```diff
@@ -0,0 +1,3 @@
+export default (moduleId: string, moduleObject: NodeModule = module): unknown => {
```
sindresorhus: While `= module` works here because it's called in the same package, it's not technically correct, as you're now requiring from the location of this file and not the callers. There should not be a default argument here and `module` should always be required to be passed.

szmarczak: I get it. Will fix.

szmarczak:
Tested, the error is gone! :D

I don't have much Webpack experience, so I must have imported it improperly (by copying the `dist` folder and the `node_modules` one - I don't know how to import it locally):

```js
WARNING in ./node_modules/keyv/src/index.js 18:14-40
Critical dependency: the request of a dependency is an expression
```

otherwise no errors :) On Chrome runs as expected, with minimal support :P

sindresorhus:
We should maybe move this logic to https://github.com/sindresorhus/require-fool-webpack so we can reuse it in other packages like `keyv` (per above).

PR #777, opened 2019-04-11
Add `context` option

### LINKED ISSUE #740: Storing request specific context in Got
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/740)

I would like to instrument a Got instance using `beforeRequest`, `afterResponse` and `beforeError` hooks. The plan is to create a new request context for each request in the `beforeRequest` hook and then access that context in the `afterResponse` and `beforeError` hooks to report the request's success/error status.

The natural way to do this would be to just mutate the options object in `beforeRequest` and set a new property containing the context, this is easy to access in both `afterResponse` and `beforeError` through the `gotOptions` given. My only worry is that by introducing new properties in the options there might be incompatibilities with other hooks or later versions of Got itself.

Are there any guidelines or rules of how to name custom variables inside options? Or alternatively is there any better place where I could store my custom property? I plan to release my instrumentation as a public library, so I just want to confirm that I'm not doing anything wrong here.

#### Checklist

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/got/issues/740).

szmarczak:
Fixes #740 

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -180,17 +180,22 @@ Type: `object`
 
 User data. In contrast to other options, `context` is not enumerable.
 
-Example:
+**Note:** The object is never merged, it's just passed through. Got will not modify the object in any way.
+
+It's very useful for storing auth tokens:
 
 ```js
 const got = require('got');
 
 const instance = got.extend({
 	hooks: {
-		afterResponse: [
-			response => {
-				const {context} = response.request.options;
-				context.latestFetchedSite = response.url;
+		beforeRequest: [
+			options => {
+				if (!options.context && !options.context.token) {
+					throw new Error('Token required');
+				}
+
+				options.header.token = options.context.token;
 			}
 		]
 	}
@@ -198,17 +203,13 @@ const instance = got.extend({
 
 (async () => {
 	const context = {
-		latestFetchedSite: ''
+		token: 'secret'
 	};
 
-	const response = await instance('https://example.com', {context});
+	const response = await instance('https://httpbin.org/headers', {context});
 
-	// Let's see our context data.
-	console.log(context.latestFetchedSite); //=> 'https://example.com/'
-
-	// The `context` won't be displayed here because it's not enumerable.
-	// You can, however, access it directly: `response.request.options.context`
-	console.log(response.request.options);
+	// Let's see the headers
+	console.log(response.body);
 })();
 ```
 
--- a/readme.md
+++ b/readme.md
@@ -176,7 +176,7 @@ JSON body. If the `Content-Type` header is not set, it will be set to `applicati
 
 ###### userData
 
-Type: `any`
+Type: `unknown`
 
 User data. In the opposite to other options, `userData` is not enumerable.
 
--- a/readme.md
+++ b/readme.md
@@ -203,10 +203,10 @@ const instance = got.extend({
 
 	const response = await instance('https://example.com', {context});
 
-	// Let's see our context data
+	// Let's see our context data.
 	console.log(context.latestFetchedSite); //=> 'https://example.com/'
 
-	// The `.context` won't be displayed here because it's not enumerable.
+	// The `context` won't be displayed here because it's not enumerable.
 	// You can, however, access it directly: `response.request.options.context`
 	console.log(response.request.options);
 })();
--- a/readme.md
+++ b/readme.md
@@ -176,7 +176,7 @@ JSON body. If the `Content-Type` header is not set, it will be set to `applicati
 
 ###### context
 
-Type: `unknown`
+Type: `object`
 
 User data. In contrast to other options, `context` is not enumerable.
 
--- a/readme.md
+++ b/readme.md
@@ -178,7 +178,9 @@ JSON body. If the `Content-Type` header is not set, it will be set to `applicati
 
 Type: `unknown`
 
-User data. In contrast to other options, `context` is not enumerable. Example:
+User data. In contrast to other options, `context` is not enumerable.
+
+Example:
 
 ```js
 const got = require('got');
@@ -201,11 +203,11 @@ const instance = got.extend({
 
 	const response = await instance('https://example.com', {context});
 
-	// Let's see our user data
-	console.log(context.latestFetchedSite); // => https://example.com/
+	// Let's see our context data
+	console.log(context.latestFetchedSite); //=> 'https://example.com/'
 
-	// `context` won't be displayed here, because it's not enumerable.
-	// Of course, you can access it in the following way: response.request.options.context
+	// The `.context` won't be displayed here because it's not enumerable.
+	// You can, however, access it directly: `response.request.options.context`
 	console.log(response.request.options);
 })();
 ```
--- a/readme.md
+++ b/readme.md
@@ -178,7 +178,7 @@ JSON body. If the `Content-Type` header is not set, it will be set to `applicati
 
 Type: `unknown`
 
-User data. In the opposite to other options, `userData` is not enumerable.
+User data. In contrast to other options, `userData` is not enumerable.
 
 ###### responseType
 
--- a/readme.md
+++ b/readme.md
@@ -174,6 +174,12 @@ Type: `object | Array | number | string | boolean | null` *(JSON-serializable va
 
 JSON body. If the `Content-Type` header is not set, it will be set to `application/json`.
 
+###### userData
+
+Type: `any`
+
+User data. In the opposite to other options, `userData` is not enumerable.
+
 ###### responseType
 
 Type: `string`<br>
--- a/readme.md
+++ b/readme.md
@@ -174,11 +174,41 @@ Type: `object | Array | number | string | boolean | null` *(JSON-serializable va
 
 JSON body. If the `Content-Type` header is not set, it will be set to `application/json`.
 
-###### userData
+###### context
 
 Type: `unknown`
 
-User data. In contrast to other options, `userData` is not enumerable.
+User data. In contrast to other options, `context` is not enumerable. Example:
+
+```js
+const got = require('got');
+
+const instance = got.extend({
+	hooks: {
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -178,6 +178,12 @@ Type: `Object` `Array` `number` `string` `boolean` `null`
 
 JSON body. The `Content-Type` header will be set to `application/json` if it's not defined.
 
+###### userData
+
+Type: `any`
```
tobenna: For type safety, would this not be better as `unknown`?

szmarczak: Indeed. Also I forgot to add it to the types.

readme.md
```diff
@@ -178,6 +178,12 @@ Type: `Object` `Array` `number` `string` `boolean` `null`
 
 JSON body. The `Content-Type` header will be set to `application/json` if it's not defined.
 
+###### userData
+
+Type: `unknown`
+
+User data. In the opposite to other options, `userData` is not enumerable.
```
sindresorhus: As a user reading this description, it's not clear what it's for or how it's useful to me as a user. I think we need to describe what they could use it for. Would also be great with a simple code example.

```suggestion
User data. In contrast to other options, `userData` is not enumerable.
```

test/arguments.ts
```diff
@@ -268,3 +268,19 @@ test('throws if the `searchParams` value is invalid', async t => {
...
+		hooks: {
+			beforeRequest: [
+				options => {
+					t.is(options.userData, 123);
+					t.false({}.propertyIsEnumerable.call(options, 'userData'));
+				}
+			]
+		}
+	});
+});
```
sindresorhus: I think we need some more tests. For example, ensure that `userData` is correctly preserved when extending the Got instance, and that it's available in hooks.

source/utils/types.ts
```diff
@@ -122,6 +122,7 @@ export interface AgentByProtocol {
 export interface Options extends Omit<https.RequestOptions, 'agent'> {
 	host: string;
 	body: string | Buffer | ReadableStream;
+	userData?: unknown;
```
sindresorhus: Nitpick, but I would place it at the bottom, above: `[key: string]: unknown;`.

sindresorhus:
We also should discuss making it `unknown` vs `object`. The benefit of `unknown` is that it can be any type. However, the benefit of having it always be an `object` is that we potentially could merge the object for the user.

sindresorhus:
// @juhovh

source/utils/types.ts
szmarczak: Sure.

sindresorhus:
> We also should discuss making it unknown vs object. The benefit of unknown is that it can be any type. However, the benefit of having it always be an object is that we potentially could merge the object for the user.

@szmarczak @juhovh @sholladay Any thoughts on this? I guess most will want an object here as they would want to name the data, for example `.userData.unicorn` instead of just `.userData`, which is not that readable. The merging factor is also tempting, but how far do we go, shallow merge, deep merge, or no merge at all? Any potential downsides of shallow merging?

sindresorhus:
@szmarczak => https://github.com/sindresorhus/got/pull/777#discussion_r278225585

szmarczak:
> @szmarczak => [#777 (comment)](https://github.com/sindresorhus/got/pull/777#discussion_r278225585)

Ah, forgot about that one.

> I guess most will want an object here as they would want to name the data, for example .userData.unicorn instead of just .userData, which is not that readable.

:+1:

> but how far do we go, shallow merge, deep merge, or no merge at all? Any potential downsides of shallow merging?

Shallow merge looks good to me. If someone wants deep merge we can discuss this later.

sholladay:
I would rather name this `app` instead of `userData`. Not only is it shorter, it feels more familiar to me, because [hapi](https://hapijs.com/) has this same exact feature named [`server.app`](https://github.com/hapijs/hapi/blob/master/API.md#server.app).

> Provides a safe place to store server-specific run-time application data without potential conflicts with the framework internals.

As for its type, I would probably go for `unknown` and avoid merging. If for no other reason than just the simplicity. Also, since merging usually requires some opinionated tradeoffs, it is worth keeping in mind that we don't "own" this namespace, so we should avoid being overly opinionated with it. In the end, though, I see this as something to be iterated on after more real-world experience.

sindresorhus:
> I would rather name this app instead of userData. Not only is it shorter, it feels more familiar to me, because hapi has this same exact feature named server.app .

There's a big difference though. `hapi` is a server framework, not a `request` client. Got is used in many situations where there's no concept of "app". I'm pretty content with the name `userData`. It makes it clear it's *for the user* and that it's for storing *data*.

> As for its type, I would probably go for unknown and avoid merging. If for no other reason than just the simplicity.

How about `object` and no merging? If we go with `unknown` now and later realize `object` is better, we'll need to do a breaking change, which will inconvenience users. I also think the readability point is an important factor: `console.log(response.userData)` vs `console.log(response.userData.unicorn)`.

juhovh:
Sorry for not commenting earlier, but I wasn’t quite sure then what would be the best option here. After giving it some thought, I believe this userData is shared between all different hooks, right? Unknown would make sense if it would be per hook, but shared data shouldn’t be of unknown type IMHO. Having an object with unknown values is fine, one argument for doing a shallow merge would be that one could throw an error in case of a conflict between hooks. For the same reason deep merge sounds wrong, since got doesn’t own the data that the hook sets, it all boils down to which part of the code owns what in the end.

sholladay:
FWIW, when I first saw this feature named `userData`, I honestly assumed it had something to do with the `user` portion of a URL (e.g. `username` and `password`). My second thought was that maybe it has something to do with extending got and customizing its default options. Perhaps that's just me, though? At any rate, now that I understand it, it very much feels like application state, not something that's necessarily static data. And above, the term _app_ is used in a pretty loose sense.

> How about object and no merging?

Yeah, I was thinking of suggesting that. Sounds good to me. 👍

szmarczak:
Regarding shallow merge:

```
const a = got.extend({
    userData: {
        unicorn: {
            value: "this won't get merged"
        },
        catSays: "meow"
    }
});

const b = a.extend({
    userData: {
        unicorn: {
            color: "rainbow"
        },
        dogSays: "woof"
    }
});

// Result:
{
    unicorn: {
        color: "rainbow",
    },
    catSays: "meow",
    dogSays: "woof"
}
```

There's one con. Should arrays be merged? Or replaced?

So yeah, it can be a object with no merge mechanics. We could simply emit old user data and new user data so the user can decide...

szmarczak:
Maybe let's leave it `unknown`? If people want to they can use it as an object too.

sindresorhus:
> FWIW, when I first saw this feature named userData, I honestly assumed it had something to do with the user portion of a URL (e.g. username and password). My second thought was that maybe it has something to do with extending got and customizing its default options. Perhaps that's just me, though?

Any other naming suggestions? I don't really want the `app` word in it, but we can maybe think of something else.

- `contextData`
- `customData`
- `contextObject`
- `context`
- `environment`
- `environmentObject` (This is what SwiftUI uses for shared data: https://developer.apple.com/documentation/swiftui/environmentobject)

> Yeah, I was thinking of suggesting that. Sounds good to me. 👍

Cool. Let's make it an object and no merging. TS typed as `{[key: string: unknown}`. 

Actually, should we make it generic? `context<T extends {[key: string: unknown}>`. That way users could make it more strictly typed if they wanted. Or am I overthinking it?

Another question, should it be a `Map`? That's usually a better data structure for key/value.

szmarczak:
IMO `customData` is too general. The type of the data should be specified. `context` / `contextData` / `contextObject` sounds good to me. When I hear `environment` I immediately think of `environment vars`. 

> Actually, should we make it generic?

:+1:

> should it be a Map?

It depends on the usage. Or maybe let's accept both: generic objects and maps?

sindresorhus:
> It depends on the usage. Or maybe let's accept both: generic objects and maps?

I think we should pick one. If we make it a union, users will have to type assert it, which is annoying.

sindresorhus:
I'm leaning towards `contextData`, but I will let the other people chime in first.

sholladay:
`context` and `contextData` are my favorite, in that order. My preference is only mild.

szmarczak:
Both `context` and `contextData` sound good to me. Is it possible to do pools in GitHub?

sindresorhus:
@sholladay Thoughts on Map vs object? I’m leaning towards Map.

sindresorhus:
Let’s go with `context`.

sindresorhus:
Hmm, Map would make it much harder to manually extend if the user wanted to do so. Let's go with object.

source/utils/types.ts
```diff
@@ -138,13 +138,11 @@ export interface Options extends Omit<https.RequestOptions, 'agent' | 'timeout'
...
 	searchParams?: Record<string, string | number | boolean | null> | URLSearchParams | string;
-	/*
-	Deprecated
-	 */
-	query?: Options['searchParams'];
+	query?: Options['searchParams']; // Deprecated
 	useElectronNet?: boolean;
 	form?: Record<string, any>;
 	json?: Record<string, any>;
+	context?: object;
```
szmarczak: How do I make this generic?

readme.md
```diff
@@ -174,6 +174,44 @@ Type: `object | Array | number | string | boolean | null` *(JSON-serializable va
 
 JSON body. If the `Content-Type` header is not set, it will be set to `application/json`.
 
+###### context
+
+Type: `unknown`
+
+User data. In contrast to other options, `context` is not enumerable.
```
sindresorhus: I think it needs some more non-code explanation than just "User data.". What is its purpose? Use-cases.

We should also mention that it's just passed through. Never merged. And Got never touches it.

PR #784, opened 2019-04-23
node 12 issues

panva:
This PR is to show that v9.x is failing under node 12


szmarczak:
We know about this. The only test that fails is `timer_wrap` check. That module has been deprecated in Node 11. Otherwise all tests pass.

panva:
Glad it’ll be handled then. Thank you

PR #787, opened 2019-04-24
Fix the retry functionality and increase coverage

### LINKED ISSUE #786: Retrying POST requests on client timeout
#### Describe the bug

 - Node.js version: v8.10 
 - Got version: v9.6.0

When setting a timeout on a POST request, and the timeout is reached, a retry is made even though the server did not close the connection, the client did. I don't expect POST requests to be retried, even if a TIMEOUT is reached because of the client timeout. 

#### Actual behavior

POST requests are retried if the client timeout is reached. 

#### Expected behavior

Do not retry POST requests, except (maybe) when the server closed the connection. 

#### Code to reproduce

invoke.js
```js
const got = require('got');

got.post('http://localhost:4001', {
  timeout: 6000,
})
  .then(resp => {
    console.log('GOT response', resp.body)
  })
  .catch(err => console.error('Some error', err))
```

slow-server.js
```js
var http = require('http');
var fs = require('fs');

let counter = 1

http.createServer(function (req, res) {
  const n = counter++
  console.log(`${n} Incoming req`, req.method, req.url)

  setTimeout(() => {
    console.log(`${n} Finishing request`, req.method, req.url)
    res.writeHead(200, { 'Content-Type': 'text/plain' });
    res.end('Hello world');
  }, 10000)
}).listen(4001);
```

Run slow-server.js and after that run invoke.js. You will see the server is receiving multiple requests even though the server is not the one closing the connection. 

#### Checklist

If we agree that this is an issue, I'm happy to work on a PR to fix it.

szmarczak:
Fixes #786


source/utils/unhandle.ts
```diff
@@ -0,0 +1,34 @@
...
+	event: Event;
+	fn: Fn;
+}
+
+interface Unhandler {
+	once: (origin: Origin, event: Event, fn: Fn) => void;
+	unhandleAll: () => void;
+}
+
+export default (): Unhandler => {
```
sindresorhus: Can you add a doc code comment about the purpose of this and what problem it solves?

source/utils/unhandle.ts
```diff
@@ -0,0 +1,34 @@
+import EventEmitter from 'events';
+
+type Origin = EventEmitter;
+type Event = string | symbol;
+type Fn = (...args: any[]) => void;
```
sindresorhus: `Fn` need a more descriptive name. Same with the other places the name `fn` is used. And use `unknown`.

source/utils/unhandle.ts
```diff
@@ -0,0 +1,34 @@
...
+	const handlers: Handler[] = [];
+
+	return {
+		once(origin: Origin, event: Event, fn: Fn) {
+			origin.once(event, fn);
+			handlers.push({origin, event, fn});
+		},
+
+		unhandleAll() {
+			handlers.forEach((handler: Handler) => {
```
sindresorhus: Use `for-of` loop.

source/utils/timed-out.ts
```diff
@@ -20,22 +21,15 @@ const reentry: symbol = Symbol('reentry');
...
-	let stopNewTimeouts = false;
-
-	const addTimeout = (delay: number, callback: (...args: any) => void, ...args: any): (() => void) => {
-		// An error had been thrown before. Going further would result in uncaught errors.
-		// See https://github.com/sindresorhus/got/issues/631#issuecomment-435675051
-		if (stopNewTimeouts) {
-			return noop;
-		}
-
+	const addTimeout = (delay: number, callback: (...args: any) => void, ...args: any): (typeof noop) => {
```
sindresorhus: Use `unknown` whenever possible.

source/utils/unhandle.ts
szmarczak: TypeScript complains about that when using `origin.removeListener(event, fn)`...

PR #788, opened 2019-04-24
Add Publish size to the Comparison

szmarczak:


szmarczak:
We need a better way to manage tables.

PR #791, opened 2019-05-01
Fix logo path

guerrerocarlos:
Make logo visible from https://www.npmjs.com/package/got otherwise it looks like this:

Screen Shot 2019-05-01 at 5 52 04 PM

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 <div align="center">
 	<br>
 	<br>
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, but this should be fixed by npm. I'm not interested in working around their bad Markdown handling. You can complain here: https://github.com/npm/marky-markdown/issues/434

PR #802, opened 2019-05-16
Ignore invalid cookies if sent by web server

iczero:
Some web servers (*cough* incapsula *cough*) will send invalid cookies. These should not fail the request, and instead should be ignored.

#### Checklist

- [ ] I have included some tests. - probably not needed?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -174,6 +174,45 @@ Type: `object | Array | number | string | boolean | null` *(JSON-serializable va
 
 JSON body. If the `Content-Type` header is not set, it will be set to `application/json`.
 
+###### context
+
+Type: `object`
+
+User data. In contrast to other options, `context` is not enumerable.
+
+**Note:** The object is never merged, it's just passed through. Got will not modify the object in any way.
+
+It's very useful for storing auth tokens:
+
+```js
+const got = require('got');
+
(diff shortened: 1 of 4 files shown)


iczero:
There exists a test for checking whether or not got throws on invalid cookies. I believe invalid cookies should be silently ignored, or at least there should exist a configuration for that.

sindresorhus:
I don't think we should silently ignore invalid cookies.

@szmarczak Do you think it's worth having an option for this?

szmarczak:
I don't think so. If we were willing to support deprecated functionalities, our code would be bloated.

iczero:
Currently, if the server sends invalid cookies, the entire request fails. As such, it would be impossible to use this module to perform requests to any server that sends invalid cookies. It's not so much about "supporting deprecated functionalities" as it is simply preserving compatibility. Currently, the invalid cookie in question fails the validation done by tough-cookie, however Chrome stores it fine.

sindresorhus:
Browsers are much more lenient than a request library should be. Got strives to be strict and not silently handle invalid data. However, I realize the real world is not perfect, so I'm ok with adding an option for this. Should be named `ignoreInvalidCookies` and should be opt-in, documented, and tested.

PR #812, opened 2019-05-28
Pass the response as the second argument to the `beforeRedirect` hook

### LINKED ISSUE #810: Pass response to beforeRedirect hook

Currently, the `beforeRedirect` hook only receives the next request options as argument. However, having the response is very useful for things like checking the url from which the redirect is returned and the number of times the request has been redirected.

#### Describe the feature

The `beforeRedirect` hook will receive the original response (the one that has a 3xx status code) as the second argument.

```javascript
const got = require('got');

got('https://example.com', {
	hooks: {
		beforeRedirect: [
			(options, response) => {
				console.log(`Redirected from ${response.url} to ${options.href}. Redirected ${response.redirectUrls.length} time(s).`)
			}
		]
	}
});
```

This should be a simple backwards-compatible change. I can submit a PR later today if all goes well.

##### Workaround(s)

Should be possible to get the response by switching to the streaming API. However, this requires code changes on the call site and (as far as I can see) not `extend`-able without refactoring call sites.

#### Checklist

clemyan:
Closes #810

> The `beforeRedirect` hook will receive the original response (the one that has a 3xx status code) as the second argument.

#### Checklist

### DIFF
--- a/source/known-hook-events.ts
+++ b/source/known-hook-events.ts
@@ -20,7 +20,7 @@ export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<v
 /**
 Called with normalized [request options](https://github.com/sindresorhus/got#options). Got will make no further changes to the request. This is especially useful when you want to avoid dead sites.
 */
-export type BeforeRedirectHook = (options: NormalizedOptions) => void | Promise<void>;
+export type BeforeRedirectHook = (options: NormalizedOptions, response: any) => void | Promise<void>;
 
 /**
 Called with normalized [request options](https://github.com/sindresorhus/got#options), the error and the retry count. Got will make no further changes to the request. This is especially useful when some extra work is required before the next try.
--- a/source/known-hook-events.ts
+++ b/source/known-hook-events.ts
@@ -1,3 +1,6 @@
+
+import http = require('http');
+import ResponseLike = require('responselike');
 import {Options, CancelableRequest, Response, NormalizedOptions} from './utils/types';
 import {HTTPError, GotError, ParseError, MaxRedirectsError} from './errors';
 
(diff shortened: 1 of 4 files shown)


source/known-hook-events.ts
```diff
@@ -21,7 +21,7 @@ export type BeforeRequestHook = (options: Options) => void | Promise<void>;
 /**
 Called with normalized [request options](https://github.com/sindresorhus/got#options). Got will make no further changes to the request. This is especially useful when you want to avoid dead sites.
 */
-export type BeforeRedirectHook = (options: Options) => void | Promise<void>;
+export type BeforeRedirectHook = (options: Options, response: any) => void | Promise<void>;
```
szmarczak: `any` -> `IncomingMessage`

sindresorhus:
Hey, sorry, we just merged https://github.com/sindresorhus/got/pull/797 which caused a lot of merge conflicts. Would you be able to fix the conflicts?

clemyan:
> Hey, sorry, we just merged #797 which caused a lot of merge conflicts. Would you be able to fix the conflicts?

I will see what I can do after I get off work.

Should I rebase onto master and squash?

sindresorhus:
You don’t need to squash, just rebase or pull from master.

source/known-hook-events.ts
```diff
@@ -20,7 +22,7 @@ export type BeforeRequestHook = (options: NormalizedOptions) => void | Promise<v
 /**
 Called with normalized [request options](https://github.com/sindresorhus/got#options). Got will make no further changes to the request. This is especially useful when you want to avoid dead sites.
 */
-export type BeforeRedirectHook = (options: NormalizedOptions) => void | Promise<void>;
+export type BeforeRedirectHook = (options: NormalizedOptions, response: http.ServerResponse | ResponseLike) => void | Promise<void>;
```
sindresorhus: Seeing as the `resonselike` package is meant to imitate a real response, we can just use the real response type.
```suggestion
export type BeforeRedirectHook = (options: NormalizedOptions, response: http.ServerResponse) => void | Promise<void>;
```

szmarczak: It's not `ServerResponse`. It's `IncomingMessage`.

szmarczak: Redirects aren't being cached, are they? No need to accept `ResponseLike` here.

clemyan: Ah! Didn't realize #797 added a type assertion.

https://github.com/sindresorhus/got/blob/4f6d287bf3a1fe62b5fed7b6a29b7ad8b797d792/source/request-as-event-emitter.ts#L122

So the hook should accept a `Response` and be passed `typedResponse` instead of accepting a `ServerResponse | ResponseLike` and being passed `response`.

PR #827, opened 2019-07-01
Add remote IP address to the response

### LINKED ISSUE #356: How to get IP address of the final destination host
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/356)

Hi,

How can I get the ip address of the final host after redirection?

Thanks,

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#827 Add remote ip address to response](https://issuehunt.io/r/sindresorhus/got/pull/827)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/got/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

Cerberooo:
Fixes #356

#### Checklist

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#356: How to get IP address of the final destination host](https://issuehunt.io/repos/18193978/issues/356)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>


source/utils/types.ts
```diff
@@ -67,6 +69,10 @@ export interface Response extends http.IncomingMessage {
 	request: { options: NormalizedOptions };
 }
 
+export interface ResponseObject extends ResponseLike {
```
Cerberooo: Is the naming okay?

Cerberooo:
For caching responses this feature needs an extension in [cacheable-request](https://github.com/lukechilds/cacheable-request) is needed.

sindresorhus:
I think it should work even when caching is enabled.

source/utils/types.ts
sindresorhus: Yes, but I think it would be better to fix `ResponseLike` to have all the expected response properties.

Cerberooo: Sure, but how to deal with RFC 7234?

sindresorhus: What about it?

Cerberooo: We can't cache the `remoteAddress` with `ResponseLike` see https://github.com/lukechilds/cacheable-request/issues/86

sindresorhus: I still don't see how that's related to the types though. https://github.com/lukechilds/cacheable-request/issues/86 is about caching.

sindresorhus:
I don't think we want to wait for https://github.com/lukechilds/cacheable-request/issues/86. We can just document that the remote IP address is not available in a cached response and link to the `cacheable-request` issue.

source/utils/types.ts
Cerberooo: But I think that lukechilds wont extend `ResponseLike` because of the RFC.

sindresorhus: That has nothing to do with the RFC. `ResponseLike` is meant to be exactly like a real response: https://github.com/lukechilds/responselike (It's a different module too).

sindresorhus:
@Cerberooo Bump :)

Cerberooo:
@sindresorhus sorry, I'll fix the tests and make a PR for `Responselike`
this or next week.

sindresorhus:
@Cerberooo This will need a test. And can you fix the merge conflict?

Cerberooo:
Sure. I'll add one.

test/ip.ts
```diff
@@ -0,0 +1,12 @@
+import test from 'ava';
+import withServer from './helpers/with-server';
+
+test('ip is defined', withServer, async (t, server, got) => {
```
szmarczak: This should be IMO located in https://github.com/sindresorhus/got/blob/master/test/promise.ts

Cerberooo: The naming is a little irritating for me. It does not check that a promise is returned, but checks individual values of the result.

szmarczak: > const {ip} = await got('');

You **do** check promises. You're using `await` to fetch the response.

Cerberooo: For me checking promises means to check if the promise itself is fullfilled or rejected. I'm checking if a attribute of the resolved value exists. But I got your point.

sindresorhus:
This fails now in Node.js 13:

```
  query.ts › overrides query from options

  /Users/sindresorhus/dev/oss/got/source/request-as-event-emitter.ts:3

   2: import {promisify} from 'util';
   3: import stream = require('stream');
   4: import EventEmitter = require('events');

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Cannot read property \'remoteAddress\' of undefined',
  }

  handleResponse (source/request-as-event-emitter.ts:3:35)
  ClientRequest.handler (node_modules/cacheable-request/src/index.js:148:7)
  ClientRequest.origin.emit (node_modules/@szmarczak/http-timer/source/index.ts:60:11)
```

Cerberooo:
From the [NodeJS 13 Changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md):

> The request.connection and response.connection properties are now runtime deprecated. The equivalent request.socket and response.socket should be used instead

Cerberooo:
```javascript
require('https').get('https://sindresorhus.com', r => console.log(r.socket.remoteAddress));
```

Works with NodeJS >=4.

PR #828, opened 2019-07-08
Make the test timeout helpers use `lolex`

### LINKED ISSUE #823: Random timeout tests are failing
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/823)

`setTimeout` is flawed - random timeout tests are failing.
We should use something like [`timekeeper`](https://github.com/vesln/timekeeper) or [`lolex`](https://github.com/sinonjs/lolex).

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#828 Update: test timeout helpers to use lolex (fix #823)](https://issuehunt.io/r/sindresorhus/got/pull/828)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/got/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

karimsa:
This PR is intended to fix #823. It includes a new test helper that produces & manages a new `lolex` clock and exposes it via a `ClientRequest`. Since the `ClientRequest` must be created using the test helper custom `request` method, the `request` override is passed via the options in `with-server`. This keeps the clock localized per `got` instance so tests can continue to run concurrently.

#### Checklist

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#823: Random timeout tests are failing](https://issuehunt.io/repos/18193978/issues/823)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>


karimsa:
Tests have been updated & now pass.

source/utils/timed-out.ts
```diff
@@ -31,14 +32,24 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
...
 
-	const addTimeout = (delay: number, callback: (...args: unknown[]) => void, ...args: unknown[]): (typeof noop) => {
 		// Event loop order is timers, poll, immediates.
 		// The timed event may emit during the current tick poll phase, so
 		// defer calling the handler until the poll phase completes.
 		let immediate: NodeJS.Immediate;
-		const timeout: NodeJS.Timeout = setTimeout(() => {
-			immediate = setImmediate(callback, delay, ...args);
+		const timeout: NodeJS.Timeout = timers.setTimeout(() => {
+			immediate = timers.setImmediate(callback, delay, type, ...args);
```
szmarczak: I don't like this approach. You're adding test code into production. These two need to be separated.

szmarczak: Use `lolex.install()` in tests (the `before` startup event).

szmarczak: `clock.tick()` seems to be proper here.

sindresorhus: Agreed. Only the test files should change, not the source files.

test/cancel.ts
```diff
@@ -157,30 +157,38 @@ test('recover from cancellation using error instance', async t => {
...
 test('throws on incomplete (canceled) response - promise', withServer, async (t, server, got) => {
-	server.get('/', downloadHandler);
+	server.get('/', downloadHandler(got));
 
 	await t.throwsAsync(got({
-		timeout: {request: 500}
-	}), got.TimeoutError);
+		timeout: {request: 500},
+		retry: 0
+	}).then(() => null), got.TimeoutError);
```
sindresorhus: Use async/await, not `.then`.

source/utils/timed-out.ts
karimsa: We can add a compile step to remove the test code but the strategy makes sense.

To use `.install()`, you would need to have a single clock instance that is shared across the entire node process. I initially tested with this & it doesn't work because:

  1. It causes issues with ava trying to schedule itself.
  2. All tests must be run serially in order to share a clock instance otherwise you end up with races.

Generally speaking, it is a better practice to use localized clocks since the dependency injection surface area is controlled.

test/cancel.ts
karimsa: `async/await` does not apply here... But I have removed this change anyways.

source/utils/timed-out.ts
karimsa: I have updated the `request.timers` property to use a symbol instead of an identifier property so that the behaviour can only be triggered from within the package.

szmarczak: Looks better :) I have no idea why you added the `type` argument - it's useless. We can call `request.setTimeout` directly.

IMO if the tests were run serially it would be a better approach.
@sindresorhus What do you think?

source/utils/timed-out.ts
```diff
@@ -23,6 +24,8 @@ export class TimeoutError extends Error {
 	}
 }
 
+export const kTimers = Symbol('$timers');
```
szmarczak: No need to place `$`

source/utils/timed-out.ts
karimsa: I can undo that change. I was previously injecting the `addTimeout` function rather than the timers object.

Even with serial execution, other modules are affected by globally injecting a fake clock. The tests just hang indefinitely.

sindresorhus:

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+import https = require('https');
+import lolex = require('lolex');
+import {kTimers} from '../../source/utils/timed-out';
+
+export function init() {
+	const clock = lolex.createClock();
+	let requestWasCreated = false;
+
+	return {
+		request: ((options: RequestOptions, cb) => {
```
sindresorhus:
```suggestion
		request: ((options: RequestOptions, callback) => {
```

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+			requestWasCreated = true;
+
+			const httpMod = options.protocol === 'https:' ? https : http;
```
sindresorhus:
```suggestion
			const httpModule = options.protocol === 'https:' ? https : http;
```

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+			const req = httpMod.request(options, cb);
```
sindresorhus:
```suggestion
			const request = httpMod.request(options, cb);
```

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+
+			let timeoutTimer;
+			let listenerAttached = false;
```
sindresorhus:
```suggestion
			let isListenerAttached = false;
```

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+						clock.setTimeout(() => {
+							if (listenerAttached) {
+								timeout();
+							}
+						}, delay);
+					}
+
+					updateTimer();
+
+					socket.on('data', () => updateTimer());
```
sindresorhus:
```suggestion
					socket.on('data', () => {
						updateTimer();
					});
```

test/timeout.ts
```diff
@@ -21,24 +21,32 @@ const keepAliveAgent = new http.Agent({
...
-		await delay(requestDelay);
-		response.end('OK');
-	});
+const defaultHandler = got => {
+	return (request, response) => {
+		request.resume();
+		request.on('end', async () => {
+			try {
+				got.tickTimers(requestDelay + 1);
+				setTimeout(() => response.end('OK'), 2);
```
sindresorhus:
```suggestion
				setTimeout(() => {
					response.end('OK');
				}, 2);
```

test/timeout.ts
```diff
@@ -21,24 +21,32 @@ const keepAliveAgent = new http.Agent({
...
+			} catch (error) {
+				console.error(error.stack || error);
```
sindresorhus:
```suggestion
				console.error(error);
```

test/timeout.ts
```diff
@@ -226,6 +241,7 @@ test('connect timeout (ip address)', withServer, async (t, server, got) => {
 				const socket = new net.Socket(options);
 				// @ts-ignore
 				socket.connecting = true;
+				setImmediate(() => got.tickTimers(requestDelay + 1));
```
sindresorhus:
```suggestion
				setImmediate(() => {
					got.tickTimers(requestDelay + 1);
				});
```

test/timeout.ts
```diff
@@ -270,11 +292,11 @@ test('secureConnect timeout not breached', withServer, async (t, server, got) =>
 });
 
 test('lookup timeout', withServer, async (t, server, got) => {
-	server.get('/', defaultHandler);
+	server.get('/', defaultHandler(got));
 
 	await t.throwsAsync(
 		got({
-			lookup: () => {},
+			lookup: () => setImmediate(() => got.tickTimers(requestDelay)),
```
sindresorhus:
```suggestion
			lookup: () => setImmediate(() => {
				got.tickTimers(requestDelay);
			}),
```

sindresorhus:
Bump

szmarczak:
I'm starting to think that your approach is much easier.

test/helpers/timed-out.ts
```diff
@@ -0,0 +1,71 @@
...
+import http = require('http');
+import https = require('https');
+import lolex = require('lolex');
+import {kTimers} from '../../source/utils/timed-out';
+
+export function init() {
+	const clock = lolex.createClock();
+	let requestWasCreated = false;
+
+	return {
```
szmarczak: Could you please help me understand this code? I just don't know what's this responsible for.

test/helpers/slow-data-stream.ts
```diff
@@ -1,18 +1,16 @@
...
+		for (let i = 0; i < 11; i++) {
 			slowStream.push('data\n'.repeat(100));
-			return;
+			await got.tickTimers(100); // eslint-disable-line no-await-in-loop
 		}
 
-		clearInterval(interval);
 		slowStream.push(null);
-	}, 100);
+	});
```
szmarczak: This is incorrect. It should be:

```js
import {Readable} from 'stream';
import {Clock} from 'lolex';

export default (clock: Clock): Readable => {
	let i = 0;

	return new Readable({
		read() {
			if (i++ < 10) {
				this.push('data\n'.repeat(100));
				clock.tick(100);
			} else {
				this.push(null);
			}
		}
	});
};
```

test/timeout.ts
szmarczak: This is also incorrect. It should be:

```
const defaultHandler = clock => (request, response) => {
	request.resume();
	request.on('end', async () => {
		clock.tick(requestDelay);

		response.end('OK');
	});
};
```

test/helpers/with-server.ts
```diff
@@ -12,10 +13,25 @@ export default async (t, run) => {
 		}
 	});
 
+	const {request, tickTimers} = timedOut.init();
+	const preparedGot = got.extend({
+		baseUrl: server.url,
+		request,
```
szmarczak: Overriding the `request` function is completely unnecessary. See #872

test/helpers/timed-out.ts
szmarczak: I think I get it. You don't override the global `setTimeout`, so you have to override `request.setTimeout`. But still, this looks way too complicated.

karimsa:
Sorry, haven't been able to work on this in a while. Looks like you guys preferred the serial route with #872, so closing in favour of that.

szmarczak:
I don't mind your solution. After #872 gets merged you can try to make it run in non-serial mode :)

PR #829, opened 2019-07-11
Replace `baseUrl` with `prefixUrl`

### LINKED ISSUE #783: Ambiguous URLs
#### Describe the bug

 - Node.js version: 8.15
 - OS & version: macOS
 - Got version: 9.6.0

#### Actual behavior

Setting up a got instance using `baseUrl` of `https://analyticsreporting.googleapis.com/v4`, and then using it with URLs with `:` in them (`/reports:batchGet`) cause an exception.

#### Expected behavior

I'd expect the following URL to be used: `https://analyticsreporting.googleapis.com/v4/reports:batchGet`

#### Code to reproduce

```js
const https = require('https')
const client = got.extend({
   baseUrl: 'https://analyticsreporting.googleapis.com/v4',
   agent: new https.Agent({ keepAlive: true }),
   encoding: null,
   decompress: false,
   throwHttpErrors: false
}
client.stream('/reports:batchGet')
```

Exception:

```
{ message: 'Unsupported protocol "reports:"',
  stack: 'UnsupportedProtocolError: Unsupported protocol "reports:"\n    at get (/<dir>/node_modules/got/source/request-as-event-emitter.js:53:10)\n    at Immediate.setImmediate [as _onImmediate] (/<dir>/node_modules/got/source/request-as-event-emitter.js:305:10)\n    at <anonymous>',
  name: 'UnsupportedProtocolError',
  host: undefined,
  hostname: '',
  method: 'GET',
  path: 'batchGet',
  socketPath: undefined,
  protocol: 'reports:',
  url: 'reports:batchGet',
  gotOptions: 
   { path: 'batchGet',
     protocol: 'reports:',
     hostname: '',
     hash: '',
     search: '',
     pathname: 'batchGet',
     href: 'reports:batchGet',
     retry: 
      { retries: [Function],
        methods: {},
        statusCodes: {},
        errorCodes: {} },
     hooks: 
      { beforeError: [],
        init: [],
        beforeRequest: [],
        beforeRedirect: [],
        beforeRetry: [],
        afterResponse: [] },
     decompress: false,
     throwHttpErrors: false,
     followRedirect: true,
     stream: true,
     form: false,
     json: false,
     cache: false,
     useElectronNet: false,
     baseUrl: 'https://analyticsreporting.googleapis.com/v4/',
     agent: 
      { domain: null,
        _events: { free: [Function] },
        _eventsCount: 1,
        _maxListeners: undefined,
        defaultPort: 443,
        protocol: 'https:',
        options: { keepAlive: true, path: null },
        requests: {},
        sockets: {},
        freeSockets: {},
        keepAliveMsecs: 1000,
        keepAlive: true,
        maxSockets: Infinity,
        maxFreeSockets: 256,
        maxCachedSessions: 100,
        _sessionCache: { map: {}, list: [] } },
     encoding: null,
     method: 'GET' } }
```

Note that doing the following in the Node REPL also simulates the same problem - so it's not a Got issue I guess but then I can't really work out what's going on! I'm only raising it here because from the [example in the README](https://github.com/sindresorhus/got#baseurl), using the above code looks like it should work!

```
> new url.URL('reports:batchGet', 'https://analyticsreporting.googleapis.com/v4/').toString()
'reports:batchGet'
```

From reading [other issues](https://github.com/sindresorhus/got/issues/562#issuecomment-412796684), this works however, but isn't ideal! It might be a case for further URL normalisation?

```
>new url.URL('./reports:batchGet', 'https://analyticsreporting.googleapis.com/v4/').toString()
'https://analyticsreporting.googleapis.com/v4/reports:batchGet'
```

#### Checklist

### LINKED ISSUE #814: got.extend doesn't support "unix:" protocol
#### Describe the bug

 - Node.js version: v10.13.0
 - OS & version: Unraid 6.8.0-rc0e

#### Actual behavior

```
UnhandledPromiseRejectionWarning: UnsupportedProtocolError: Unsupported protocol "unix:"
```

#### Expected behavior

```
UnhandledPromiseRejectionWarning: RequestError: connect ENOENT /var/run/docker.sock
```

#### Code to reproduce

Doesn't work:
```js
got.extend({ baseUrl: 'unix:/var/run/docker.sock' }).get('/images/json');
```

Works:
```js
got('unix:/var/run/docker.sock:/images/json');
```

#### Checklist

szmarczak:
#### Checklist

Fixes #783
Fixes #814

### DIFF
--- a/advanced-creation.md
+++ b/advanced-creation.md
@@ -261,10 +261,10 @@ const merged = got.mergeInstances(controlRedirects, limitDownloadUpload, httpbin
 	 */
 
 	const MEGABYTE = 1048576;
-	await merged('http://ipv4.download.thinkbroadband.com/5MB.zip', {downloadLimit: MEGABYTE});
+	await merged('http://ipv4.download.thinkbroadband.com/5MB.zip', {downloadLimit: MEGABYTE, prefixUrl: ''});
 	// CancelError: Exceeded the download limit of 1048576 bytes
 
-	await merged('https://jigsaw.w3.org/HTTP/300/301.html', {allowedHosts: ['google.com']});
+	await merged('https://jigsaw.w3.org/HTTP/300/301.html', {allowedHosts: ['google.com'], prefixUrl: ''});
 	// CancelError: Redirection to jigsaw.w3.org is not allowed
 })();
 ```
--- a/advanced-creation.md
+++ b/advanced-creation.md
@@ -220,7 +220,7 @@ const noUserAgent = got.extend({
 
 ```js
 const httpbin = got.extend({
-	baseUrl: 'https://httpbin.org/'
+	prefixUrl: 'https://httpbin.org/'
 });
 ```
 
--- a/migration-guides.md
+++ b/migration-guides.md
@@ -68,8 +68,6 @@ The [`timeout` option](https://github.com/sindresorhus/got#timeout) has some ext
 
 The [`searchParams` option](https://github.com/sindresorhus/got#searchParams) is always serialized using [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) unless it's a `string`.
 
-The [`baseUrl` option](https://github.com/sindresorhus/got#baseurl) appends the ending slash if it's not present.
-
 There's no `maxRedirects` option. It's always set to `10`.
 
 To use streams, just call `got.stream(url, options)` or `got(url, {stream: true, ...}`).
@@ -83,6 +81,7 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
 - No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
+- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present.
 
 ```js
 const gotInstance = got.extend({
--- a/migration-guides.md
+++ b/migration-guides.md
@@ -80,7 +80,7 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `agentClass`/`agentOptions`/`pool` option.
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
-- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present. Will be always prepended unless `url` is an instance of URL.
+- No `baseUrl` option. Instead, there is `prefixUrl` which appends a trailing slash if not present. It will be always prepended unless `url` is an instance of URL.
 - No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
 
 ```js
--- a/migration-guides.md
+++ b/migration-guides.md
@@ -80,8 +80,8 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `agentClass`/`agentOptions`/`pool` option.
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
+- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present. Will be always prepended unless `url` is an instance of URL.
 - No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
-- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present. If `prefixUrl` is used, it will be always prepended to the `url` argument.
 
 ```js
 const gotInstance = got.extend({
--- a/migration-guides.md
+++ b/migration-guides.md
@@ -81,7 +81,7 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
 - No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
-- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present.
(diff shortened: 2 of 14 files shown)


readme.md
```diff
@@ -114,28 +114,21 @@ Type: `object`
...
-If you specify an absolute URL, it will skip the `baseUrl`.
+Type: `string | URL`
 
-Very useful when used with `got.extend()` to create niche-specific Got instances.
+When specified, `prefixUrl` will be prepended to `url`. The prefix can be any valid URL, either relative or absolute. A trailing slash `/` is optional, one will be added automatically, if needed, when joining `prefixUrl` and `url`. The `url` argument cannot start with a `/` when using this option.
 
-Can be a string or a [WHATWG `URL`](https://nodejs.org/api/url.html#url_class_url).
-
-Slash at the end of `baseUrl` and at the beginning of the `url` argument is optional:
+Useful when used with `got.extend()1 to create niche-specific Got-instances.
```
```suggestion
Useful when used with `got.extend()` to create niche-specific Got-instances.
```

migration-guides.md
```diff
@@ -83,6 +81,7 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
 - No `removeRefererHeader` option. You can remove the referer header in a [`beforeRequest` hook](https://github.com/sindresorhus/got#hooksbeforeRequest):
+- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present.
```
szmarczak:
```suggestion
- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present. If `prefixUrl` is used, it will be always prepended to the `url` argument.
```

Actually should we do

```js
if (options.prefixUrl && !urlArgument.startsWith('http:') && !urlArgument.startsWith('https:') && !urlArgument.startsWith('unix:')) {
    ...
}
```

or let this behavior be?

advanced-creation.md
```diff
@@ -220,7 +220,7 @@ const noUserAgent = got.extend({
 
 ```js
 const httpbin = got.extend({
-	baseUrl: 'https://httpbin.org/'
+	prefixUrl: 'https://httpbin.org/'
```
szmarczak: This is broken for now. See https://github.com/sindresorhus/got/pull/829/files#r302707096

migration-guides.md
szmarczak: /cc @sholladay

sholladay: My opinion has always been that `input` should not be able to override the host of a `prefixUrl`, otherwise it's not really a prefix. If such an override _is_ needed, and it very well may be for some use cases, then I feel that full URL resolution is the answer (I.e. `baseUrl`). I don't think we have any business keeping a list of "special" schemes like `http:` where absolute URLs can override the target host. That is the job of a URL resolver, and I'd prefer to either fully embrace URL resolution or not at all.

I'm personally fine with having both `prefixUrl` and `baseUrl`. But I know Sindre felt that would be confusing.

szmarczak: What if we pass a URL instance to `input`? I think we should ignore `prefixUrl` then (same for Ky).

sholladay: Why does a URL instance change the behavior as opposed to a string? I would expect them to behave the same.

szmarczak: The URL instance is an absolute URL. Is there any use case for joining two absolute URLs? The string does not need to be absolute. That's the difference.

sholladay: Oh, right. I forgot that `URL` instances are always absolute, because that's such a silly limitation in the `URL` class. So, yeah, if absolute URLs are allowed to override `prefixUrl`, then it makes sense. However, in terms of the code, I would probably not explicitly check for `URL` instances, if possible. After all, what if `URL` starts supporting relative URLs in the future? I think they could introduce support for that without breaking too many things. And I hope they do, personally.

migration-guides.md
```diff
@@ -82,6 +80,7 @@ To use streams, just call `got.stream(url, options)` or `got(url, {stream: true,
 - No `agentClass`/`agentOptions`/`pool` option.
 - No `forever` option. You need to use [forever-agent](https://github.com/request/forever-agent).
 - No `proxy` option. You need to [pass a custom agent](readme.md#proxies).
+- No `baseUrl` option. Instead, there is `prefixUrl` which appends ending slash if not present. Will be always prepended unless `url` is an instance of URL.
```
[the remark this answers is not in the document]
sindresorhus: Why can't we prepend it when it's a URL instance?

test/cache.ts
```diff
@@ -134,7 +134,7 @@ test('doesn\'t cache response when received HTTP error', withServer, async (t, s
 
 test('DNS cache works', withServer, async (t, _server, got) => {
 	const map = new Map();
-	await t.notThrowsAsync(got('https://example.com', {dnsCache: map}));
+	await t.notThrowsAsync(got('https://example.com', {dnsCache: map, prefixUrl: null}));
```
sindresorhus: Why are you passing in `prefixUrl: null` here?

advanced-creation.md
```diff
@@ -261,10 +261,10 @@ const merged = got.mergeInstances(controlRedirects, limitDownloadUpload, httpbin
 	 */
 
 	const MEGABYTE = 1048576;
-	await merged('http://ipv4.download.thinkbroadband.com/5MB.zip', {downloadLimit: MEGABYTE});
+	await merged('http://ipv4.download.thinkbroadband.com/5MB.zip', {downloadLimit: MEGABYTE, prefixUrl: ''});
```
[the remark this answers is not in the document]
szmarczak: Because the instance has already defined the `prefixUrl` option. If defined, it's always prepended to the URL.

migration-guides.md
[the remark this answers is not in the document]
szmarczak: https://github.com/sindresorhus/got/pull/829#discussion_r303217292

test/cache.ts
szmarczak: https://github.com/sindresorhus/got/pull/829#discussion_r307972807

(it doesn't matter if the string's empty or it's null)

advanced-creation.md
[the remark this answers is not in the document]
sindresorhus: Ah, sorry. I missed that.

sindresorhus:
There are still some references to `baseUrl`:

```
readme.md
715:	baseUrl: 'https://example.com',
733:		baseUrl: 'httpbin.org',
1024:	baseUrl: 'https://<api-id>.execute-api.<api-region>.amazonaws.com/<stage>/',
1149:Use `got.extend()` to make it nicer to work with REST APIs. Especially if you use the `baseUrl` option.
1158:	baseUrl: 'example.com',
```

test/cache.ts
sindresorhus: Then I think you should pass an empty string here too for consistency and clarity.

migration-guides.md
sindresorhus: > My opinion has always been that input should not be able to override the host of a prefixUrl

👍 (We can reconsider if we get a lot of complain about this)

sindresorhus:
> There are still some references to baseUrl:

Fixed

sindresorhus:
Nice work! :)

PR #830, opened 2019-07-11
Don't break the default behavior when using custom retry function

### LINKED ISSUE #753: Custom `retry.retries` function unexpectedly breaks `statusCode` and other conditional retries
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/753)

Since `retry.statusCodes` and `retry.errorCodes` option support is implemented as part of the default `retries` function, those options are incompatible with custom retry function. Moreover even default `statusCodes` become not working so you get retries for HTTP 401, for example.
This is not quite obvious and documentation says nothing about that.

I think this is a bug to be fixed or at least it should be stated in readme.

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/got/issues/753).

szmarczak:
#### Checklist

Fixes #753

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -322,20 +322,21 @@ This also accepts an `object` with the following fields to constrain the duratio
 
 Type: `number | object`<br>
 Default:
-- retries: `2`
+- limit: `2`
+- retryFunction: `(iteration, retryOptions, error, computedValue) => computedValue`
 - methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
 - statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`500`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)
 - maxRetryAfter: `undefined`
 - errorCodes: `ETIMEDOUT` `ECONNRESET` `EADDRINUSE` `ECONNREFUSED` `EPIPE` `ENOTFOUND` `ENETUNREACH` `EAI_AGAIN`
 
-An object representing `retries`, `methods`, `statusCodes`, `maxRetryAfter` and `errorCodes` fields for the time until retry, allowed methods, allowed status codes, maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time and allowed error codes.
+An object representing `limit`, `retryFunction`, `methods`, `statusCodes`, `maxRetryAfter` and `errorCodes` fields for maximum iteration count, retry handler, allowed methods, allowed status codes, maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time and allowed error codes.
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`.<br>
 If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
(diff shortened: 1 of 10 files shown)


source/utils/types.ts
```diff
@@ -67,20 +67,22 @@ export interface Response extends http.IncomingMessage {
 	request: { options: NormalizedOptions };
 }
 
-export type RetryFunction = (retry: number, error: Error | GotError | ParseError | HTTPError | MaxRedirectsError) => number;
+export type RetryFunction = (retry: number, retryOptions: NormalizedRetryOptions, error: Error | GotError | ParseError | HTTPError | MaxRedirectsError, computedValue: number) => number;
```
sindresorhus: I think we need a clearer name for the `retry` argument.

And would be good to add doc comments for at least `computedValue` as it's not clear to be what that is.

There's a lot of arguments now. Since it's a breaking change anyway, maybe we should just pass in one object as an argument?

readme.md
```diff
@@ -322,20 +322,21 @@ This also accepts an `object` with the following fields to constrain the duratio
 
 Type: `number | object`<br>
 Default:
-- retries: `2`
+- limit: `2`
+- retryFunction: `(iteration, retryOptions, error, computedValue) => computedValue`
```
sindresorhus: I think we should use a clearer name here. Maybe `delay()`, `calculateDelay()`, or `getDelay()`. Other suggestions?

readme.md
```diff
@@ -322,20 +322,21 @@ This also accepts an `object` with the following fields to constrain the duratio
...
+An object representing `limit`, `retryFunction`, `methods`, `statusCodes`, `maxRetryAfter` and `errorCodes` fields for maximum iteration count, retry handler, allowed methods, allowed status codes, maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time and allowed error codes.
 
 If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`.<br>
 If [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header is greater than `maxRetryAfter`, it will cancel the request.
 
 Delays between retries counts with function `1000 * Math.pow(2, retry) + Math.random() * 100`, where `retry` is attempt number (starts from 1).
 
-The `retries` property can be a `number` or a `function` with `retry` and `error` arguments. The function must return a delay in milliseconds (`0` return value cancels retry).
+The `retryFunction` property is a `function` with `iteration`, `retryOptions`, `error` and `computedValue` arguments. The function must return a delay in milliseconds (`0` return value cancels retry).
```
sindresorhus: Would be good to explain what the different arguments are and what they are for. Same in the code as a doc comment.

source/utils/types.ts
szmarczak: > I think we need a clearer name for the retry argument.

`iterationCount`?

> There's a lot of arguments now. Since it's a breaking change anyway, maybe we should just pass in one object as an argument?

Sounds good.

readme.md
szmarczak: I'd go with `calculateDelay`.

source/utils/types.ts
sindresorhus: Is it the number of retries including the original try or not? In p-retry I use `attemptNumber`, but that including the original try. https://github.com/sindresorhus/p-retry

szmarczak: It is. What about `attemptCount`? Or just plural: `attempts`?

sindresorhus: > attemptCount

👍

readme.md

readme.md
```diff
@@ -322,20 +322,21 @@ This also accepts an `object` with the following fields to constrain the duratio
...
 - maxRetryAfter: `undefined`
 - errorCodes: `ETIMEDOUT` `ECONNRESET` `EADDRINUSE` `ECONNREFUSED` `EPIPE` `ENOTFOUND` `ENETUNREACH` `EAI_AGAIN`
 
-An object representing `retries`, `methods`, `statusCodes`, `maxRetryAfter` and `errorCodes` fields for the time until retry, allowed methods, allowed status codes, maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time and allowed error codes.
+An object representing `limit`, `retryFunction`, `methods`, `statusCodes`, `maxRetryAfter` and `errorCodes` fields for maximum iteration count, retry handler, allowed methods, allowed status codes, maximum [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) time and allowed error codes.
```
sindresorhus: `iteration` => `retry`

PR #833, opened 2019-07-15
Experimental pagination feature

### LINKED ISSUE #722: Pagination support
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/722)

With the recent enhancements such as create/extend and hooks, is there any clean way to implement pagination with the current approach?

Note: there exists [a discussion](https://github.com/sindresorhus/gh-got/issues/22) in `gh-got` already but I want to discuss it generically here.

#### Checklist

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/got/issues/722).

szmarczak:
Fixes #722

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1558,7 +1558,7 @@ Some of the Got features may not work properly. See [#899](https://github.com/si
 \* It's almost API compatible with the browser `fetch` API.\
 \*\* Need to switch the protocol manually. Doesn't accept PUSH streams and doesn't reuse HTTP/2 sessions.\
 \*\*\* Currently, only `DownloadProgress` event is supported, `UploadProgress` event is not supported.\
-:sparkle: Experimental feature. Don't hestitate to try it out!\
+:sparkle: Almost-stable feature, but the API may change. Don't hestitate to try it out!\
 :grey_question: Feature in early stage of development. Very experimental.
 
 <!-- GITHUB -->
--- a/readme.md
+++ b/readme.md
@@ -669,7 +669,9 @@ Checks whether the item should be emitted or not.
 Type: `Function`\
 Default: `(item, allItems) => true`
 
-Checks whether the pagination should continue. For example, if you need to stop **before** emitting an entry with some flag, you should use `(item, allItems) => !item.flag`. If you want to stop **after** emitting the entry, you should use `(item, allItems) => allItems.some(entry => entry.flag)` instead.
+Checks whether the pagination should continue.\
+For example, if you need to stop **before** emitting an entry with some flag, you should use `(item, allItems) => !item.flag`.\
+If you want to stop **after** emitting the entry, you should use `(item, allItems) => allItems.some(entry => entry.flag)` instead.
 
 ###### _pagination.countLimit
 
--- a/readme.md
+++ b/readme.md
@@ -637,34 +637,34 @@ got('https://api.github.com/some-endpoint', {
 });
 ```
 
-##### _pagination
+##### \_pagination
 
 Type: `Object`
 
 **Note:** This feature is experimental, although it's fairly stable to use.
 
-###### _pagination.transform
+###### \_pagination.transform
 
 Type: `Function`\
 Default: `response => JSON.parse(response.body)`
 
 A function that transform [Response](#response) into an array of items. This is where you should do the parsing.
 
-###### _pagination.paginate
+###### \_pagination.paginate
 
 Type: `Function`\
 Default: [`Link` header logic](source/index.ts)
 
 A function that returns an object representing Got options pointing to the next page. If there are no more pages, `false` should be returned.
 
-###### _pagination.filter
+###### \_pagination.filter
 
 Type: `Function`\
 Default: `(item, allItems) => true`
 
 Checks whether the item should be emitted or not.
 
-###### _pagination.shouldContinue
+###### \_pagination.shouldContinue
 
 Type: `Function`\
 Default: `(item, allItems) => true`
@@ -673,7 +673,7 @@ Checks whether the pagination should continue.\
 For example, if you need to stop **before** emitting an entry with some flag, you should use `(item, allItems) => !item.flag`.\
 If you want to stop **after** emitting the entry, you should use `(item, allItems) => allItems.some(entry => entry.flag)` instead.
 
-###### _pagination.countLimit
+###### \_pagination.countLimit
 
 Type: `number`\
 Default: `Infinity`
--- a/readme.md
+++ b/readme.md
@@ -26,7 +26,9 @@ For browser usage, we recommend [Ky](https://github.com/sindresorhus/ky) by the 
 
 ## Highlights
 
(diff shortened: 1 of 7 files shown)


source/index.ts
```diff
@@ -52,7 +53,34 @@ const defaults: Partial<Defaults> = {
 		dnsCache: false,
 		useElectronNet: false,
 		responseType: 'text',
-		resolveBodyOnly: false
+		resolveBodyOnly: false,
+		pagination: {
+			parse: response => {
+				return JSON.parse(response.body);
+			},
+			paginate: response => {
```
szmarczak: @sindresorhus I don't think we should implement a `paginate` function by default. Some people may want to go forwards, some will choose backwards, some will want to filter the elements etc.

Should `filter` be another property?

sindresorhus: > I don't think we should implement a paginate function by default. Some people may want to go forwards, some will choose backwards, some will want to filter the elements etc.

I still think it's nice to have a good default. Just getting all the pages or a set amount of items is the 90% use-case.

> Should filter be another property?

Yes

source/index.ts
```diff
@@ -52,7 +53,34 @@ const defaults: Partial<Defaults> = {
...
+				if (response.headers.link) {
```
sindresorhus: Use an early-return.

source/index.ts
```diff
@@ -52,7 +53,34 @@ const defaults: Partial<Defaults> = {
...
+					const items = (response.headers.link as string).split(',');
+
+					let next: string;
+					for (const item of items) {
+						const parsed = item.split(';');
+
+						if (parsed[1].includes('next')) {
+							next = parsed[0].trimStart().trim();
+							next = next.substring(1, next.length - 1);
```
sindresorhus: Use `.slice()`.

source/index.ts
```diff
@@ -52,7 +53,34 @@ const defaults: Partial<Defaults> = {
 		dnsCache: false,
 		useElectronNet: false,
 		responseType: 'text',
-		resolveBodyOnly: false
+		resolveBodyOnly: false,
+		pagination: {
+			parse: response => {
```
sindresorhus: Maybe this should be named `transform`? As it could be used for other things too. Per: https://github.com/sindresorhus/got/issues/722#issuecomment-493811004

sindresorhus:
We need an option for `itemCount` too.

source/index.ts
szmarczak: Agreed.

szmarczak:
> We need an option for itemCount too.

Instead of an `itemCount` option I think it's better to have this:

```js
const results = await got.paginate.all({
	pagination: {
		shouldContinue: allItems => allItems.length !== 3
	}
});
```

because people may want have custom conditions.

szmarczak:
@rarkins Could you look at the [tests](https://github.com/sindresorhus/got/pull/833/files#diff-5ea4dcf0b133bc707f2dc0b08924509e) and tell if the API is ok?

rarkins:
> @rarkins Could you look at the [tests](https://github.com/sindresorhus/got/pull/833/files#diff-5ea4dcf0b133bc707f2dc0b08924509e) and tell if the API looks ok?

I think it looks good so far, thanks

source/as-stream.ts
```diff
@@ -91,7 +92,7 @@ export default function asStream<T>(options: NormalizedOptions): ProxyStream<T>
 				// It's not possible to decompress already decompressed data, is it?
 				const isAllowed = options.decompress ? key !== 'content-encoding' : true;
 				if (isAllowed) {
-					destination.setHeader(key, value);
+					destination.setHeader(key, value as any);
```
szmarczak: this needs to be fixed later

szmarczak:
I need to make more tests here

rarkins:
@szmarczak one day I would like to see pagination with a `limit`, e.g. you might want the first 1000 Pull Requests from GitHub's API. Would you anticipate that being an *option* to `got.paginate.all` or would it be a new function like `got.paginate.some`? Also we'd ideally like to combine it with a filter too eventually, e.g. you can get the first X paginated items that match logic Y.

szmarczak:
A side note for me: There's a missing test for `options.pagination.shouldContinue`.

@rarkins You can do something like this:

```js
const results = await got.paginate.all({
    pagination: {
        ...,
        shouldContinue: items => items.length < 1000
    }
});
```

That will give you max 1000 items. But if it's a common use case, then I can make `options.pagination.countLimit`.

rarkins:
> But if it's a common use case

I'm pretty sure it would be commonly used

szmarczak:
> I'm pretty sure it would be commonly used

So you have a new `countLimit` option now :) I just need to update the docs and the PR is good to merge.

szmarczak:
I'm gonna write the docs tomorrow.

source/index.ts
```diff
@@ -33,7 +33,7 @@ const defaults: Defaults = {
 				'ENETUNREACH',
 				'EAI_AGAIN'
 			],
-			maxRetryAfter: undefined,
+			maxRetryAfter: Infinity,
```
szmarczak:
```suggestion
			maxRetryAfter: undefined,
```

sindresorhus:
How about we prefix the option with an underscore and document that the option is experimental and that users should lock the Got dependency as it might change in minor releases? This will give users a chance to try it out, but also gives us a chance to improve it without having to do lots of major releases?

szmarczak:
Sure, will finish this today.

readme.md
```diff
@@ -1488,7 +1557,8 @@ Some of the Got features may not work properly. See [#899](https://github.com/si
 \* It's almost API compatible with the browser `fetch` API.\
 \*\* Need to switch the protocol manually. Doesn't accept PUSH streams and doesn't reuse HTTP/2 sessions.\
 \*\*\* Currently, only `DownloadProgress` event is supported, `UploadProgress` event is not supported.\
-:grey_question: Experimental support.
+:sparkle: Experimental feature. Don't hestitate to try it out!\
```
```suggestion
:sparkle: Almost-stable feature, but the API may change. Don't hestitate to try it out!\
```

readme.md
```diff
@@ -635,6 +637,49 @@ got('https://api.github.com/some-endpoint', {
...
+
+Type: `Function`\
+Default: `response => JSON.parse(response.body)`
+
+A function that transform [`Response`](#response) into an array of items. This is where you should do the parsing.
+
+###### \_pagination.paginate
+
+Type: `Function`\
+Default: [`Link` header logic](source/index.ts)
```
sindresorhus: I think we should explain the Link header logic in English prose, not just link to the source.

sindresorhus:
https://github.com/sindresorhus/got/issues/1053

sindresorhus:
This is looking great! Nice work, @szmarczak 👌

PR #846, opened 2019-07-31
Cancel all timeouts when one occurs

timdp:
As I've explained a few times under previous issues that our team reported, our HTTP client code relies heavily on socket reuse. In this case, it's imperative that all listeners and timeouts attached to a request get cleaned up before got finishes handling the request. Sorry to keep coming back to that use case but issues with it can hurt us badly.

So... When firing many requests to the same IP, we see spurious `lookup` timeout errors. They seem to be happening because a timeout from a previous call leaked into the current one due to socket reuse. Our `lookup` timeout is 500 ms and we heavily optimize DNS so that the `lookup()` call actually takes less than 1 ms, so the 500 ms is only supposed to hit at 100% CPU, which is definitely not the case.

`timed-out` currently seems to do a better job at removing event listeners than at clearing timeouts. Timeouts will only get cleared upon an error event (unless it's a socket hang up?), but the occurrence of one timeout will not cancel any other ones.

This PR tries to address that by:

- calling `cancelTimeouts` from the timeout callback itself, so that any timeout will clear all the other ones;

- clearing the arrays of listeners and timeouts so that they only get removed once.

We're still setting up a high-volume test for this logic but I already wanted to test the water.

Along the way, I also noticed some related things:

- `cancelTimeouts` actually clears timeouts _and_ removes event listeners, so something like `dispose`, `finalize`, `cleanUp`, ... would probably be a better name.

- `unhandle` could be useful to other people as well so it could be factored out into a package. In fact, many years ago, I already wrote [event-registry](https://npmjs.com/package/event-registry) which addresses the same problem. I may have mentioned that before.

- Conversely, I'm not sure `unhandle` is necessary if you already maintain an array of `cancelers`. Currently, disposal/cleanup is spread across two modules and I think it'd be much nicer if there were a single place where all of the cleanup logic converged. As mentioned above, `cancelTimeouts` also does part of it but not all. Is there a good reason for keeping the two separate?

### DIFF
--- a/source/utils/timed-out.ts
+++ b/source/utils/timed-out.ts
@@ -32,12 +32,23 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 	const cancelers: Array<typeof noop> = [];
 	const {once, unhandleAll} = unhandler();
 
+	const cancelTimeouts = (): void => {
+		for (const cancel of cancelers) {
+			cancel();
+		}
+
+		cancelers.length = 0;
+
+		unhandleAll();
+	};
+
 	const addTimeout = (delay: number, callback: (...args: unknown[]) => void, ...args: unknown[]): (typeof noop) => {
 		// Event loop order is timers, poll, immediates.
 		// The timed event may emit during the current tick poll phase, so
 		// defer calling the handler until the poll phase completes.
 		let immediate: NodeJS.Immediate;
 		const timeout: NodeJS.Timeout = setTimeout(() => {
+			cancelTimeouts();
 			immediate = setImmediate(callback, delay, ...args);
 			/* istanbul ignore next: added in node v9.7.0 */
 			if (immediate.unref) {
@@ -65,14 +76,6 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 	const timeoutHandler = (delay: number, event: string): void => {
 		request.emit('error', new TimeoutError(delay, event));
 		request.abort();
-	};
-
-	const cancelTimeouts = (): void => {
-		for (const cancel of cancelers) {
-			cancel();
-		}
-
-		unhandleAll();
 	};
 
 	request.on('error', error => {
--- a/source/utils/unhandle.ts
+++ b/source/utils/unhandle.ts
@@ -33,6 +33,8 @@ export default (): Unhandler => {
 				const {origin, event, fn} = handler;
 				origin.removeListener(event, fn);
 			}
+
+			handlers.length = 0;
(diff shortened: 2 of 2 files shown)


szmarczak:
> Timeouts will only get cleared upon an error event (unless it's a socket hang up?)

`timed-out` ignores `socket hang up` errors as these are **not** thrown by Got. Why does Got doesn't throw these? IIRC to support retrying, because once it's thrown the whole request is gone.

szmarczak:
> unhandle could be useful to other people

Everything located in `utils/*` should be a different package.

szmarczak:
> Conversely, I'm not sure unhandle is necessary if you already maintain an array of cancelers. 

The cancelers array stores timeouts and immediates, which are about to happen.

Unhandle removes all attached listeners - they may happen in the future.

> Is there a good reason for keeping the two separate?

More readable code.

source/utils/timed-out.ts
```diff
@@ -32,12 +32,23 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
...
+		unhandleAll();
+	};
+
 	const addTimeout = (delay: number, callback: (...args: unknown[]) => void, ...args: unknown[]): (typeof noop) => {
 		// Event loop order is timers, poll, immediates.
 		// The timed event may emit during the current tick poll phase, so
 		// defer calling the handler until the poll phase completes.
 		let immediate: NodeJS.Immediate;
 		const timeout: NodeJS.Timeout = setTimeout(() => {
+			cancelTimeouts();
```
szmarczak: I don't think it has any sense. It's useless. Once the error is thrown all timeouts are canceled.

timdp: Unless the error isn't emitted, of course. Or if another listener for the error event did something weird.

timdp:
See #857 for followup. I still think something similar to this PR would be good to have, but our discovery of that issue alleviated the need for this.

source/utils/timed-out.ts
szmarczak: If it isn't emitted, only one of these two cases could happen:
1. The error was thrown and cancelled all the timeouts.
2. The request was aborted, so no further errors are thrown:

https://github.com/sindresorhus/got/blob/985a3b2d384072bd7ddc9f2f89474cd3a41592a7/source/request-as-event-emitter.ts#L196

I have double checked, it's not possible for the error event listener to behave weird.

szmarczak:
I'll close this one as it doesn't change anything. #857 is fixed.

PR #849, opened 2019-08-03
Make the TypeScript types strict and better

LinusU:
Blocked on: form-data/form-data#435 or DefinitelyTyped/DefinitelyTyped#37331

Next steps:

- [x] Upgrade to TypeScript 3.5
- [ ] Make `test/**/*.ts` conform to strict mode
- [ ] Turn on strict mode in `tsconfig.json`
- [ ] Everything else from #758

There is currently 1218 TypeScript errors in test files when running in strict mode 😂 

Will submit some typings upstream and then type everything up!

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 		"node": ">=8.6"
 	},
 	"scripts": {
-		"test": "xo && nyc ava",
+		"test": "xo && nyc ava && tsc --noEmit --strict",
 		"release": "np",
 		"build": "del-cli dist && tsc",
 		"prepublishOnly": "npm run build"
--- a/source/utils/timed-out.ts
+++ b/source/utils/timed-out.ts
@@ -3,6 +3,9 @@ import {ClientRequest, IncomingMessage} from 'http';
 import {Delays} from './types';
 import unhandler from './unhandle';
 
+// Help TypeScript to understand how we are passing arguments thru setImmediate
+declare function setImmediate<T extends any[]>(callback: (arg0: number, ...args: T) => void, arg0: number, ...args: T): NodeJS.Immediate;
+
 const reentry = Symbol('reentry');
 const noop = (): void => {};
 
@@ -32,7 +35,7 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 	const cancelers: Array<typeof noop> = [];
 	const {once, unhandleAll} = unhandler();
 
-	const addTimeout = (delay: number, callback: (...args: unknown[]) => void, ...args: unknown[]): (typeof noop) => {
+	const addTimeout = <T extends any[]>(delay: number, callback: (delay: number, ...args: T) => void, ...args: T): (typeof noop) => {
 		// Event loop order is timers, poll, immediates.
 		// The timed event may emit during the current tick poll phase, so
 		// defer calling the handler until the poll phase completes.
@@ -91,7 +94,7 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 
 	if (delays.socket !== undefined) {
 		const socketTimeoutHandler = (): void => {
-			timeoutHandler(delays.socket, 'socket');
+			timeoutHandler(delays.socket!, 'socket');
 		};
 
 		request.setTimeout(delays.socket, socketTimeoutHandler);
@@ -110,15 +113,15 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 
 		/* istanbul ignore next: hard to test */
 		if (socket.connecting) {
-			if (delays.lookup !== undefined && !socketPath && !net.isIP(hostname || host)) {
+			if (delays.lookup !== undefined && !socketPath && !net.isIP(hostname || host || '')) {
 				const cancelTimeout = addTimeout(delays.lookup, timeoutHandler, 'lookup');
 				once(socket, 'lookup', cancelTimeout);
 			}
 
 			if (delays.connect !== undefined) {
-				const timeConnect = (): (() => void) => addTimeout(delays.connect, timeoutHandler, 'connect');
+				const timeConnect = (): (() => void) => addTimeout(delays.connect!, timeoutHandler, 'connect');
 
-				if (socketPath || net.isIP(hostname || host)) {
+				if (socketPath || net.isIP(hostname || host || '')) {
 					once(socket, 'connect', timeConnect());
 				} else {
 					once(socket, 'lookup', (error: Error): void => {
@@ -131,14 +134,14 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 
 			if (delays.secureConnect !== undefined && options.protocol === 'https:') {
 				once(socket, 'connect', (): void => {
-					const cancelTimeout = addTimeout(delays.secureConnect, timeoutHandler, 'secureConnect');
+					const cancelTimeout = addTimeout(delays.secureConnect!, timeoutHandler, 'secureConnect');
 					once(socket, 'secureConnect', cancelTimeout);
 				});
 			}
 		}
 
 		if (delays.send !== undefined) {
-			const timeRequest = (): (() => void) => addTimeout(delays.send, timeoutHandler, 'send');
+			const timeRequest = (): (() => void) => addTimeout(delays.send!, timeoutHandler, 'send');
 			/* istanbul ignore next: hard to test */
 			if (socket.connecting) {
 				once(socket, 'connect', (): void => {
--- a/source/utils/timed-out.ts
+++ b/source/utils/timed-out.ts
@@ -3,7 +3,7 @@ import {ClientRequest, IncomingMessage} from 'http';
 import {Delays} from './types';
 import unhandler from './unhandle';
 
-// Help TypeScript to understand how we are passing arguments thru setImmediate
+// Help TypeScript understand how we are passing arguments through `setImmediate`.
 declare function setImmediate<T extends any[]>(callback: (arg0: number, ...args: T) => void, arg0: number, ...args: T): NodeJS.Immediate;
 
 const reentry = Symbol('reentry');
--- a/source/utils/types.ts
+++ b/source/utils/types.ts
@@ -69,7 +69,7 @@ export interface Response extends http.IncomingMessage {
 
 export type RetryFunction = (retry: number, error: Error | GotError | ParseError | HTTPError | MaxRedirectsError) => number;
 
-export type HandlerFunction = <T extends ProxyStream | CancelableRequest<Response>>(options: Options, next: (options: Options) => T) => T;
+export type HandlerFunction = <T extends ProxyStream | CancelableRequest<Response>>(options: NormalizedOptions, next: (options: NormalizedOptions) => T) => T;
 
 export interface RetryOption {
 	retries?: RetryFunction | number;
@@ -146,7 +146,7 @@ export interface Options extends Omit<https.RequestOptions, 'agent' | 'timeout' 
 	context?: {[key: string]: unknown};
 }
 
-export interface NormalizedOptions extends Omit<Required<Options>, 'timeout' | 'dnsCache' | 'retry'> {
+export interface NormalizedOptions extends Omit<Required<Options>, 'timeout' | 'dnsCache' | 'retry' | 'auth' | 'body' | 'port'> {
 	hooks: Hooks;
 	gotTimeout: Required<Delays>;
 	retry: NormalizedRetryOptions;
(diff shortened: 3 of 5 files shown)


source/utils/unhandle.ts
```diff
@@ -2,7 +2,7 @@ import EventEmitter = require('events');
 
 type Origin = EventEmitter;
 type Event = string | symbol;
-type Fn = (...args: unknown[]) => void;
+type Fn = (...args: any[]) => void;
```
LinusU: This matches how it's typed upstream (from `import EventEmitter = require('events');`)

source/utils/types.ts
```diff
@@ -155,6 +155,9 @@ export interface NormalizedOptions extends Omit<Required<Options>, 'timeout' | '
 	path: string;
 	hostname: string;
 	host: string;
+	auth?: Options['auth'];
+	body?: Options['body'];
+	port?: Options['port'];
```
LinusU: `Required` makes everything required, but these are in fact optional

source/utils/timed-out.ts
```diff
@@ -110,15 +113,15 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 
 		/* istanbul ignore next: hard to test */
 		if (socket.connecting) {
-			if (delays.lookup !== undefined && !socketPath && !net.isIP(hostname || host)) {
+			if (delays.lookup !== undefined && !socketPath && !net.isIP(hostname || host || '')) {
```
LinusU: `net.isIP(undefined)` returns `0` but the types for `net.isIP` doesn't allow passing `undefined`.

It doesn't seem impossible that `net.isIP(undefined)` will start throwing `TypeError` in a future version of Node.js, so I think it's better to pass in an empty string which should always return `0`.

LinusU:
@sindresorhus as soon as form-data/form-data#435 or DefinitelyTyped/DefinitelyTyped#37331 is merged we could merge just this commit if you want to. It adds strict type checkin to the library part, and enforces it with an extra `tsc` call in the `test` phase.

sindresorhus:
// @vladfrangu FYI

sindresorhus:
> Or we could wait for me to finish typing the tests ☺️

I would prefer merging this when it passes first. Smaller PRs are good and makes it less likely to conflict with other changes.

package.json
```diff
@@ -9,7 +9,7 @@
 		"node": ">=8.6"
 	},
 	"scripts": {
-		"test": "xo && nyc ava",
+		"test": "xo && nyc ava && tsc --noEmit --strict",
```
sindresorhus: I assumed `ts-node` already did this kind of validation? Aren't we running TS twice now?

https://github.com/sindresorhus/got/blob/985a3b2d384072bd7ddc9f2f89474cd3a41592a7/package.json#L103

source/utils/timed-out.ts
```diff
@@ -3,6 +3,9 @@ import {ClientRequest, IncomingMessage} from 'http';
 import {Delays} from './types';
 import unhandler from './unhandle';
 
+// Help TypeScript to understand how we are passing arguments thru setImmediate
+declare function setImmediate<T extends any[]>(callback: (arg0: number, ...args: T) => void, arg0: number, ...args: T): NodeJS.Immediate;
```
sindresorhus: Is this something you could upstream to the Node.js types?

source/utils/timed-out.ts
```diff
@@ -3,6 +3,9 @@ import {ClientRequest, IncomingMessage} from 'http';
 import {Delays} from './types';
 import unhandler from './unhandle';
 
+// Help TypeScript to understand how we are passing arguments thru setImmediate
```
sindresorhus:
```suggestion
// Help TypeScript understand how we are passing arguments through `setImmediate`.
```

source/utils/timed-out.ts
LinusU: Unfortunately it requires a very recent version of TypeScript, and they generally don't accept PRs with that since it breaks for a lot of users using DefinitivelyTyped 😞

package.json
LinusU: Yeah, but only one of them in strict mode. If we add `strict: true` to the `tsconfig.json` it will try to compile all the test code as strict mode as well. So this was the middle ground in the mean time, it does an extra compile of the library code in strict mode, just to check that it's conformant.

LinusU: It's not currently possible to have some parts non-strict: https://github.com/Microsoft/TypeScript/issues/28306

source/utils/timed-out.ts
sindresorhus: Alright, I can you add a TODO comment about upstreaming it at some point?

LinusU:
Running into microsoft/TypeScript#32693 when upgrading to TS 3.5 😬

source/utils/timed-out.ts
LinusU: We'll also be running into this when it's upstreamed 😬 

https://github.com/microsoft/TypeScript/issues/26113

Do you want to me to **a)**

Declare `setImmediate` to something that could be upstreamed:

```ts
// TODO: Upstream this once DefinitelyTyped accepts `T extends any[]`
declare function setImmediate<T extends any[]>(callback: (...args: T) => void, ...args: T): NodeJS.Immediate;
```

and then ignore the error where we use it:

```ts
// @ts-ignore https://github.com/microsoft/TypeScript/issues/26113
immediate = setImmediate(callback, delay, ...args);
```

or **b)**

Keep our specialised functions and add both comments there:

```ts
// TODO: Remove `arg0` when https://github.com/microsoft/TypeScript/issues/26113 is fixed
// TODO: Upstream this once DefinitelyTyped accepts `T extends any[]`
declare function setImmediate<T extends any[]>(callback: (arg0: number, ...args: T) => void, arg0: number, ...args: T): NodeJS.Immediate;
```

source/utils/timed-out.ts
```diff
@@ -110,15 +113,15 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
...
 				const cancelTimeout = addTimeout(delays.lookup, timeoutHandler, 'lookup');
 				once(socket, 'lookup', cancelTimeout);
 			}
 
 			if (delays.connect !== undefined) {
-				const timeConnect = (): (() => void) => addTimeout(delays.connect, timeoutHandler, 'connect');
+				const timeConnect = (): (() => void) => addTimeout(delays.connect!, timeoutHandler, 'connect');
 
-				if (socketPath || net.isIP(hostname || host)) {
+				if (socketPath || net.isIP(hostname || host || '')) {
```
vladfrangu: I see that this net.isIP check is used at least twice, can't we abstract it to one central variable?

source/utils/timed-out.ts
```diff
@@ -32,7 +35,7 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 	const cancelers: Array<typeof noop> = [];
 	const {once, unhandleAll} = unhandler();
 
-	const addTimeout = (delay: number, callback: (...args: unknown[]) => void, ...args: unknown[]): (typeof noop) => {
+	const addTimeout = <T extends any[]>(delay: number, callback: (delay: number, ...args: T) => void, ...args: T): (typeof noop) => {
```
vladfrangu: Does the callback actually take the delay as the first argument? it doesn't seem like it

source/utils/timed-out.ts
vladfrangu: I am not really a fan of manually overriding the function, while it it how *we* use it. Thoughts on just `// @ts-ignore` with a TODO to remove once the issue is closed?

sindresorhus: Let's go with just a tsignore comment.

package.json
sindresorhus: Makes sense 👍

sindresorhus:
TypeScript 3.6 is out. Maybe that fixes some bugs? https://devblogs.microsoft.com/typescript/announcing-typescript-3-6/

szmarczak:
@LinusU ping :wink:

sindresorhus:
I've pulled the current changes of this PR into master as I plan to do an early alpha release: https://github.com/sindresorhus/got/commit/5002e5445ca23a17439280739fd30d63cccd43d5

sindresorhus:
I have upgraded TS to latest now: https://github.com/sindresorhus/got/commit/640c30bbe37fa399a6f9c96357d897f45e54f52f

szmarczak:
@yovanoc Why? There's no need.

yovanoc:
Check yourself, with NodeJS setTimeout return a number and global.setTimeout return a NodeJS.Timeout

szmarczak:
I did and you're wrong:

```
$ node
Welcome to Node.js v12.10.0.
Type ".help" for more information.
> setTimeout(() => {}, 1000);
Timeout {
  _idleTimeout: 1000,
  _idlePrev: [TimersList],
  _idleNext: [TimersList],
  _idleStart: 20031,
  _onTimeout: [Function],
  _timerArgs: undefined,
  _repeat: null,
  _destroyed: false,
  [Symbol(refed)]: true,
  [Symbol(asyncId)]: 98,
  [Symbol(triggerId)]: 5
}
> global.setTimeout(() => {}, 1000);
Timeout {
  _idleTimeout: 1000,
  _idlePrev: [TimersList],
  _idleNext: [TimersList],
  _idleStart: 43542,
  _onTimeout: [Function],
  _timerArgs: undefined,
  _repeat: null,
  _destroyed: false,
  [Symbol(refed)]: true,
  [Symbol(asyncId)]: 272,
  [Symbol(triggerId)]: 5
}
```

pro-src:
@yovanoc How would that make sense given the type definitions?

<details><summary>Definitions</summary>

<br>

[Set & Clear methods](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6fa32aafe709137635b9b401cbd1eb9cdfc07b73/types/node/globals.d.ts#L178-L186) <-- Root scope

declare namespace NodeJS `{`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;interface Global `{`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Clear methods](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6fa32aafe709137635b9b401cbd1eb9cdfc07b73/types/node/globals.d.ts#L1110-L1112)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Set methods](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6fa32aafe709137635b9b401cbd1eb9cdfc07b73/types/node/globals.d.ts#L1127-L1129)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Timer, Immediate, Timeout](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6fa32aafe709137635b9b401cbd1eb9cdfc07b73/types/node/globals.d.ts#L1138-L1157)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`}`

`}`
</details>

source/utils/timed-out.ts
```diff
@@ -91,7 +94,7 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 
 	if (delays.socket !== undefined) {
 		const socketTimeoutHandler = (): void => {
-			timeoutHandler(delays.socket, 'socket');
+			timeoutHandler(delays.socket!, 'socket');
```
SamHH: Shouldn't if if statement act as a type guard, thus removing the need for the non-null assertion?

sindresorhus: Already fixed in master: https://github.com/sindresorhus/got/commit/640c30bbe37fa399a6f9c96357d897f45e54f52f#diff-7ab9a52ac2fccf434c1a34ce48f2883bL95-R95

SamHH: Ah, my mistake! :+1:

sindresorhus:
Closing for lack of activity.

LinusU:
Sorry for dragging my feet on this, closing is definitely the right approach 👍 

Hopefully I'll get some time to work on this later ☺️

Issue #876, opened 2019-09-17
Got v10 beta feedback


sindresorhus:
This issue is for discussing the upcoming Got v10 version which is now in ~alpha~ beta.

Beta 1: https://github.com/sindresorhus/got/releases/tag/v10.0.0-beta.1

Let us know if you encounter any issues with upgrading or if anything is unclear in the release notes.

kevva:
Just an FYI, but after installing the `dist` folder looks like this:

![image](https://user-images.githubusercontent.com/709159/65034336-1c2b0e80-d947-11e9-994b-58c792931787.png)

sindresorhus:
@kevva Weird. I manually fixed it:

```
$ npm install got@10.0.0-alpha.1.1
```

or 

```
$ npm install got@next
```

matus123:
Hi, i would like to report some issues... i don't know if this was desired.

### cannot specify own statusCodes in retry object (typescript error)
```
const test = got.create({
            options: {
                retry: {
                    statusCodes: [401],
                },
            },
        });

        got('http://localhost:5000', {
            retry: {
                statusCodes: [401],
            },
        });
```
### afterResponse hooks cannot return promise (typescript error)
```
        const test = got.create({
            options: {
                hooks: {
                    afterResponse: [
                        async (response, retry) => {
                            const options = response.request.options;
                            await doSomethingAsync();
                            return retry(options);
                        },
                    ],
                },
            },
        });
```

jaulz:
I raised a PR for the second issue here: https://github.com/sindresorhus/got/pull/883

szmarczak:
@jaulz Great! What about `cannot specify own statusCodes in retry object (typescript error)`?

jaulz:
@szmarczak just created another PR for it 😊 Also could we export the types from a central place? Currently, it imports the types from multiple files and the dist folder which looks quite confusing and untypical.

```
import { CancelableRequest, Response } from 'got/dist/utils/types'
import { Got, GotStream } from 'got/dist/create'
import { ProxyStream } from 'got/dist/as-stream'
```

sindresorhus:
@jaulz Yes, we should re-export all relevant types in https://github.com/sindresorhus/got/blob/master/source/index.ts You should not have to go digging into the folder structure.

jaulz:
@sindresorhus is that something you are already working on or should I raise a PR?

sindresorhus:
[...]

PR #889, opened 2019-09-23
Export useful types for end-users

jaulz:
As discussed in https://github.com/sindresorhus/got/issues/876 this PR exports most of the types in the index file, e.g. instead of:
```
import { CancelableRequest, Response } from 'got/dist/utils/types'
import { Got, GotStream } from 'got/dist/create'
import { ProxyStream } from 'got/dist/as-stream'
```

it would be:
```
import { Got, GotStream, ProxyStream, CancelableRequest, Response } from 'got'
```

#### Checklist

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -64,3 +64,43 @@ export default got;
 // For CommonJS default export support
 module.exports = got;
 module.exports.default = got;
+
+// Export types
+export * from './utils/types';
+export {
+	Got,
+	GotStream,
+	ReturnResponse,
+	ReturnStream,
+	GotReturn
+} from './create';
+export {
+	ProxyStream
+} from './as-stream';
+export {
+	GotError,
+	CacheError,
+	RequestError,
+	ParseError,
+	HTTPError,
+	MaxRedirectsError,
+	UnsupportedProtocolError,
+	TimeoutError,
+	CancelError
+} from './errors';
+export {
+	InitHook,
+	BeforeRequestHook,
+	BeforeRedirectHook,
+	BeforeRetryHook,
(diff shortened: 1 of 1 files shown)


source/index.ts
```diff
@@ -64,3 +64,43 @@ export default got;
...
+export * from './utils/types';
+export {
+	Got,
+	GotStream,
+	ReturnResponse,
+	ReturnStream,
+	GotReturn
+} from './create';
+export {
+	ProxyStream
```
sindresorhus: Not sure `ProxyStream` is a good name to expose.

source/index.ts
```diff
@@ -64,3 +64,43 @@ export default got;
...
+	BeforeErrorHook,
+	AfterResponseHook,
+	HookType,
+	Hooks,
+	HookEvent
+} from './known-hook-events';
+export {
+	GetMethodRedirectCodes,
+	AllMethodRedirectCodes,
+	RequestAsEventEmitter
```
sindresorhus: Do you really need these?

source/index.ts
sindresorhus: I think this needs a clearer name.

jaulz: Shall we rename it to `ResponseStream`?

szmarczak: Sounds good to me.

jaulz: Better now?

sindresorhus: Yup

jaulz:
Thanks!

sindresorhus:
https://github.com/sindresorhus/got/releases/tag/v10.0.0-alpha.2

PR #890, opened 2019-09-23
Fix the build script

jaulz:
The recent release 10.0.0-alpha.2 does not point to the correct files in the dist folder because the `package.json` is required.

#### Checklist


sindresorhus:
See https://github.com/sindresorhus/got/issues/891, that's not it. It shouldn't even generate in in `dist/source`, but rather `dist`.

jaulz:
In my environment the build result looks like this:
image

This is also how it is defined in `tsconfig.json`. I'm not sure if I understood what you expect.

sindresorhus:
Run `npm install got@10.0.0-alpha.2` and look at `node_modules/got/dist`.

jaulz:
@sindresorhus strange indeed but even the new package does not have any source files

sindresorhus:
`npm install got@10.0.0-alpha.2.2` should be good

jaulz:
@sindresorhus okay, I think I found the issue. Since the `package.json` was required it was also compiled to `dist` and this somehow confused the `npm publish`. I just pushed a commit which reads the `package.json` file and parses it instead.

source/index.ts
```diff
@@ -1,7 +1,9 @@
-import packageJson = require('../package.json');
+import { readFileSync } from 'fs';
 import create from './create';
 import {Defaults} from './utils/types.js';
 
+const packageJson = JSON.parse(readFileSync('./../package.json', { encoding: 'utf8' }));
```
sindresorhus: You have to use `__dirname` and path.join here.

jaulz: okay, done 😊

sindresorhus:
I think you need to update https://github.com/sindresorhus/got/blob/76ab7e0a38b0068081c9fbb867ec6be285ca7aef/package.json#L7 too, to `dist/source`

sindresorhus:
You're correct for this PR, but master is different. Note this change: https://github.com/sindresorhus/got/commit/640c30bbe37fa399a6f9c96357d897f45e54f52f#diff-e5e546dd2eb0351f813d63d1b39dbc48R12-R14

jaulz:
Okay, I fixed it now.

sindresorhus:
Thanks for fixing this 🙌

PR #893, opened 2019-10-01
Fix TypeScript definition for projects not using the DOM types

SamHH:
Fixes [this](https://github.com/sindresorhus/got/issues/876#issuecomment-537127127).

`url` comes from `@types/node`, which is a devDependency, and more importantly should be present in any project using this library as this library targets Node.js.

### DIFF
--- a/source/utils/types.ts
+++ b/source/utils/types.ts
@@ -2,7 +2,7 @@ import http = require('http');
 import https = require('https');
 import {Readable as ReadableStream} from 'stream';
 import PCancelable = require('p-cancelable');
-import {URL} from 'url';
+import {URL, URLSearchParams} from 'url';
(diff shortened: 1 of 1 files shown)


source/utils/types.ts
```diff
@@ -139,7 +139,7 @@ export interface Options extends Except<https.RequestOptions, 'agent' | 'timeout
 	timeout?: number | Delays;
 	dnsCache?: Map<string, string> | Keyv | false;
 	url?: URL | string;
-	searchParams?: Record<string, string | number | boolean | null> | URLSearchParams | string;
+	searchParams?: Record<string, string | number | boolean | null> | import('url').URLSearchParams | string;
```
szmarczak:
```
import {URLSearchParams} from 'url';
```

seems to be a better solution

PR #908, opened 2019-11-01
Fix `Stream.pipeline(...)` usage and properly detect aborted requests

### LINKED ISSUE #907: Promise never resolved/ rejected if connection is accepted but socket is terminated
#### Describe the bug

 - Node.js version: N/A
 - OS & version: N/A

#### Actual behavior

Promise is never resolved/ rejected.

#### Expected behavior

Promise resolved/ rejected.

#### Code to reproduce

https://repl.it/@gajus/CapitalDangerousMultitasking

szmarczak:
Fixes #907

### DIFF
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -194,7 +194,7 @@ export default (options: NormalizedOptions, input?: TransformStream) => {
 
 			currentRequest = request;
 
-			request.on('error', error => {
+			const onError = (error: Error): void => {
 				if (typeof request.aborted === 'number' || error.message === 'socket hang up') {
 					return;
 				}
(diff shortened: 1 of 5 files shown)


source/request-as-event-emitter.ts
```diff
@@ -208,7 +208,22 @@ export default (options: NormalizedOptions, input?: TransformStream) => {
...
+				}
+
+				// No need to attach an error handler here,
+				// as `stream.pipeline(...)` doesn't remove this handler
+				// to allow stream reuse.
+
+				request.emit('upload-complete');
+			};
+
+			request.on('error', onError);
```
szmarczak: It needs to be here in case `timer(...)` or `timedOut(...)` throws...

sindresorhus:
The timeout tests seem to fail on Node.js 10. I confirmed this locally too.

PR #910, opened 2019-11-01
Don't merge `context`, `body`, `json`, `form` & fix types

### LINKED ISSUE #897: 'context' keyword is stripped out from request body
#### Describe the bug

 - Node.js version: 12.11.0
 - OS & version: ubuntu 18.04
 - got version: 10.0.0-alpha.2.2

`context` key is stripped from body.

#### Actual behavior
when we include `context` in our request body it is stripped out from outgoing request.
...

#### Code to reproduce

```js
const got = require('got');
(async () => {
  let resp;
  try {

    resp = await got('http://google.com',{
      method : 'POST',
      json   : {
        context : { a: 'test'},
        a       : { b: 'asdasd'}
      }
    });
  } catch (e) {
    console.log(JSON.parse(e.options.body));
  }

  console.log(resp);
})();
```

#### Checklist

szmarczak:
Fixes #897

### DIFF
--- a/test/arguments.ts
+++ b/test/arguments.ts
@@ -15,7 +15,7 @@ test('`url` is required', async t => {
 		got(),
 		{
 			instanceOf: TypeError,
-			message: 'Parameter `url` must be a string or object, not undefined'
+			message: 'Parameter `url` must be a string or an Object, not undefined'
 		}
 	);
 });
@@ -174,7 +174,7 @@ test('throws TypeError when `options.hooks` is not an object', async t => {
 		got('https://example.com', {hooks: 'not object'}),
 		{
 			instanceOf: TypeError,
-			message: 'Parameter `hooks` must be an object, not string'
+			message: 'Parameter `hooks` must be an Object, not string'
 		}
 	);
 });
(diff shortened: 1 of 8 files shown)


test/arguments.ts
```diff
@@ -185,7 +185,7 @@ test('throws TypeError when known `options.hooks` value is not an array', async
 		got('https://example.com', {hooks: {beforeRequest: {}}}),
 		{
 			instanceOf: TypeError,
-			message: 'options.hooks.beforeRequest is not iterable'
+			message: 'Parameter `beforeRequest` must be an array, not Array'
 		}
```
sindresorhus: This error message makes no sense. It should be more helpful.

test/post.ts
```diff
@@ -169,7 +169,7 @@ test('throws when form body is not a plain object or array', async t => {
 });
 
 // See https://github.com/sindresorhus/got/issues/897
-test.failing('the `json` payload is not touched', withServer, async (t, server, got) => {
```
sindresorhus: Can you update the test to also check context, form, body, and hooks? Just to ensure we never regress the other properties either.

szmarczak: Done. The tests for `context` already exist in the `arguments.ts` file.

sindresorhus:
Great 🙌

PR #911, opened 2019-11-02
Do not import package.json

### LINKED ISSUE #891: Fix building Got
Currently, when publishing, the package only ends up containing a `dist/package.json` file...

However, running `npm build` normally at least puts the correct files in `dist/source`, although that's incorrect too, they should be in `dist`.

szmarczak:
Fixes #891

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 		"prepare": "npm run build"
 	},
 	"files": [
-		"dist"
+		"dist/source"
 	],
 	"keywords": [
 		"http",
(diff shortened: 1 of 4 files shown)


sindresorhus:
You need to remove it in the test too.

szmarczak:
Yeah, trying to fix this and that...

szmarczak:
Is it possible to test publishing somehow? E.g. using Travis?

sindresorhus:
Not sure. Something more useful would be a TS lint rule to prevent importing package.json.

PR #913, opened 2019-11-02
Allow method rewriting on redirects

### LINKED ISSUE #568: POST request does not follow 301/ 302 redirects
It is common for a HTTP POST/ PUT request to get a 3xx response. In this case, the expected behaviour is that the HTTP method is changed to GET and client is redirected to the target destination.

I was looking for a reference to back up this claim in the official specs, but the closest I found is Moz suggesting that in case of [permanent redirects](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#Permanent_redirections) (301) "[requests] may or may not be changed to GET."

Currently the following test case is failing:

```js
test('3xx redirect changes the request method to GET if the original request method is not safe to repeat (e.g. POST)', async (t) => {
  nock('http://applaudience.com')
    .post('/')
    .reply(301, 'foo', {
      location: 'http://applaudience.com/foo'
    });

  nock('http://applaudience.com')
    .get('/foo')
    .reply(200, 'bar');

  const response = await got('http://applaudience.com/', {
    method: 'post'
  });

  t.true(response.body === 'bar');
});

```

I would advocate for following the 3xx redirects using GET at all time, as this the behaviour that all major browsers implement. However, since this is not defined anywhere in the spec (as far as I can tell), the decision could be delegated to a `followRedirects` option as a boolean function (`followRedirects: (request, response) => { ... }`).

### LINKED ISSUE #885: `options.body` is not cleaned up when using the `json` option
#### Describe the bug

 - Node.js version: 12.9.0
 - OS & version: Windows 10
 - Got version: 10.0.0-alpha.1.1

#### Actual behavior
I am trying to refresh auth token when I get 401 status code, but after calling retryWithMergedOptions I am getting an error "The `body` option cannot be used with the `json` option or `form` option"

To bypass this behavior I had to set the body option to null in the updated options
...

#### Expected behavior
Retry method of the afterResponse hook should respect json mode
...

#### Code to reproduce

```js

const jsonClient = got.extend({
  prefixUrl: 'http://some-api.com',
  responseType: 'json',
  resolveBodyOnly: true,
  headers: {
    'Content-Type': 'application/json',
    Accept: 'application/json'
  },
  hooks: {
    afterResponse: [
      async (response, retryWithMergedOptions) => {
        if (response.statusCode === 401) {
          // Unauthorized
          const token = await getToken();
          const updatedOptions = {
           // body: null,
            headers: {
              token: token // Refresh the access token
            }
          };

          // Save for further requests
          jsonClient.defaults.options = got.mergeOptions(
            jsonClient.defaults.options,
            updatedOptions
          );

          // Make a new retry
          return retryWithMergedOptions(updatedOptions);
        }

        // No changes otherwise
        return response;
      }
    ]
  },
  mutableDefaults: true
});

const response = await jsonClient.post('some/url', {
  json: {
    some: 'data'
  }
});

```

#### Checklist

szmarczak:
Fixes #568
Fixes #885

### DIFF
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -20,13 +20,8 @@ import urlToOptions from './utils/url-to-options';
 import {RequestFunction, NormalizedOptions, Response, ResponseObject, AgentByProtocol} from './utils/types';
 import dynamicRequire from './utils/dynamic-require';
 
-export type GetMethodRedirectCodes = 300 | 301 | 302 | 303 | 304 | 305 | 307 | 308;
-export type AllMethodRedirectCodes = 300 | 303 | 307 | 308;
-export type WithoutBody = 'GET' | 'HEAD';
-
-const getMethodRedirectCodes: ReadonlySet<GetMethodRedirectCodes> = new Set([300, 301, 302, 303, 304, 305, 307, 308]);
-const allMethodRedirectCodes: ReadonlySet<AllMethodRedirectCodes> = new Set([300, 303, 307, 308]);
-const withoutBody: ReadonlySet<WithoutBody> = new Set(['GET', 'HEAD']);
+const redirectCodes: ReadonlySet<number> = new Set([300, 301, 302, 303, 304, 307, 308]);
+const withoutBody: ReadonlySet<string> = new Set(['GET', 'HEAD']);
 
 export interface RequestAsEventEmitter extends EventEmitter {
 	retry: <T extends Error>(error: T) => boolean;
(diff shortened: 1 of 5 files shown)


source/request-as-event-emitter.ts
```diff
@@ -20,13 +20,8 @@ import urlToOptions from './utils/url-to-options';
 import {RequestFunction, NormalizedOptions, Response, ResponseObject, AgentByProtocol} from './utils/types';
 import dynamicRequire from './utils/dynamic-require';
 
-export type GetMethodRedirectCodes = 300 | 301 | 302 | 303 | 304 | 305 | 307 | 308;
```
szmarczak: `305` has been deprecated

sindresorhus:
I think this new capability be documented somewhere.

szmarczak:
don't merge yet, I have to check sth

szmarczak:
okay you may merge now :ok_hand:

szmarczak:
wait not yet, i'll fix another issue with this one

PR #915, opened 2019-11-03
Use socket instead of connection

Cerberooo:
The request.connection and response.connection properties are now runtime deprecated. (https://github.com/sindresorhus/got/pull/827)

#### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - '13'
   - '12'
   - '10'
 after_success:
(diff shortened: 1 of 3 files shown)


sindresorhus:
Looks like the node typings are not updated yet. Just use a `// @ts-ignore` comment.

sindresorhus:
Can you also add Node 13 to Travis.yml to ensure it works.

Cerberooo:
I wonder why the error occurs because `socket` is there: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L297

PR #921, opened 2019-11-05
Major refactoring

### LINKED ISSUE #580: Use `http.request(url, options, callback)` when targeting Node 10.9.0
1. [ ] Use `http.request(url, options, callback)`
2. [x] Use [`Stream.pipeline(from, ..., to, callback)`](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback)
3. [ ] (Maybe) deprecate legacy `Url` usage

### LINKED ISSUE #792: Memory leak when using cache feature.
Hello, 

We are facing a memory leak in our production environment, using got and it's cache feature (with a KeyvRedis storage backend).

A heap dump revealed that most memory is retained by multiple error event listener on the single KeyvRedis instance.
Further analysis showed that for each request, a new CacheableRequest is instantiated : 
```
if (options.cache) {
	const cacheableRequest = new CacheableRequest(fn.request, options.cache);
	const cacheRequest = cacheableRequest(options, handleResponse);

	cacheRequest.once('error', error => {
		if (error instanceof CacheableRequest.RequestError) {
			emitError(new RequestError(error, options));
		} else {
			emitError(new CacheError(error, options));
		}
	});

	cacheRequest.once('request', handleRequest);
}
```
 Then, for each CacheableRequest, a new Keyv is instantiated : 
```
this.cache = new Keyv({
	uri: typeof cacheAdapter === 'string' && cacheAdapter,
	store: typeof cacheAdapter !== 'string' && cacheAdapter,
	namespace: 'cacheable-request'
});
```
KeyvRedis being an EventEmitter, a eventListener is registered.
```
if (typeof this.opts.store.on === 'function') {
	this.opts.store.on('error', err => this.emit('error', err));
}
```
The full context of the request being retained by this event listener.

Is there something wrong with the way we use got ? Is there a way to prevent CacheableRequest to be instantiated for each request ? 

We tried using got.create, but that didn't work.

Thanks.

### LINKED ISSUE #867: Unexpected behavior with `beforeRetry` and `baseUrl`
#### Actual behavior

The examples show updating `hostname` in the options, but if you originally used `baseUrl` and then try to modify `baseUrl` in your hook, it does not re-derive the url.  It seems like inconsistent/unexpected behavior.

#### Expected behavior

Run the same logic which sets hostname, path, etc if the hook modifies `baseUrl`

#### Code to reproduce

```js
const r = got.extend('https://foo.com/path', {
	hooks: {
		beforeRetry: [
			(options, response) => {
				options.baseUrl = 'https://bar.com/other-path';
			}
		]
	}
});

await r('/more/path')
```

In the above example I would expect two requests (if the first failed):

`https://foo.com/path/path/more/path`
`https://bar.com/other-path/more/path`

What happens is a second request to `https://foo.com/path/path/more/path`.

#### Checklist

### LINKED ISSUE #877: TypeScript - Only allow supported event names
Only allow supported event names where we use EventEmitter. To prevent typos and improve auto-complete.

```js
import got from 'got';

// Pass
got.stream(url).on('error', error => {});

// TS compile error
got.stream(url).on('err', error => {});
```

### LINKED ISSUE #882: Replace tough-cookie

I'm back from here. https://github.com/sindresorhus/got/issues/811
It is an issue of tough-cookie library. But the fix is in stale status for several months...

#### Describe the feature

So I'm thinking it would be really nice if `got` provides an abstract interface so I can use other cookie store than tough-cookie.

#### Checklist

### LINKED ISSUE #918: Retrying inside `afterResponse` hook should trigger `beforeRetry` hook
Just add

```js
					for (const hook of options.hooks.beforeRetry) {
						// eslint-disable-next-line no-await-in-loop
						await hook(options, error, retryCount);
					}
```

and make some tests.

https://github.com/sindresorhus/got/blob/b0dfc958cbe9878645582c3b98f225f11ad22908/source/as-promise.ts#L86

### LINKED ISSUE #925: Don't throw on `204 No Content` when parsing response
See https://github.com/sindresorhus/ky/issues/193

### LINKED ISSUE #930: When redirect fails, don't retry from scratch

### LINKED ISSUE #931: `promise.json()` doesn't work if the promise has been already resolved

szmarczak:
- [x] Deprecate legacy `Url` usage
- [x] Drop `query` option
- [x] Drop `got.mergeInstances`
- [x] Fixes #867 (if possible, replaces `prefixUrl` in the URL)
- [x] Fixes #792: [click for solution](https://github.com/sindresorhus/got/pull/921/files#diff-02301bc46e8b878f10e9a8339efb7de7R136)
- [x] Fixes #580
- [x] Fixes #925
- [x] Fixes #877
- [x] Fixes #882
- [x] Fixes #930
- [x] Fixes #931
- [x] Fixes #918
- [x] Renamed `options.stream` to `options.isStream`
- [x] Removed `got.create(...)`
- [x] Headers set to `undefined` are removed (bf206caf4b154ab36a8270d7128eef89569d2f33 was made because #496 was followed by #495 and #495 had different `merge` mechanism)
- [x] Fixed a bug that sometimes caused the Node process to hang: [click for solution](https://github.com/sindresorhus/got/pull/921/files#diff-7b439d0c6f823422fb34bf344e2e149cR56-R58)
- [x] Fixed a bug where cookies weren't reset on redirect between two different sites
- [x] Code cleanup

#### Checklist


source/utils/types.ts
```diff
@@ -116,15 +114,10 @@ export interface Delays {
 export type Headers = Record<string, string | string[]>;
 
 // The library overrides the type definition of `agent`, `host`, 'headers and `timeout`.
-export interface Options extends Except<https.RequestOptions, 'agent' | 'timeout' | 'host' | 'headers'> {
-	host?: string;
+export interface Options extends URLOptions {
```
szmarczak: TODO: Lots of HTTP options missing here.

source/utils/options-to-url.ts
```diff
@@ -0,0 +1,76 @@
...
+	searchParams?: Record<string, string | number | boolean | null> | URLSearchParams | string;
+	hash?: string;
+}
+
+const keys = [
+	'username',
+	'password',
+	'pathname',
+	'search',
+	'hash'
```
szmarczak: `port` needs to be here also

source/index.ts
```diff
@@ -1,6 +1,10 @@
 import create from './create';
 import {Defaults} from './utils/types.js';
 
+// A suggestion from @szmarczak:
+// Rename `stream` to `isStream`, because
+// people may think that `stream` is an input stream.
```
sindresorhus: Agreed!

szmarczak:
I'll also try to fix https://github.com/sindresorhus/got/issues/918

source/normalize-arguments.ts
```diff
@@ -141,6 +142,21 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
 		);
 	}
 
+	// `options.cookieJar`
+	if (options.cookieJar) {
+		let {setCookie, getCookieString} = options.cookieJar;
+
+		if (!is.asyncFunction(setCookie)) {
```
sindresorhus: This is not a correct check as it would not allow normal functions that return a promise.

readme.md
```diff
@@ -246,11 +246,19 @@ When set to `true` the promise will return the [Response body](#body-1) instead
...
+Type: `object` | [`tough.CookieJar` instance](https://github.com/salesforce/tough-cookie#cookiejar)
 
 **Note:** If you provide this option, `options.headers.cookie` will be overridden.
 
-Cookie support. You don't have to care about parsing or how to store them. [Example.](#cookies)
+Cookie support. You don't have to care about parsing or how to store them. [Example](#cookies).
+
+###### cookieJar.setCookie
+
+The function takes two arguments: `rawCookie` (`string`) and `url` (`string`). It needs to be an async function or some callback-style function.
```
sindresorhus: Let's only support async/promise-returning functions and just add a note that users needs to promisify callback-style functions themselves. Same with the below.

readme.md
```diff
@@ -1303,7 +1303,7 @@ const got = require('got');
 
 ### User Agent
 
-It's a good idea to set the `'user-agent'` header so the provider can more easily see how their resource is used. By default, it's the URL to this repo. You can omit this header by setting it to `null`.
+It's a good idea to set the `'user-agent'` header so the provider can more easily see how their resource is used. By default, it's the URL to this repo. You can omit this header by setting it to `undefined` (or `null`).
```
sindresorhus: Let's remove support for setting it as `null`. Also update the PR description so we don't forget to mention it in the release notes.

source/request-as-event-emitter.ts
```diff
@@ -235,37 +190,34 @@ export default (options: NormalizedOptions, input?: TransformStream) => {
...
-						request,
-						uploadComplete
-					);
+				if (options.method === 'POST' || options.method === 'PUT' || options.method === 'PATCH') {
+					if (is.nodeStream(httpOptions.body)) {
+						// `stream.pipeline(…)` handles `error` for us.
+						request.removeListener('error', onError);
+
+						stream.pipeline(
+							// @ts-ignore TS is dumb.
```
sindresorhus: This is actually my fault: https://github.com/sindresorhus/is/commit/aeb3f74d653cf745abf2460f0896207de00c6059

source/request-as-event-emitter.ts
```diff
@@ -274,16 +226,15 @@ export default (options: NormalizedOptions, input?: TransformStream) => {
...
 		};
 
-		const {timeout} = options;
-		delete options.timeout;
-
-		if (options.dnsCache) {
-			options.lookup = options.dnsCache.lookup;
-		}
+		if (is.object(options.cache)) {
+			// `cacheable-request` doesn't support Node 10 API, fallback.
```
sindresorhus: Can you open an `cacheable-request` issue and link to it in this comment?

source/request-as-event-emitter.ts
szmarczak: Ok, let me know when a new release is out :)

readme.md
szmarczak: Should we throw an error if the header is null?

source/normalize-arguments.ts
szmarczak: `setCookie.length !== 2` because the third argument would be a callback

readme.md
szmarczak: Then we need a special case for `tough-cookie`, since they are callback-style.

source/normalize-arguments.ts
```diff
@@ -103,143 +120,290 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
...
+		options.dnsCache = new CacheableLookup({cacheAdapter: options.dnsCache as Keyv});
+	}
+
+	// `options.method`
+	if (options.method) {
+		options.method = options.method.toUpperCase() as Method;
+	}
+
+	// Better memory management, so we don't have to generate a new object every time
+	if (Reflect.has(options, 'cache') && options.cache !== false) {
```
szmarczak: The `Reflect.has` check is useless here

szmarczak: The whole condition should be replaced with `is.object(options.cache)`

source/normalize-arguments.ts
```diff
@@ -103,143 +120,290 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
...
+export const normalizeRequestArguments = async (options: NormalizedOptions): Promise<NormalizedRequestArguments> => {
+	options = mergeOptions(options);
 
-		searchParams = options.query;
-		delete options.query;
+	let uploadBodySize: number | undefined;
+
+	// Serialize body
+	const {headers} = options;
+	const isForm = !is.nullOrUndefined(options.form);
```
szmarczak: This should be `!is.undefined(...)`, as `null` is still valid JSON.

source/normalize-arguments.ts
```diff
@@ -103,143 +120,290 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
...
+		throw new TypeError(`The \`${options.method}\` method cannot be used with a body`);
 	}
 
-	if (is.nonEmptyString(searchParams) || is.nonEmptyObject(searchParams) || (searchParams && searchParams instanceof URLSearchParams)) {
-		if (!is.string(searchParams)) {
-			if (!(searchParams instanceof URLSearchParams)) {
-				// @ts-ignore
-				validateSearchParams(searchParams);
+	if (isBody) {
+		if (isForm || isJSON) {
```
szmarczak: Missing check: `isForm && isJSON`

source/normalize-arguments.ts
```diff
@@ -103,143 +120,290 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
...
+			throw new TypeError('The `body` option cannot be used with the `json` option or `form` option');
+		}
+
+		if (is.object(options.body) && isFormData(options.body)) {
+			// Special case for https://github.com/form-data/form-data
+			if (!Reflect.has(headers, 'content-type')) {
```
szmarczak: `isFormData` could be `Reflect.has(options.body, 'getBoundary')`

source/normalize-arguments.ts
```diff
@@ -103,143 +120,290 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
...
-		const matches = /(?<socketPath>.+?):(?<path>.+)/.exec(options.path);
+	// UNIX sockets
+	if (options.url.hostname === 'unix') {
+		const matches = /(?<socketPath>.+?):(?<path>.+)/.exec(options.url.pathname);
 
 		if (matches?.groups) {
 			const {socketPath, path} = matches.groups;
+
+			// It's a bug!
+			// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
szmarczak: @sindresorhus can you take a look at this?

readme.md
szmarczak: How do you want to detect `tough-cookie` since the dependency is optional? :P 
I mean we can support callback-style, but the readme would say the opposite.

source/request-as-event-emitter.ts
szmarczak: I'm gonna rewrite `cacheable-request` completely and send a PR (it'll be a major version bump) :)

readme.md
szmarczak: done

readme.md
szmarczak: done

szmarczak:
Okay, the only thing to do is to review documentation. Make necessary changes, then it's good to merge.

source/normalize-arguments.ts
```diff
@@ -405,5 +407,8 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 
 	delete options.timeout;
 
+	// `http-cache-semantics` check this
```
sindresorhus: `check` => `checks`

readme.md
```diff
@@ -862,7 +900,57 @@ Options are deeply merged to a new object. The value of each key is determined a
...
+
+An array of functions. You execute them directly by calling `got()`. They are some sort of "global hooks" - these functions are called first. The last handler (*it's hidden*) is either [`asPromise`](../source/as-promise.ts) or [`asStream`](../source/as-stream.ts), depending on the `options.isStream` property.
+
+Each handler takes two arguments:
+
+###### [options](readme.md#options)
+
+###### next()
+
+Returns a `Promise` or a `Stream` depending on [`options.isStream`](../readme.md#isstream).
```
```suggestion
Returns a `Promise` or a `Stream` depending on [`options.isStream`](#isstream).
```

readme.md
```diff
@@ -862,7 +900,57 @@ Options are deeply merged to a new object. The value of each key is determined a
...
+The Got defaults used in that instance.
+
+##### [options](readme.md#options)
+
+##### handlers
+
+Type: `Function[]`<br>
+Default: `[]`
+
+An array of functions. You execute them directly by calling `got()`. They are some sort of "global hooks" - these functions are called first. The last handler (*it's hidden*) is either [`asPromise`](../source/as-promise.ts) or [`asStream`](../source/as-stream.ts), depending on the `options.isStream` property.
```
```suggestion
An array of functions. You execute them directly by calling `got()`. They are some sort of "global hooks" - these functions are called first. The last handler (*it's hidden*) is either [`asPromise`](source/as-promise.ts) or [`asStream`](source/as-stream.ts), depending on the `options.isStream` property.
```

readme.md
```diff
@@ -862,7 +900,57 @@ Options are deeply merged to a new object. The value of each key is determined a
 
 Type: `object`
 
-The default Got options used in that instance.
+The Got defaults used in that instance.
+
+##### [options](readme.md#options)
```
```suggestion
##### [options](#options)
```

readme.md
```diff
@@ -862,7 +900,57 @@ Options are deeply merged to a new object. The value of each key is determined a
...
+##### handlers
+
+Type: `Function[]`<br>
+Default: `[]`
+
+An array of functions. You execute them directly by calling `got()`. They are some sort of "global hooks" - these functions are called first. The last handler (*it's hidden*) is either [`asPromise`](../source/as-promise.ts) or [`asStream`](../source/as-stream.ts), depending on the `options.isStream` property.
+
+Each handler takes two arguments:
+
+###### [options](readme.md#options)
```
```suggestion
###### [options](#options)
```

readme.md
```diff
@@ -862,7 +900,57 @@ Options are deeply merged to a new object. The value of each key is determined a
...
+
+const jsonGot = got.create(settings);
+```
+
+##### mutableDefaults
+
+Type: `boolean`<br>
+Default: `false`
+
+A read-only boolean describing whether the defaults are mutable or not. If set to `true`, you can [update headers over time](../readme.md#hooksafterresponse), for example, update an access token when it expires.
```
```suggestion
A read-only boolean describing whether the defaults are mutable or not. If set to `true`, you can [update headers over time](#hooksafterresponse), for example, update an access token when it expires.
```

source/normalize-arguments.ts
sindresorhus: Yeah, looks like a bug in typescript-eslint. It's so buggy I don't want to spend time even opening an issue. Let's leave it like this and open an issue there sometime in the future if it's still not fixed.

source/utils/types.ts
sindresorhus: Add a TODO comment in the code if you intend to fix this later.

source/utils/types.ts
```diff
@@ -155,29 +153,20 @@ export interface Options extends Except<https.RequestOptions, 'agent' | 'timeout
...
 	headers: Headers;
 	hooks: Hooks;
 	timeout: Delays;
 	dnsCache?: CacheableLookup | false;
 	retry: Required<RetryOptions>;
-	readonly prefixUrl?: string;
+	prefixUrl: string;
 	method: Method;
+	url: URL;
+	cacheableRequest?: (opts: string | URL | http.RequestOptions, cb?: (response: http.ServerResponse | ResponseLike) => void) => CacheableRequest.Emitter;
```
```suggestion
	cacheableRequest?: (options: string | URL | http.RequestOptions, callback?: (response: http.ServerResponse | ResponseLike) => void) => CacheableRequest.Emitter;
```

source/utils/types.ts
```diff
@@ -186,22 +175,29 @@ export interface ExtendedOptions extends Options {
...
+	_rawHandlers?: HandlerFunction[];
 	mutableDefaults: boolean;
 }
 
-export type URLOrOptions = URLArgument | (Options & {url: URLArgument});
+export type URLOrOptions = Options | string;
+
+export interface Progress {
+	percent: number;
+	transfered: number;
```
```suggestion
	transferred: number;
```

test/arguments.ts
```diff
@@ -29,6 +29,11 @@ test('`url` should be utf-8 encoded', async t => {
 	);
 });
 
+test('throws if no arguments provided', async t => {
+	// @ts-ignore This is on purpose.
+	await t.throwsAsync(got(), TypeError, 'Missing `url` argument');
```
sindresorhus:
```suggestion
	await t.throwsAsync(got(), {
		instanceOf: TypeError,
		message: 'Missing `url` argument'
	});
```

test/arguments.ts
```diff
@@ -236,19 +242,36 @@ test('backslash in the end of `prefixUrl` option is optional', withServer, async
 	t.is(body, '/test/foobar');
 });
 
-test('throws when trying to modify `prefixUrl` after options got normalized', async t => {
-	const instanceA = got.create({
-		options: {prefixUrl: 'https://example.com'},
+test('`prefixUrl` can be changed if the URL contains the old one', withServer, async (t, server) => {
```
sindresorhus: Support for changing `prefixUrl` needs to be documented.

source/normalize-arguments.ts
szmarczak: Good spot! Didn't know that `semantics` is uncountable.

readme.md
```diff
@@ -528,6 +528,8 @@ got.post('https://example.com', {
 });
 ```
 
+**Note:** When retrying in a `afterResponse` hook, all remaining `beforeRetry` hooks will be called without the `error` and `retryCount` arguments.
```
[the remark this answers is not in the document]
szmarczak: Currently it's by design. The retry count is stored in `request-as-event-emitter.ts`. And it's a private variable so I cannot access it.

#932 should fix this

source/known-hook-events.ts
```diff
@@ -25,7 +25,7 @@ export type BeforeRedirectHook = (options: NormalizedOptions, response: Response
 /**
 Called with normalized [request options](https://github.com/sindresorhus/got#options), the error and the retry count. Got will make no further changes to the request. This is especially useful when some extra work is required before the next try.
 */
-export type BeforeRetryHook = (options: NormalizedOptions, error: Error | GotError | ParseError | HTTPError | MaxRedirectsError, retryCount: number) => void | Promise<void>;
+export type BeforeRetryHook = (options: NormalizedOptions, error?: Error | GotError | ParseError | HTTPError | MaxRedirectsError, retryCount?: number) => void | Promise<void>;
```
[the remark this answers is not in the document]
szmarczak: Any ideas on how to solve this are welcome :)

source/normalize-arguments.ts
```diff
@@ -405,8 +408,18 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 		options.request = electron.net.request ?? electron.remote.net.request;
 	}
 
+	// We're not compatible
```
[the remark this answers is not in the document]
szmarczak: = Got

Will edit this.

source/progress.ts
```diff
@@ -75,8 +71,7 @@ export function uploadProgress(request: ClientRequest, emitter: EventEmitter, up
 		const onSocketConnect = (): void => {
 			progressInterval = setInterval(() => {
 				const lastUploadedBytes = uploadedBytes;
-				/* istanbul ignore next: see #490 (occurs randomly!) */
-				const headersSize = (request as any)._header ? Buffer.byteLength((request as any)._header) : 0;
+				const headersSize = Buffer.byteLength((request as any)._header);
```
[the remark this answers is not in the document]
szmarczak: Yup, double checked that. Is always initialized on `request.end()`. I couldn't find any code that would overwrite it ;)

source/utils/types.ts
```diff
@@ -41,10 +41,10 @@ export type ErrorCode =
 	| 'ENETUNREACH'
 	| 'EAI_AGAIN';
 
-export type ResponseType = 'json' | 'buffer' | 'text';
+export type ResponseType = 'json' | 'buffer' | 'text' | '';
```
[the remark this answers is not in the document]
szmarczak: `''` is the default one. It will follow `options.encoding`. For example if `options.encoding` is `null`, it will return a Buffer. If `responseType` is `text`, then that buffer will be converted into String.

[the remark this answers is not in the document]
szmarczak: I'm not sure if I documented that.

source/utils/types.ts
```diff
@@ -41,10 +41,10 @@ export type ErrorCode =
...
 
 export interface Response extends http.IncomingMessage {
-	body: Buffer | string | any;
+	body: any;
```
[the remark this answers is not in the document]
szmarczak: Will fix this. If `responseType` is set to `''` people can give their own objects as the response in `afterResponse` hook.

test/arguments.ts
szmarczak: Meh, I did this a few times 😅

source/utils/types.ts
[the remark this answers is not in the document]
sindresorhus: Kinda weird to use an empty string for that. I think I would prefer either `'auto'`, `'default'`, (or a better name), or `undefined`.

source/progress.ts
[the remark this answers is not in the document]
sindresorhus: But you can't guarantee that won't change in a future Node.js version. I'm being paranoid as it's an internal variable (the `_` prefix), so I think it would be smart to stay on the safe side here.

readme.md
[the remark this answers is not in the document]

source/known-hook-events.ts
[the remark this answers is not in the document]
sindresorhus: Move it to `types.ts` with a name and then import it where needed?

source/normalize-arguments.ts
szmarczak: You mean not plural? yeah.

source/utils/types.ts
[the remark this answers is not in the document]
szmarczak: Or make the new behavior the default one. I mean let's use `options.encoding` and then convert it into either `json`, `buffer` or `text`. What do you think?

source/progress.ts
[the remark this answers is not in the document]
szmarczak: But then it would fail silently. It's better to fail loudly IMO

source/utils/types.ts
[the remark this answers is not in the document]
szmarczak: But then people couldn't return their own objects, so maybe lets go with `'default'`?

source/utils/types.ts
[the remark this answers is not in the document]
szmarczak: > Will fix this. 

Actually I can't fix this. The type of `body` is dependent on `options.responseType`, which can change over time. IMO the "correct" solution would be:

1. `Response` -> `Response<BodyType>`
2. Replace

```typescript
	(url: URLOrOptions | Options & {isStream?: false}, options?: Options & {isStream?: false}): CancelableRequest<Response>;
```

with

```typescript
	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly?: false, responseType?: 'default'}, options?: Options & {isStream?: false, resolveBodyOnly?: false, responseType?: 'default'}): CancelableRequest<Response<any>>;

	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'text'}, options?: Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'text'}): CancelableRequest<Response<string>>;
	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'json'}, options?: Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'json'}): CancelableRequest<Response<object>>;
	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'buffer'}, options?: Options & {isStream?: false, resolveBodyOnly?: false, responseType: 'buffer'}): CancelableRequest<Response<Buffer>>;

	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly: true, responseType: 'text'}, options?: Options & {isStream?: false, resolveBodyOnly: true, responseType: 'text'}): CancelableRequest<string>;
	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly: true, responseType: 'json'}, options?: Options & {isStream?: false, resolveBodyOnly: true, responseType: 'json'}): CancelableRequest<object>;
	(url: URLOrOptions | Options & {isStream?: false, resolveBodyOnly: true, responseType: 'buffer'}, options?: Options & {isStream?: false, resolveBodyOnly: true, responseType: 'buffer'}): CancelableRequest<Buffer>;
```

source/known-hook-events.ts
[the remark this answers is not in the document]
szmarczak: Thanks, didn't know it was that simple :joy: I thought it would require a new function or something...

test/arguments.ts
szmarczak: Do I really need to change this? It looks so slim...

sindresorhus: You don't have to. Was just a suggestion to improve readability.

source/utils/types.ts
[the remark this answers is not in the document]
sindresorhus: > But then people couldn't return their own objects, so maybe lets go with 'default'?

Sure

source/utils/types.ts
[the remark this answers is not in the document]
sindresorhus: Yes, that's the correct way to handle it.

source/progress.ts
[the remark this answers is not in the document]
sindresorhus: So do a `console.warn` then. Crashing hard over an implementation detail is not a good experience. Generally, when using private APIs, you're expected to properly guard it. Node.js is free to change it in a patch version and it will affect people before we can get to patch it. It's better to just return a dummy value and warn than crashing hard.

source/utils/types.ts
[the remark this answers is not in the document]
szmarczak: will do :fire:

source/progress.ts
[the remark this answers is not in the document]
szmarczak: will do

readme.md
```diff
@@ -242,11 +242,11 @@ const instance = got.extend({
 ###### responseType
 
 Type: `string`<br>
-Default: `text`
+Default: `'default'`
```
sindresorhus: You should probably explain what `default` actually means.

source/progress.ts
```diff
@@ -71,6 +72,19 @@ export function uploadProgress(request: ClientRequest, emitter: EventEmitter, up
 		const onSocketConnect = (): void => {
 			progressInterval = setInterval(() => {
 				const lastUploadedBytes = uploadedBytes;
+
+				/* istanbul ignore next: future versions of Node may not have this property */
+				if (!is.string((request as any)._header)) {
+					clearInterval(progressInterval);
+
+					const url = new URL('https://github.com/sindresorhus/got/issues/new');
+					url.searchParams.set('title', '`request._header` is not present');
```
sindresorhus: Later, we should open a Node.js issue about getting something official for this instead of having to use a private property.

source/utils/types.ts
```diff
@@ -43,10 +43,10 @@ export type ErrorCode =
 	| 'ENETUNREACH'
 	| 'EAI_AGAIN';
 
-export type ResponseType = 'json' | 'buffer' | 'text' | '';
+export type ResponseType = 'json' | 'buffer' | 'text' | 'default';
 
-export interface Response extends http.IncomingMessage {
-	body: any;
+export interface Response<BodyType = any> extends http.IncomingMessage {
```
sindresorhus: `any` => `unknown`

readme.md
szmarczak: will do

source/progress.ts
szmarczak: Will add a TODO here

szmarczak: I'll do it now :)

szmarczak: https://github.com/nodejs/node/issues/30496

szmarczak:
TODO: `preNormalizeArguments` should merge defaults - that can be done in a separate PR :)

readme.md
```diff
@@ -246,7 +246,14 @@ Default: `'default'`
 
 **Note:** When using streams, this option is ignored.
 
-Parsing method used to retrieve the body from the response. Can be `'default'`, `'text'`, `'json'` or `'buffer'`. The promise has `.json()` and `.buffer()` and `.text()` functions which set this option automatically.
+Parsing method used to retrieve the body from the response.
+
+- `'default'` - if `options.encoding` is `null`, the body will be a Buffer. Otherwise it will be a string unless it's overwritten in a `afterResponse` hook,
```
sindresorhus: Should we remove the `null` thing to get `buffer` as the user can just do `responseType: 'buffer'` instead? Which is also much clearer.

sindresorhus:
Really nice work on this!! 🙌 I'm glad you got to this before the release.

readme.md
szmarczak: Sure, I'll send a PR.

PR #928, opened 2019-11-13
Make TypeScript types conforms to strict mode

pmmmwh:
This PR tries to make all code within the `source` folder conforms to strict mode.

There are some parts that I had to trace the source code and make educated guesses as I am uncertain with the actual available values, so please do tell me if I am wrong! Also - there are some places we could either use an intermediate variable or use non-null assertions. Please tell me which one you would prefer to maintain consistency 👀

#### Next Steps
- [x] Confirm source code is running fine
- [x] Fix types in tests
- [ ] Fix ts-ignore and typescript eslint errors
- [ ] Check for potential `const` assertions that can be applied

#### Checklist


szmarczak:
Beware of #921 - if it gets merged you'll have to rework many many things :P I'd prefer if you waited for its merge.

pmmmwh:
> Beware of #921 - if it gets merged you'll have to rework many many things :P I'd prefer if you waited for its merge.

Hmm. Actually I have kinda finished the whole process of typing stuff. I think maybe I'll push first, then wait for its merge, and then tackle the remaining parts that are broken after that? I wouldn't expect a review to be coming soon, but maybe some of this code can be referenced in that PR too?

szmarczak:
TypeScript needs to ignore files in `tests/*` because they are not TypeScript yet.

source/as-promise.ts
```diff
@@ -13,7 +13,7 @@ type ResponseReturn = Response | Buffer | string | any;
 
 export const isProxiedSymbol: unique symbol = Symbol('proxied');
 
-export default function asPromise(options: NormalizedOptions): CancelableRequest<Response> {
+export default function asPromise(options: NormalizedOptions): CancelableRequest<Response> | Response['body'] {
```
szmarczak: Good catch! :)

source/as-promise.ts
```diff
@@ -56,7 +56,7 @@ export default function asPromise(options: NormalizedOptions): CancelableRequest
 				return;
 			}
 
-			if (response.req?.aborted) {
+			if (response.req.aborted) {
```
szmarczak:
```suggestion
			if (response.req?.aborted) {
```

Because the cached response doesn't have `req` property

source/as-promise.ts
```diff
@@ -107,7 +107,7 @@ export default function asPromise(options: NormalizedOptions): CancelableRequest
 
 			if (statusCode !== 304 && (statusCode < 200 || statusCode > limitStatusCode)) {
 				const error = new HTTPError(response, options);
-				if (emitter.retry(error) === false) {
+				if (!emitter.retry(error)) {
```
szmarczak: :+1:

source/calculate-retry-delay.ts
```diff
@@ -1,40 +1,47 @@
...
-	if (response && Reflect.has(response.headers, 'retry-after') && retryAfterStatusCodes.has(response.statusCode)) {
-		let after = Number(response.headers['retry-after']);
-		if (is.nan(after)) {
-			after = Date.parse(response.headers['retry-after']) - Date.now();
-		} else {
-			after *= 1000;
-		}
-
-		if (after > retryOptions.maxRetryAfter) {
+	if (error instanceof GotError) {
```
szmarczak: Isn't this always true?

source/create.ts
```diff
@@ -26,15 +27,17 @@ export type HTTPAlias =
 	| 'head'
 	| 'delete';
 
-export type ReturnResponse = (url: URLArgument | Options & {stream?: false; url: URLArgument}, options?: Options & {stream?: false}) => CancelableRequest<Response>;
-export type ReturnStream = (url: URLArgument | Options & {stream: true; url: URLArgument}, options?: Options & {stream: true}) => ProxyStream;
-export type GotReturn = ProxyStream | CancelableRequest<Response>;
+export type ReturnResponse = (url: URLOrOptions | Options & {stream?: false}, options?: Options & {stream?: false}) => ReturnType<typeof asPromise>;
+export type ReturnStream = (url: URLOrOptions | Options & {stream?: true}, options?: Options & {stream?: true}) => ReturnType<typeof asStream>;
+export type GotReturn = ReturnType<ReturnResponse> | ReturnType<ReturnStream>;
```
szmarczak: :+1:

source/create.ts
```diff
@@ -131,24 +144,24 @@ const create = (nonNormalizedDefaults: Defaults): Got => {
...
+		const options: Array<PartialDeep<DefaultOptions> & Options> = [defaults.options];
 		const handlers: HandlerFunction[] = [...defaults.handlers];
-		let mutableDefaults: boolean;
+		let mutableDefaults: boolean | undefined;
 
 		for (const value of instancesOrOptions) {
-			if (Reflect.has(value, 'defaults')) {
-				options.push((value as Got).defaults.options);
-				handlers.push(...(value as Got).defaults.handlers.filter(handler => handler !== defaultHandler));
+			if (isGotInstance(value)) {
```
szmarczak: :+1: nice work here

source/utils/timed-out.ts
```diff
@@ -85,11 +85,12 @@ export default (request: ClientRequest, delays: Delays, options: TimedOutOptions
 	}
 
 	if (typeof delays.socket !== 'undefined') {
+		const {socket} = delays;
```
szmarczak: No need to extract `socket` for only one use. Same below.

source/calculate-retry-delay.ts
pmmmwh: I am not sure - in tests I see a check on user-defined custom error codes, so it might not be?

source/as-promise.ts
pmmmwh: Okies - will reflect this change in the `Response` type too!

source/create.ts
pmmmwh: Actually, this can be further slimmed down in the instance type for `Got` below. One could use `Parameters<T>` and `ResponseType<T>`, but I am not sure if it will hurt readability in the users' side.

source/utils/timed-out.ts
pmmmwh: The reason I did this change is that both TypeScript (and me) is confused on whether issues due to time will occur, since the code below is defined within a callback. If the object will be always defined, maybe I can switch to using non-null assertions instead.

pmmmwh:
> TypeScript needs to ignore files in `tests/*` because they are not TypeScript yet.

Actually after the changes in the source types tests aren't too difficult to type anymore, so I also typed them. I verified by running `tsc --noEmit --pretty` against the whole project and no errors are reported, so I am not sure if I am confusing `ts-node` by doing nothing, because the project compiles ...

source/calculate-retry-delay.ts
szmarczak: But it's still GotError, isn't it? It's created in `onError` function IIRC.

source/as-promise.ts
pmmmwh: Done :)

source/calculate-retry-delay.ts
pmmmwh: I checked the source code - it seems that there is only one place calling `calculateRetryDelay`: `emitter.retry`. It is being called in two places, with the possible error choices `TimeoutError`, `HTTPError` and `RequestError`. I think you are right - I will update to reflect this in the code.

source/utils/timed-out.ts
pmmmwh: Fixed.

pmmmwh:
So I fixed the issue - apparently `ts-node` will need to use `typeRoots` to correctly read type definitions, so I also did that move in the PR. Now I wonder why tests are failing on CI for node 10 and 12, because they apparently passes on my local machine 🤔

source/utils/types.ts
```diff
@@ -113,10 +119,27 @@ export interface Delays {
 	request?: number;
 }
 
-export type Headers = Record<string, string | string[]>;
+export type Headers = Record<string, string | string[] | null | undefined>;
```
```suggestion
export type Headers = Record<string, string | string[] | undefined>;
```

We decided to only allow `undefined` to clear headers and not `null`.

test/cookies.ts
```diff
@@ -131,11 +130,11 @@ test('no unhandled errors', async t => {
 	const options = {
 		cookieJar: {
 			setCookie: () => {},
-			getCookieString: (_, cb) => cb(new Error(message))
-		}
+			getCookieString: (_: string, cb: (err: Error) => void) => cb(new Error(message))
```
sindresorhus:
```suggestion
			getCookieString: (_: string, callback: (error: Error) => void) => callback(new Error(message))
```

package.json
```diff
@@ -39,8 +39,6 @@
 	"dependencies": {
 		"@sindresorhus/is": "^1.0.0",
 		"@szmarczak/http-timer": "^3.0.0",
-		"@types/cacheable-request": "^6.0.1",
-		"@types/tough-cookie": "^2.3.5",
```
sindresorhus: I think these needs to be dependencies as they're used in the actual code.

szmarczak: https://stackoverflow.com/a/45176764

szmarczak: Only TypeScript users need this.

source/errors.ts
```diff
@@ -45,41 +46,42 @@ export class ReadError extends GotError {
...
-	readonly response: Response;
+	readonly response!: Response;
 
 	constructor(error: Error, response: Response, options: NormalizedOptions) {
-		super(`${error.message} in "${format(options as unknown as URL)}"`, error, options);
+		super(`${error.message} in "${format(options as unknown as URL | UrlObject)}"`, error, options);
 		this.name = 'ParseError';
 
 		Object.defineProperty(this, 'response', {
+			enumerable: false,
```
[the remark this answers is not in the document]
pmmmwh: Since TS 3.7 with the `useDefineForClassFields` flag on, all class defined properties will be initialized with `enumerable: true`, `configurable: true`, and `writable: true`. I didn't realize I could mitigate this with the declare handler to mitigate that, will fix that in a commit.

source/known-hook-events.ts
```diff
@@ -102,6 +102,7 @@ export interface Hooks {
 	@default []
 	*/
 	afterResponse?: AfterResponseHook[];
+	[hookName: string]: any;
```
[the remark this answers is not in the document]
pmmmwh: I added this after seeing a test on "allows extra keys in `options.hooks`".

Is the correct behavior as follows?
1. Allow custom keys passed into hooks
2. Normalized/merged options only contains known hook events

source/errors.ts
```diff
@@ -45,41 +46,42 @@ export class ReadError extends GotError {
 }
 
 export class ParseError extends GotError {
-	readonly response: Response;
+	readonly response!: Response;
```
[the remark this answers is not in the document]
pmmmwh: Typescript complains that response is not initialized and definitely assigned in the constructor, because it won't pickup `Object.defineProperty` calls as a property assignment. I added this since from the source it is known that this will be defined within the constructor.

source/utils/types.ts
pmmmwh: This will need to update the test on removing null headers - should I also do it?

package.json
pmmmwh: > Only TypeScript users need this.

Do you think I should include them in dependencies since our exported type definitions includes them?

source/errors.ts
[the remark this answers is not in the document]
pmmmwh: OK ... so I realized why I didn't do that earlier. Apparently this feature is [broken](https://github.com/microsoft/TypeScript/issues/34972). Should we wait on that, or just add a comment to ignore that for now and move on?

test/cookies.ts
pmmmwh: Fixed!

pmmmwh:
I am currently in the process of fixing merge conflicts. There's one thing I would like to ask though - can you enlighten me on why the global `URL` and `URLSearchParams` objects are preferred instead of the ones from the `url` module? I am just asking because I am curious, and this does generate quite a few cases where URL !== URL and URLSearchParams !== URLSearchParams.

szmarczak:
> URL !== URL and URLSearchParams !== URLSearchParams

`// @ts-ignore` that. It's a well-known bug in TypeScript. 

> why the global URL and URLSearchParams objects are preferred instead of the ones from the url module?

Because `require('url')` is not necessary to get URL and URLSearchParams. It's to prevent moot code.

package.json
szmarczak: I think that we shouldn't include the types as dependencies. They should be in devDependencies, because people that don't use TS will have to download 1.5MB+ data. People that want to use TS would need to `npm install @types/... @types/... @types/...`.

source/errors.ts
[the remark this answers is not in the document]
sindresorhus: > Should we wait on that, or just add a comment to ignore that for now and move on?

Just add a comment that links to https://github.com/microsoft/TypeScript/issues/34972 and move on.

source/errors.ts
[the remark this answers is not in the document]
sindresorhus: Alright. Thanks for explaining.

sindresorhus:
> Because require('url') is not necessary to get URL and URLSearchParams. It's to prevent moot code.

Explicitly importing them is only a legacy thing. But the main reason is consistency with browsers where they are only globals.

package.json
sindresorhus: Actually, I was not entirely correct. `@types/tough-cookie` can be a dev dependency as we don't directly depend on it, and it's up to the user to add it if they use `tough-cookie`. However, we do depend on `cacheable-request`, and we use its types, so I think we need to have `@types/cacheable-request` as a dependency, otherwise Got will not run when published and used in a TS project, right? Unless I'm missing something.

source/utils/types.ts
sindresorhus: Yes, thanks :)

source/known-hook-events.ts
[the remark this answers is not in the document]
sindresorhus: It was added in https://github.com/sindresorhus/got/commit/107756fc28dce579c521b15248f8bba730ae9e54. I think it was just a mistake. We definitely don't want to allow random keys. That makes no sense to me. @szmarczak ?

package.json
szmarczak: Many people don't use TS and for them it's only bloat. NPM should somehow detect typescript projects and install types automatically...

source/known-hook-events.ts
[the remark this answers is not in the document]
szmarczak: I agree with @sindresorhus since we have a `context` option for this :) the test need to be done removed then.

package.json
pmmmwh: > Actually, I was not entirely correct. `@types/tough-cookie` can be a dev dependency as we don't directly depend on it, and it's up to the user to add it if they use `tough-cookie`. However, we do depend on `cacheable-request`, and we use its types, so I think we need to have `@types/cacheable-request` as a dependency, otherwise Got will not run when published and used in a TS project, right? Unless I'm missing something.

I am going to do this. I think both of you are correct, but maintaining out-of-the-box working status is important for TS users too, so the bloat might be inevitable ...

source/as-promise.ts
```diff
@@ -28,16 +27,16 @@ const parseBody = (body: Response['body'], responseType: NormalizedOptions['resp
 	throw new Error(`Failed to parse body of type '${typeof body}' as '${responseType}'`);
 };
 
-export default function asPromise(options: NormalizedOptions) {
+export default function asPromise<T extends any>(options: NormalizedOptions): CancelableRequest<T> {
```
sindresorhus: `any` => `unknown`, if possible

source/known-hook-events.ts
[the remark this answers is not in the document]
pmmmwh: I have removed the key and added a comment on the test saying that we won't throw but the hook won't get passed in.

test/create.ts
```diff
@@ -27,7 +32,7 @@ test('supports instance defaults', withServer, async (t, server, got) => {
 			'user-agent': 'custom-ua-string'
 		}
 	});
-	const headers = await instance('').json();
+	const headers = await instance('').json() as Headers;
```
szmarczak:
```suggestion
	const headers = await instance('').json<Headers>();
```

szmarczak:
I start to hate TypeScript.

test/create.ts
pmmmwh: This couldn't be done because TS complains `Headers` is not assignable to `JsonValue`. I am still thinking of a way to solve that ...

I tried both ways below with no luck:
```js
const headers = await instance('').json<Headers>();
```

```js
const headers: Headers = await instance('').json();
```

pmmmwh:
Okay, with a lot of try and errors I made tests run on Node 10/12 on Travis.
One test failed. I checked the history and I'm pretty sure that I didn't touch that part of the code.

I used `JsonValue` to type JSON responses, but it seems to have some side-effect and will complain any object like interface !== JsonValue. I'll probably think about that tomorrow.

pmmmwh:
> I start to hate TypeScript.

I saw @sindresorhus tweet about it. TS is still short when to infer/track types when we are in a highly-mutable, very functional (with recursions) world, or when there are a lot of overloads to the same function. It does offer a lot of promise, but when a lot of stuff is unknown (especially true for request body), it is so tempting to just go with `any`.

source/as-promise.ts
```diff
@@ -27,7 +27,7 @@ const parseBody = (body: Response['body'], responseType: NormalizedOptions['resp
 	throw new Error(`Failed to parse body of type '${typeof body}' as '${responseType}'`);
 };
 
-export default function asPromise<T extends any>(options: NormalizedOptions): CancelableRequest<T> {
+export default function asPromise<T extends unknown = unknown>(options: NormalizedOptions): CancelableRequest<T> {
```
sindresorhus: I don't think you need to constrain these to `unknown`. The default should be enough: https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#breaking-changes I think it could be just `<T>(..`. Same in the other places we have this.

source/create.ts
```diff
@@ -23,38 +23,37 @@ export type HTTPAlias =
...
 type OptionsOfBufferResponseBody = Options & {isStream?: false; resolveBodyOnly?: false; responseType: 'buffer'};
 type ResponseBodyOnly = {resolveBodyOnly: true};
 
 interface GotFunctions {
 	// `asPromise` usage
-	(url: string | OptionsOfDefaultResponseBody, options?: OptionsOfDefaultResponseBody): CancelableRequest<Response>;
+	<T extends unknown = unknown>(url: string | OptionsOfDefaultResponseBody, options?: OptionsOfDefaultResponseBody): CancelableRequest<Response<T>>;
 	(url: string | OptionsOfTextResponseBody, options?: OptionsOfTextResponseBody): CancelableRequest<Response<string>>;
-	(url: string | OptionsOfJSONResponseBody, options?: OptionsOfJSONResponseBody): CancelableRequest<Response<object>>;
+	<T extends JsonValue = JsonValue>(url: string | OptionsOfJSONResponseBody, options?: OptionsOfJSONResponseBody): CancelableRequest<Response<T>>;
```
sindresorhus: Let's not constrain this. `JsonValue` type creates more problems than it solves. See: https://github.com/sindresorhus/ky/blob/2f37c3f999efb36db9108893b8b3d4b3a7f5ec45/index.d.ts#L303-L326 (and related links) Applies to other places `JsonValue` is used too.

source/utils/types.ts
```diff
@@ -117,7 +117,7 @@ export interface Delays {
 	request?: number;
 }
 
-export type Headers = Record<string, string | string[] | undefined>;
+export type Headers = Record<string, string | string[] | number | undefined>;
```
sindresorhus: I don't think we should support `number` here. Headers are strings, so I think it's better to force users to explicitly coerce/cast to a string themselves.

source/utils/types.ts
```diff
@@ -173,7 +174,7 @@ export interface Options extends PartialDeep<DefaultOptions>, URLOptions, Except
 	useElectronNet?: boolean;
 	form?: Record<string, any>;
 	json?: Record<string, any>;
-	context?: {[key: string]: unknown};
+	context?: Record<string, any>;
```
sindresorhus: `{[key: string]: any}`

test/cache.ts
```diff
@@ -44,7 +44,7 @@ test('cached response is re-encoded to current encoding option', withServer, asy
 	const firstEncoding = 'base64';
 	const secondEncoding = 'hex';
 
-	const firstResponse = await got({cache, encoding: firstEncoding});
+	const firstResponse = await got<string>({cache, encoding: firstEncoding});
```
sindresorhus: `got` by default returns a string, so I think it should automatically infer a `string` here. It should also have overloads to change the type depending on the `responseType` option.

test/create.ts
sindresorhus: Like I've commented elsewhere, don't use `JsonValue`. It creates more problems than it solves.

Generally, don't use `as` unless absolutely needed, as it can hide problem, since it just force casts anything to anything.

test/create.ts
```diff
@@ -228,21 +228,21 @@ test('extend with custom handlers', withServer, async (t, server, got) => {
...
+	t.is(b.defaults.options.prefixUrl!.toString(), 'https://example.com/');
 });
 
 test('extend with a chain', t => {
 	const a = got.extend({prefixUrl: 'https://example.com/'});
 	const b = got.extend(a, {headers: {foo: 'bar'}});
-	t.is(b.defaults.options.prefixUrl.toString(), 'https://example.com/');
-	t.is(b.defaults.options.headers.foo, 'bar');
+	t.is(b.defaults.options.prefixUrl!.toString(), 'https://example.com/');
+	t.is(b.defaults.options.headers!.foo, 'bar');
```
sindresorhus: `.headers` should always be defined. The default should just be an empty object.

test/hooks.ts
```diff
@@ -258,7 +258,7 @@ test('beforeRedirect is called with options and response', withServer, async (t,
 					t.is(options.url.hostname, 'localhost');
 
 					t.is(response.statusCode, 302);
-					t.is(new URL(response.url).pathname, '/redirect');
+					t.is(new URL(response.url!).pathname, '/redirect');
```
sindresorhus: `url` is not documented to possibly be `undefined`: https://github.com/sindresorhus/got#url-1

test/progress.ts
```diff
@@ -72,7 +72,7 @@ test('download progress', withServer, async (t, server, got) => {
 
 	const events: Progress[] = [];
 
-	const {body} = await got({encoding: null})
+	const {body} = await got<Buffer>({encoding: null})
```
sindresorhus: Note: https://github.com/sindresorhus/got/commit/6cc3d9f3027f28ce4dcbe8605d104e65f7e66787

source/request-as-event-emitter.ts
```diff
@@ -12,7 +12,7 @@ import getResponse from './get-response';
 import {normalizeRequestArguments} from './normalize-arguments';
 import {createProgressStream} from './progress';
 import timedOut, {TimeoutError as TimedOutTimeoutError} from './utils/timed-out';
-import {NormalizedOptions, Response, ResponseObject} from './utils/types';
+import {GenericError, NormalizedOptions, Response, ResponseObject} from './utils/types';
```
sindresorhus: I don't like the name `GenericError` as "generic" means something different with types. Maybe `BaseError` or `GeneralError`?

source/utils/types.ts
pmmmwh: Do we need to change form/json too?

sindresorhus: Yes. `{[key: string]: any}` has special behavior in TypeScript. https://github.com/sindresorhus/typescript-definition-style-guide/pull/8/files `Record` is fine as a input value where the value is not `any`/`unknown` or a return value. For this reason, I just prefer the `{[key: string]: any}` syntax as I can just consistently use it. Yet another thing I dislike about TypeScript. So much inconsistent and hidden behavior.

source/utils/types.ts
pmmmwh: This is referenced from the test for `content-length: 0`. Should I alter the test?

source/request-as-event-emitter.ts
pmmmwh: Renamed to `GeneralError`.

test/progress.ts
pmmmwh: Merged and fixed 🤞

test/hooks.ts
pmmmwh: Fixed.

test/create.ts
pmmmwh: Fixed!

test/cache.ts
pmmmwh: Updated in types and tests.

source/utils/types.ts
pmmmwh: Fixed accordingly 👍

source/utils/types.ts
sindresorhus: Yes

source/create.ts
```diff
@@ -24,36 +23,39 @@ export type HTTPAlias =
...
+export type ReturnStream = <T>(url: string | Merge<Options, {isStream?: true}>, options?: Merge<Options, {isStream?: true}>) => ProxyStream<T>;
+export type GotReturn<T = unknown> = CancelableRequest<T> | ProxyStream<T>;
 
 const getPromiseOrStream = (options: NormalizedOptions): GotReturn => options.isStream ? asStream(options) : asPromise(options);
 
-type OptionsOfDefaultResponseBody = Options & {isStream?: false; resolveBodyOnly?: false; responseType?: 'default'};
-type OptionsOfTextResponseBody = Options & {isStream?: false; resolveBodyOnly?: false; responseType: 'text'};
-type OptionsOfJSONResponseBody = Options & {isStream?: false; resolveBodyOnly?: false; responseType: 'json'};
-type OptionsOfBufferResponseBody = Options & {isStream?: false; resolveBodyOnly?: false; responseType: 'buffer'};
+const isGotInstance = (value: any): value is Got => (
```
sindresorhus:
```suggestion
const isGotInstance = (value: unknown): value is Got => (
```

source/normalize-arguments.ts
```diff
@@ -217,13 +224,13 @@ export const normalizeArguments = (url: URLOrOptions, options?: Options, default
 	if (is.urlInstance(url) || is.string(url)) {
 		options.url = url;
 
-		options = mergeOptions(defaults && defaults.options, options);
+		options = mergeOptions((defaults && defaults.options) ?? {}, options);
```
sindresorhus:
```suggestion
		options = mergeOptions(defaults?.options ?? {}, options);
```

source/normalize-arguments.ts
```diff
@@ -217,13 +224,13 @@ export const normalizeArguments = (url: URLOrOptions, options?: Options, default
...
 	} else {
 		if (Reflect.has(url, 'resolve')) {
 			throw new Error('The legacy `url.Url` is deprecated. Use `URL` instead.');
 		}
 
-		options = mergeOptions(defaults && defaults.options, url, options);
+		options = mergeOptions((defaults && defaults.options) ?? {}, url, options);
```
sindresorhus:
```suggestion
		options = mergeOptions(defaults?.options ?? {}, url, options);
```

test/agent.ts
```diff
@@ -29,7 +30,8 @@ const prepareServer = server => {
 	});
 };
 
-const createAgentSpy = Cls => {
+// TODO: Type this stricter if possible
+const createAgentSpy = <T extends any>(Cls: T) => {
```
sindresorhus:
```suggestion
const createAgentSpy = <T>(Cls: T) => {
```

And let's use a clearer name than `Cls`.

test/arguments.ts
```diff
@@ -104,13 +107,13 @@ test('overrides `searchParams` from options', withServer, async (t, server, got)
...
-				get(key) {
+				get(key: string) {
 					t.is(key, `cacheable-request:GET:${server.url}/?test=wow`);
 				},
-				set(key) {
+				set(key: string) {
 					t.is(key, `cacheable-request:GET:${server.url}/?test=wow`);
 				}
-			}
+			} as unknown as StorageAdapter
```
sindresorhus: Is this really needed? I think you just need to define the missing methods. The definition is:

```ts
interface Store<TValue> {
        get(key: string): TValue | Promise<TValue | undefined> | undefined;
        set(key: string, value: TValue, ttl?: number): any;
        delete(key: string): boolean | Promise<boolean>;
        clear(): void | Promise<void>;
    }
```

test/cookies.ts
```diff
@@ -14,7 +15,7 @@ test('reads a cookie', withServer, async (t, server, got) => {
 
 	const cookieJar = new toughCookie.CookieJar();
 
-	await got({cookieJar});
+	await got({cookieJar} as unknown as OptionsOfDefaultResponseBody);
```
sindresorhus: Why does this need force-casting?

test/create.ts
```diff
@@ -251,14 +252,16 @@ test('async handlers', withServer, async (t, server, got) => {
 		handlers: [
 			async (options, next) => {
 				const result = await next(options);
+				// @ts-ignore Manual tests
 				result.modified = true;
 
 				return result;
 			}
-		]
+		] as HandlerFunction[]
```
sindresorhus: Why does this need force-casting?

test/retry.ts
```diff
@@ -68,7 +70,7 @@ test('retry function gets iteration count', withServer, async (t, server, got) =
 		retry: {
 			calculateDelay: ({attemptCount}) => {
 				t.true(is.number(attemptCount));
-				return attemptCount < 2;
+				return Number(attemptCount < 2);
```
sindresorhus: Why is this needed? `attemptCount` is not documented to ever be a different value or `undefined`.

test/types/create-test-server/index.d.ts
```diff
@@ -0,0 +1,19 @@
+declare module 'create-test-server' {
+	import {Express} from 'express';
+
+	function createTestServer(options: any): Promise<createTestServer.TestServer>;
```
sindresorhus:
```suggestion
	function createTestServer(options: unknown): Promise<createTestServer.TestServer>;
```

test/types/create-test-server/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	export = createTestServer;
+
+	namespace createTestServer {
+		export interface TestServer extends Express {
+			caCert: any;
```
sindresorhus:
```suggestion
			caCert: unknown;
```

sindresorhus:
There are still a lot of force-casting (`as`) in the tests that should ideally not be there.

test/agent.ts
pmmmwh: Hmm. `unknown` doesn't seems to work here because it has to be something with a construct-able signature. Since `unknown` is not callable, it can never fulfill the type requirements?

test/cookies.ts
pmmmwh: I did this because I hit an issue where TypeScript is always using the first overload of the CookieJar's methods to match the use cases in our source code, which is actually using the third overload. Is there a way to make a proper comparison happen?

test/retry.ts
pmmmwh: From reading the source it seems that the return value of `calculateDelay` would be consumed by `computedValue`, which expects a number. When we return `attemptCount < 2`, it is returning a `boolean` type, and to cast it to a number would need `Number()`, or use `attemptCount < 2 ? 1 : 0`

test/create.ts
pmmmwh: This can be resolved, it was my oversight.

test/arguments.ts
pmmmwh: This is resolved by implementing all methods.

test/retry.ts
sindresorhus: `attemptCount < 2 ? 1 : 0` would be much clearer. Can you update https://github.com/sindresorhus/got#testing too?

test/create.ts
sindresorhus: Not clear to me what is complaining, but users should not need to force cast this.

test/cookies.ts
sindresorhus: Overloads are order-dependent, so maybe try changing the order? If this is in fact a TS issue, it would be good to try to find an issue on the TS issue tracker and add a code comment linking to it.

test/agent.ts
sindresorhus: Ah, but it should at least be something more specific than `any` then.

sindresorhus:
@pmmmwh I'm merging this now as tests are passing, to reduce potential merge conflicts for you and so that we can get this tested out in a beta version I plan to release now. Can you continue the remaining work in a new PR?

sindresorhus:
Just want to say that you've done an amazing job so far 🙌✨

pmmmwh:
> @pmmmwh I'm merging this now as tests are passing, to reduce potential merge conflicts for you and so that we can get this tested out in a beta version I plan to release now. Can you continue the remaining work in a new PR?

Sure! I will continue the work tomorrow probably - thanks for merging this!

Issue #447, opened 2018-01-15
Async Stacktrace lost on HTTPError


sparebytes:
The async stacktrace isn't very helpful when an `HTTPError` is thrown. On the other hand a timeout which throws a `RequestError` keeps the relevant stack info.

Here is example code that runs on Node 8:
```typescript
const got = require('got');

async function gotStackTraceLost() {
    try {
        const z = await got.post('https://google.com/bad-endpoint');
    }
    catch (e) {
        console.error("\n\nGOT Error and lost stacktrace!", e);
    }
}

async function gotStackTraceKept() {
    try {
        const z = await got.post('https://127.0.0.1:9999/bad-endpoint', { timeout: 1 });
    }
    catch (e) {
        console.error("\n\nGOT Error and kept stacktrace!", e);
    }
}

async function controlTest() {
    try {
        throw new Error("This is a test");
    }
    catch (e) {
        console.error("\n\nControl Error!", e);
    }
}

gotStackTraceLost();
gotStackTraceKept();
controlTest();
```

brandon93s:
Got now maintains the stacktrace as expected for the `HTTPError`:

![image](https://user-images.githubusercontent.com/4730164/42413871-751363d8-81f7-11e8-83bf-1d9f12f2a5af.png)


Note that you can also set `throwHttpErrors` to false if you'd prefer to handle these yourself as a response.

*Tested against master.*

szmarczak:
@brandon93s No and yes. No, because the stacktrace should show more details. It's because `setImmediate` is used. But that's how Node.JS deals with that, so I don't think we can help with that.

```js
    const EventEmitter = require('events');

    const givesPromise = () => new Promise((resolve, reject) => {
        const ee = new EventEmitter();

        ee.once('error', e => {
            reject(e);
        });
        // return ee.emit('error', new Error('stacktrace not lost')); return; // uncomment this line and you'll see it gives more info
        setImmediate(async _ => {
            ee.emit('error', new Error('stacktrace lost'));
        })
    });

    (async function() {
        try {
            await givesPromise();
        } catch (e) {
            console.log(e);
        }
    })();
```

sindresorhus:
Relevant Node.js issue: https://github.com/nodejs/node/issues/11865

PR #935, opened 2019-11-17
Preserve stacktrace when wrapping errors

### LINKED ISSUE #795: Wrapping `RequestError` obscures its source

When I’m working on and with nock, I often see stack traces that look like this:

```
OverriddenClientRequest.request.once.error (node_modules/got/source/request-as-event-emitter.js:178:14)
OverriddenClientRequest.origin.emit.args (node_modules/@szmarczak/http-timer/source/index.js:37:11)
lib/request_overrider.js:24:552
```

Or sometimes:

```
OverriddenClientRequest.request.once.error (/Users/pnm/code/nock/node_modules/got/source/request-as-event-emitter.js:178:14)
OverriddenClientRequest.emit (events.js:187:15)
OverriddenClientRequest.EventEmitter.emit (domain.js:442:20)
OverriddenClientRequest.origin.emit.args (/Users/pnm/code/nock/node_modules/@szmarczak/http-timer/source/index.js:37:11)
/Users/pnm/code/nock/lib/request_overrider.js:24:552
process._tickCallback (internal/process/next_tick.js:61:11)
```

The underlying stack trace is this:

```
end (/Users/pnm/code/nock/lib/request_overrider.js:32:187)
/Users/pnm/code/nock/lib/request_overrider.js:13:2198
OverriddenClientRequest.RequestOverrider.req.write (/Users/pnm/code/nock/lib/request_overrider.js:13:1387)
OverriddenClientRequest.RequestOverrider.req.end (/Users/pnm/code/nock/lib/request_overrider.js:13:1970)
handleRequest (/Users/pnm/code/nock/node_modules/got/source/request-as-event-emitter.js:212:14)
get (/Users/pnm/code/nock/node_modules/got/source/request-as-event-emitter.js:236:5)
Immediate.setImmediate (/Users/pnm/code/nock/node_modules/got/source/request-as-event-emitter.js:307:10)
```

However when got wraps the error in a RequestError, the original stack trace is no longer available.

To be fair, part of what makes this confusing is that the test runner’s output (which has nothing to do with nock or got) and looks like this:

```
 FAIL  tests/test_intercept.js
Nock: No match for request {   "method": "GET",   "url": "http://example.com/",   "headers": {     "user-agent": "got/9.6.0 (https://github.com/sindresorhus/got)",     "accept-encoding": "gzip, deflate"   } }

				} else {
					error = new RequestError(error, options);
-------------^
				}

  stack: >
    OverriddenClientRequest.request.once.error
    (node_modules/got/source/request-as-event-emitter.js:178:14)

    OverriddenClientRequest.origin.emit.args (node_modules/@szmarczak/http-timer/source/index.js:37:11)

    lib/request_overrider.js:24:552
  at:
    line: 178
    column: 14
    file: node_modules/got/source/request-as-event-emitter.js
    function: OverriddenClientRequest.request.once.error
```

For a long time it was unclear that the error and had anything to do with Nock. Nock doesn’t appear in the stack trace despite being the layer that threw the error. And the test runner's output doesn't make it quite so clear to me that “Nock: No match for request” was the message for the error that was being emitted by got.

There is [not much that is documented about RequestError](https://github.com/sindresorhus/got/blob/dfb46ad0bf2427f387968f67ac943476597f0a3b/readme.md#gotrequesterror), though I imagine the benefit of wrapping errors in RequestError is to associate with the thrown error the additional bits information about the request. In this case, however, the wrapping did not provide a benefit. Instead, it made it harder to understand what was happening. 

Some possible ideas for making this easier for developers to understand:

1. Preserve the original `error.stack` and attach it to the new RequestError. This could be done in GotError’s constructor. I experimented with this locally and it seems to work. (Though it is slightly disorienting to see Nock’s stack trace attached to something called RequestError, which is not from nock.)
2. Replace the wrapping with decorating. Instead of creating a new error, attach the got-related `options` to the existing error object.
3. Modify nock so it doesn’t emit this error through the request.

I’d be really curious to hear  your thoughts on (1) and (2)! Regardless of what's decided here, I think (3) could be worth pursuing, as it will make it easier both to test nock and also to use it.

I can’t find a discussion on this specific topic, though there was some discussion about async stack traces in #279 and #447.

Thanks so much for all your work on this amazing library!

#### Checklist

szmarczak:
Fixes #795

#### Checklist

- [ ] ~If it's a new feature, I have included documentation updates.~

### DIFF
--- a/source/errors.ts
+++ b/source/errors.ts
@@ -7,7 +7,7 @@ export class GotError extends Error {
 	code?: string;
 	readonly options: NormalizedOptions;
 
-	constructor(message: string, error: (Error & {code?: string}) | {code?: string}, options: NormalizedOptions) {
+	constructor(message: string, error: Partial<Error & {code?: string}>, options: NormalizedOptions) {
 		super(message);
 		Error.captureStackTrace(this, this.constructor);
 		this.name = 'GotError';
@@ -19,6 +19,20 @@ export class GotError extends Error {
 		Object.defineProperty(this, 'options', {
 			value: options
 		});
+
+		// Recover the original stacktrace
+		if (!is.undefined(error.stack)) {
+			const indexOfMessage = this.stack.indexOf(this.message) + this.message.length;
+			const thisStackTrace = this.stack.slice(indexOfMessage).split('\n').reverse();
+			const errorStackTrace = error.stack.slice(error.stack.indexOf(error.message) + error.message.length).split('\n').reverse();
+
+			// Remove duplicated traces
+			while (errorStackTrace.length !== 0 && errorStackTrace[0] === thisStackTrace[0]) {
+				thisStackTrace.shift();
+			}
+
+			this.stack = `${this.stack.slice(0, indexOfMessage)}${thisStackTrace.reverse().join('\n')}${errorStackTrace.reverse().join('\n')}`;
+		}
 	}
 }
 
@@ -96,7 +110,7 @@ export class TimeoutError extends GotError {
 	event: string;
 
 	constructor(error: TimedOutError, timings: Timings, options: NormalizedOptions) {
-		super(error.message, {code: 'ETIMEDOUT'}, options);
+		super(error.message, error, options);
 		this.name = 'TimeoutError';
 		this.event = error.event;
 		this.timings = timings;
--- a/test/error.ts
+++ b/test/error.ts
@@ -197,3 +197,17 @@ test('errors are thrown directly when options.stream is true', t => {
 		message: 'Parameter `hooks` must be an Object, not boolean'
 	});
 });
+
+test('the old stacktrace is recovered', async t => {
+	const error = await t.throwsAsync(got('https://example.com', {
+		request: () => {
+			throw new Error('foobar');
+		}
+	}));
+
+	t.true(error.stack.includes('at Object.request'));
+
+	// The first `at get` points to where the error was wrapped,
+	// the second `at get` points to the real cause.
+	t.not(error.stack.indexOf('at get'), error.stack.lastIndexOf('at get'));
+});
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -283,7 +283,7 @@ export default (options: NormalizedOptions) => {
 		}
 	};
 
-	setImmediate(async () => {
+	(async () => {
 		try {
 			for (const hook of options.hooks.beforeRequest) {
 				// eslint-disable-next-line no-await-in-loop
@@ -294,7 +294,7 @@ export default (options: NormalizedOptions) => {
 		} catch (error) {
 			emitError(error);
 		}
-	});
+	})();
 
 	return emitter;
 };
--- a/source/request-as-event-emitter.ts
+++ b/source/request-as-event-emitter.ts
@@ -14,6 +14,7 @@ import {CacheError, MaxRedirectsError, RequestError, TimeoutError} from './error
 import urlToOptions from './utils/url-to-options';
 import {NormalizedOptions, Response, ResponseObject} from './utils/types';
 
+const setImmediateAsync = () => new Promise(resolve => setImmediate(resolve));
 const pipeline = promisify(stream.pipeline);
 
 const redirectCodes: ReadonlySet<number> = new Set([300, 301, 302, 303, 304, 307, 308]);
@@ -284,6 +285,8 @@ export default (options: NormalizedOptions) => {
 	};
 
 	(async () => {
+		await setImmediateAsync();
+
 		try {
 			for (const hook of options.hooks.beforeRequest) {
 				// eslint-disable-next-line no-await-in-loop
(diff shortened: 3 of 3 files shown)


test/error.ts
```diff
@@ -199,3 +199,17 @@ test('errors are thrown directly when options.stream is true', t => {
...
+	const error = await t.throwsAsync(got('https://example.com', {
+		request: () => {
+			throw new Error('foobar');
+		}
+	}));
+
+	t.true(error.stack.includes('at Object.request'));
+
+	// The first `at get` points to where the error was wrapped,
+	// the second `at get` points to the real cause.
```
szmarczak: Now:

```
GotError: foobar
    at get (/home/szm/Desktop/got/source/request-as-event-emitter.ts:250:15)
    at Object.request (/home/szm/Desktop/got/test/error.ts:206:10)
    at get (/home/szm/Desktop/got/source/request-as-event-emitter.ts:248:31)
    at Immediate.<anonymous> (/home/szm/Desktop/got/source/request-as-event-emitter.ts:313:4)
```

Before:

```
GotError: foobar
    at get (/home/szm/Desktop/got/source/request-as-event-emitter.ts:250:15)
    -- the line above only points to where the error was captured  --
	-- missing what caused the error --
    at Immediate.<anonymous> (/home/szm/Desktop/got/source/request-as-event-emitter.ts:313:4)
```

szmarczak:
As you can see, the stacktrace begins with `at Immediate.<anonymous>`. It's [this](https://github.com/sindresorhus/got/blob/518f0f5906e1fcd38691555b17fb91db643f3d47/source/request-as-event-emitter.ts#L306). But what called it, we don't know. See https://github.com/sindresorhus/got/issues/447#issuecomment-403237446

To fix this, we need to either remove `setImmediate` or capture another stacktrace there. Or... leave this as is, because it's default Node behavior.

source/errors.ts
```diff
@@ -19,6 +19,20 @@ export class GotError extends Error {
...
 		});
+
+		// Recover the original stacktrace
+		if (!is.undefined(error.stack)) {
+			const indexOfMessage = this.stack.indexOf(this.message) + this.message.length;
+			const thisStackTrace = this.stack.slice(indexOfMessage).split('\n').reverse();
+			const errorStackTrace = error.stack.slice(error.stack.indexOf(error.message) + error.message.length).split('\n').reverse();
+
+			// Remove duplicated traces
+			while (errorStackTrace.length !== 0 && errorStackTrace[0] === thisStackTrace[0]) {
```
szmarczak: This assumes these stacktraces have **only** common beginning.

szmarczak: Should these stack traces be separated by a line break?

sindresorhus:
> To fix this, we need to either remove setImmediate (no big impact on performance)

Why do we even have it there? From what I faintly can remember, it had something to do with a race issue. Maybe we could fix it another way and remove the `setImmediate`?

> capture another stacktrace there (this would have a big impact on performance)

I doubt it has big impact on performance relatively to networking.

---

Did you check if the `setImmediate` problem is the same on Node.js 13 too? Later Node.js versions got much better support for preserving async stack traces.

source/errors.ts
sindresorhus: > Should these stack traces be separated by a line break to indicate that they are from two different sources?

No, that would potentially break stack trace parsers.

szmarczak:
> Did you check if the setImmediate problem is the same on Node.js 13 too? Later Node.js versions got much better support for preserving async stack traces.

Yep, it's the same.

> I doubt it has big impact on performance relatively to networking.

```js
(() => {
    let timings = [];
    let start;
    for (let i = 0; i < 100000; i++) {
        start = performance.now();
        (new Error()).stack;
        timings.push(performance.now() - start);
    }

	return timings.reduce((a, b) => a + b) / timings.length;
})();
```

gives me `0.0075` ms / stacktrace. So you're right. I thought it would be time-consuming.

>  From what I faintly can remember, it had something to do with a race issue. Maybe we could fix it another way and remove the setImmediate?

It was introduced here: https://github.com/sindresorhus/got/pull/304

We can remove it because the errors are handled in a much different way than at the time.

szmarczak:
I'll fix this in a few hours, need sleep now

szmarczak:
We need `setImmediate` to allow calling `promise.json()` etc. Promises are executed immediately, so if there were no `setImmediate`, `promise.json()` would have no effect on the headers.

magnusottosson:
@szmarczak Is this supposed to solve the async stack trace problems? On 10.6.0 I get a stack trace that look like this:

`
HTTPError: Response code 400 (Bad Request)
    at EventEmitter.<anonymous> ([redacted]/node_modules/got/dist/source/as-promise.js:118:31)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
`

I have no idea where this error occurs :)

My request looks something like this:

`
const response = await got('someurl', {
        json: {...},
        headers: {
            'Cache-Control': 'no-cache, no-store',
        },
        method: 'POST',
    });
`

szmarczak:
@magnusottosson The title of the PR says "when wrapping errors", not "when handling async errors". You're confusing this PR with #1077

magnusottosson:
Ah sorry. My bad :)

PR #940, opened 2019-11-23
Use `responseType` instead of `options.encoding` to get a Buffer

szmarczak:
https://github.com/sindresorhus/got/pull/921#discussion_r347079358

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -248,10 +248,10 @@ Default: `'default'`
 
 Parsing method used to retrieve the body from the response.
 
-- `'default'` - if `options.encoding` is `null`, the body will be a Buffer. Otherwise it will be a string unless it's overwritten in a `afterResponse` hook,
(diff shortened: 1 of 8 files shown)


sindresorhus:
Can you add a test to ensure `got` and `responseType: 'buffer'` works even when the `encoding` option is set?

sindresorhus:
You need to remove the `null` and Buffer mention here: https://github.com/sindresorhus/got#encoding And also say that it only applies when the response type is some kind of string.

sindresorhus:
You also need to update:

```
test/progress.ts
74:	const {body} = await got({encoding: null})
95:	const stream = got.stream({encoding: null})
```

sindresorhus:
I think we should throw if `got({encoding: null})` saying that the user should use `got({responseType: 'buffer'})` instead.

PR #942, opened 2019-11-24
Add `methodRewriting` option

### LINKED ISSUE #934: Disable method overwriting
#### Describe the bug

 - Node.js version: v13.1.0
 - OS & version: ArchLinux
 - Got library version: commit https://github.com/sindresorhus/got/commit/518f0f5906e1fcd38691555b17fb91db643f3d47

When doing a POST request with `followAllRedirects: true` as an option to an endpoint that redirect with a status code of 302 the library `got` send a POST request to the new endpoint instead of doing a simple GET request like the `request` library would do it.
This bug report is related to this issue: https://github.com/sindresorhus/got/issues/568

#### Actual behavior
Send a POST request on the second request after following the redirect.
```
HTTPError: Response code 405 (Method Not Allowed)
    at EventEmitter.<anonymous> (/tmp/node_modules/got/dist/source/as-promise.js:110:31)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  name: 'HTTPError'
}
Wrong method on redirect.
```

#### Expected behavior

Send a simple GET request on the second request after following the redirect.
```
200
Good method on redirect.
```

#### Code to reproduce

Code with the `got` library:
```js
const got = require('got');

(async () => {
    try {
        const response = await got("http://localhost:3000", {
            method: "POST",
            form: {
                username: "username",
                password: "password"
            },
            followRedirect: true
        });
        console.log(response.statusCode);
        console.log(response.body);
    } catch (error) {
        console.log(error);
        console.log(error.response.body);
    }
})();
```

Code with the `request` library:
```js
const request = require('request');

const options = {
    method: 'POST',
    uri: 'http://localhost:3000',
    form: {
        username: "username",
        password: "password"
    },
    followAllRedirects: true
};

request(options, function(err, res, body) {
    console.log(res.statusCode);
    console.log(body);
});
```

Express code to test the behavior:
```js
const express = require('express');
const app = express();

app.get('/redirect', (req, res) => {
    res.send('Good method on redirect.');
    console.log('Good method on redirect.');
});

app.post('/redirect', (req, res) => {
    res.status(405).send('Wrong method on redirect.');
    console.log('Wrong method on redirect.');
});

app.post('/', (req, res) => {
    res.redirect(302, "/redirect");
    console.log('Received POST request, doing a redirect...');
});

app.listen(3000);
```

#### Checklist

szmarczak:
Fixes #934

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -408,7 +408,14 @@ Defines if redirect responses should be followed automatically.
 
 Note that if a `303` is sent by the server in response to any request type (`POST`, `DELETE`, etc.), Got will automatically request the resource pointed to in the location header via `GET`. This is in accordance with [the spec](https://tools.ietf.org/html/rfc7231#section-6.4.4).
 
-This supports [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend that request to the new location.
+###### enableMethodRewriting
+
+Type: `boolean`<br>
+Default: `true`
+
+By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location.
+
+To disable method rewriting, set this to `false`.
 
 ###### maxRedirects
 
--- a/readme.md
+++ b/readme.md
@@ -413,7 +413,7 @@ Note that if a `303` is sent by the server in response to any request type (`POS
 Type: `boolean`<br>
 Default: `true`
 
-By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location.
+By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location using the same HTTP method (`POST` in this case).
 
 To disable method rewriting, set this to `false`.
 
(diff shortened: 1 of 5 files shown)


sindresorhus:
I think it can just be named `methodRewriting` (enable is implied).

.dockerignore
```diff
@@ -0,0 +1,6 @@
+node_modules
+yarn.lock
+coverage
+.nyc_output
+dist
+.*.dockerfile
```
sindresorhus: ?

readme.md
```diff
@@ -408,7 +408,14 @@ Defines if redirect responses should be followed automatically.
 
 Note that if a `303` is sent by the server in response to any request type (`POST`, `DELETE`, etc.), Got will automatically request the resource pointed to in the location header via `GET`. This is in accordance with [the spec](https://tools.ietf.org/html/rfc7231#section-6.4.4).
 
-This supports [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend that request to the new location.
+###### methodRewriting
+
+Type: `boolean`<br>
+Default: `true`
+
+By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location.
```
```suggestion
By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location using the same HTTP method (`POST` in this case).
```

?

readme.md
```diff
@@ -408,7 +408,14 @@ Defines if redirect responses should be followed automatically.
...
+
+To disable method rewriting, set this to `false`.
```
sindresorhus: I feel like this is moot. We already explain what it does and that it's on by default. This just repeats what's already shown.

.dockerignore
szmarczak: lemme revert that :joy:

PR #946, opened 2019-11-26
Improve types

pmmmwh:
This is a follow-up to #928 . Hopefully this is the last batch of (type breaking?) changes 🤞 .

#### Changes
- [x] Remove force casting when appropriate (some are left cause they are intended to be errors)
- [x] Better exported types so that users wouldn't have to force cast
- [x] Lint project to fully comply with XO rules
- [x] All TODO comments related to types should have been resolved by #928 or this PR
- [x] All fixable `ts-ignore` comments should have been removed by #928 or this PR
- [x] All non-fixable `ts-ignore` comments should have a description stating what happened there
- [ ] Refactor/Write/Copy-n-Paste Doc-String for exported types

#### Checklist

### DIFF
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -7,7 +7,7 @@ import {normalizeArguments, mergeOptions} from './normalize-arguments';
 import requestAsEventEmitter, {proxyEvents} from './request-as-event-emitter';
 import {CancelableRequest, GeneralError, NormalizedOptions, Response} from './utils/types';
 
-const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']): unknown | never => {
+const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']): unknown => {
 	if (responseType === 'json' && is.string(body)) {
 		return statusCode === 204 ? '' : JSON.parse(body);
 	}
@@ -79,7 +79,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 							resolveBodyOnly: false
 						}));
 
-						// Remove any further hooks for that request, because we'll call them anyway.
+						// Remove any furthe1r hooks for that request, because we'll call them anyway.
 						// The loop continues. We don't want duplicates (asPromise recursion).
 						updatedOptions.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
 
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -7,7 +7,7 @@ import {normalizeArguments, mergeOptions} from './normalize-arguments';
 import requestAsEventEmitter, {proxyEvents} from './request-as-event-emitter';
 import {CancelableRequest, GeneralError, NormalizedOptions, Response} from './utils/types';
 
-const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']) => {
+const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']): unknown | never => {
 	if (responseType === 'json' && is.string(body)) {
 		return statusCode === 204 ? '' : JSON.parse(body);
 	}
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -79,7 +79,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 							resolveBodyOnly: false
 						}));
 
-						// Remove any furthe1r hooks for that request, because we'll call them anyway.
+						// Remove any further hooks for that request, because we'll call them anyway.
 						// The loop continues. We don't want duplicates (asPromise recursion).
 						updatedOptions.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
 
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -24,7 +24,7 @@ const parseBody = (body: Response['body'], responseType: NormalizedOptions['resp
 		return body;
 	}
 
-	throw new Error(`Failed to parse body of type '${typeof body}' as '${responseType}'`);
+	throw new Error(`Failed to parse body of type '${typeof body}' as '${responseType!}'`);
 };
 
 export default function asPromise<T>(options: NormalizedOptions): CancelableRequest<T> {
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -141,7 +141,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 		emitter.once('error', reject);
 
 		proxyEvents(proxy, emitter);
-	}) as CancelableRequest<any>;
+	}) as CancelableRequest<T>;
 
 	promise.on = (name: string, fn: (...args: any[]) => void) => {
 		proxy.on(name, fn);
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -66,7 +66,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 
 			try {
 				for (const [index, hook] of options.hooks.afterResponse.entries()) {
-					// @ts-ignore
+					// @ts-ignore Promise is not assignable to CancelableRequest
 					// eslint-disable-next-line no-await-in-loop
 					response = await hook(response, async (updatedOptions: NormalizedOptions) => {
 						updatedOptions = normalizeArguments(mergeOptions(options, {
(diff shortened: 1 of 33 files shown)


sindresorhus:
Did you see https://github.com/sindresorhus/got/issues/876#issuecomment-558813981 ?

pmmmwh:
> Did you see [#876 (comment)](https://github.com/sindresorhus/got/issues/876#issuecomment-558813981) ?
> 
> [DefinitelyTyped/DefinitelyTyped#34960](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960)

I actually just read that. Should we just augment the global scope and add the two classes there? I don't see another way out ...

sindresorhus:
> I actually just read that. Should we just augment the global scope and add the two classes there? I don't see another way out ...

Yeah, we can do that for now.

pmmmwh:
Pushed a commit to finish up the work on `ts-ignore` comments. Afaik most of the work related to types (other than doc strings) should be done?

source/types/url/index.d.ts
```diff
@@ -0,0 +1,37 @@
+// TODO: Remove this when DefinitelyTyped/#34960 is resolved
```
sindresorhus: Use the full URL.

sindresorhus:
One more that could use a better type: https://github.com/sindresorhus/got/blob/d968e4904b92c8b24816cc9f35262214b0e7f103/source/as-stream.ts#L17

And maybe also: https://github.com/sindresorhus/got/blob/d968e4904b92c8b24816cc9f35262214b0e7f103/test/helpers/with-server.ts#L30

sindresorhus:
Regarding the doc strings. I think that should be a separate PR after this one. I'd like to get this one merged asap so we can do the final Got v10 release.

source/as-promise.ts
```diff
@@ -7,7 +7,7 @@ import {normalizeArguments, mergeOptions} from './normalize-arguments';
 import requestAsEventEmitter, {proxyEvents} from './request-as-event-emitter';
 import {CancelableRequest, GeneralError, NormalizedOptions, Response} from './utils/types';
 
-const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']) => {
+const parseBody = (body: Response['body'], responseType: NormalizedOptions['responseType'], statusCode: Response['statusCode']): unknown | never => {
```
szmarczak: `unknown | never` is useless here I think.

source/normalize-arguments.ts
```diff
@@ -79,7 +80,7 @@ export const preNormalizeArguments = (options: Options, defaults?: NormalizedOpt
 
 	if (defaults) {
 		for (const event of knownHookEvents) {
-			// @ts-ignore TS is dumb.
+			// @ts-ignore Union type array is not assignable to union array type
```
szmarczak: Thanks for explaining this, I am tired of TS' dumbness.

source/utils/types.ts
```diff
@@ -86,7 +86,7 @@ export interface RetryObject {
 
 export type RetryFunction = (retryObject: RetryObject) => number;
 
-export type HandlerFunction = <T extends GotReturn>(options: NormalizedOptions, next: (options: NormalizedOptions) => T) => T;
+export type HandlerFunction = <T extends GotReturn>(options: NormalizedOptions, next: (options: NormalizedOptions) => T) => T | Promise<T>;
```
szmarczak: Instead of `T | Promise<T>` you can use `type-fest`: `Promisable<T>`

source/utils/types.ts
```diff
@@ -101,7 +101,7 @@ export interface RetryOptions extends Partial<DefaultRetryOptions> {
 	retries?: number;
 }
 
-export type RequestFunction = typeof https.request;
+export type RequestFunction = typeof http.request | typeof https.request;
```
szmarczak: I think it's useless to provide both. They have the same API.

source/utils/types.ts
```diff
@@ -120,10 +120,15 @@ export interface Delays {
 
 export type Headers = Record<string, string | string[] | undefined>;
 
-interface CookieJar {
+interface ToughCookieJar {
+	getCookieString(currentUrl: string, options: {[key: string]: unknown}, cb: (err: Error | null, cookies: string) => void): void;
 	getCookieString(url: string, callback: (error: Error | null, cookieHeader: string) => void): void;
-	getCookieString(url: string): Promise<string>;
+	setCookie(cookieOrString: unknown, currentUrl: string, options: {[key: string]: unknown}, cb: (err: Error | null, cookie: unknown) => void): void;
```
szmarczak: Why the change?

source/utils/types.ts
```diff
@@ -159,7 +164,7 @@ export interface Options extends Partial<Except<DefaultOptions, 'retry'>>, Merge
 	method?: Method;
 	retry?: RetryOptions | number;
 	throwHttpErrors?: boolean;
-	cookieJar?: CookieJar;
+	cookieJar?: ToughCookieJar | PromiseCookieJar;
```
szmarczak: :+1:

szmarczak:
Also I've fixed some types here: https://github.com/sindresorhus/got/pull/833/files#diff-05c76183bbee7e002aa6711e6baca2ef so you may want to take a look and fix it here also.

szmarczak:
LGTM but it'd be nice if you merged my fixes into this PR, I don't expect my PR to get merged before yours.

pmmmwh:
> One more that could use a better type:
> 
> https://github.com/sindresorhus/got/blob/d968e4904b92c8b24816cc9f35262214b0e7f103/source/as-stream.ts#L17
> 
> And maybe also:
> 
> https://github.com/sindresorhus/got/blob/d968e4904b92c8b24816cc9f35262214b0e7f103/test/helpers/with-server.ts#L30

The first one - I think the interface is actually 100% equal to the ServerResponse class from the 'http' library? I checked `stream.Writable` and it didn't seem to have a good fit of properties.

The second one - I haven't found a way to work around that. It seems to be a TypeScript problem as stated [here](https://github.com/Microsoft/TypeScript/issues/7294)? I can try removing the generics for `CancelableRequest` and `ProxyStream` (as suggested as a cause by someone in the thread) and see if it fixes that.

source/as-promise.ts
```diff
@@ -79,7 +79,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 							resolveBodyOnly: false
 						}));
 
-						// Remove any further hooks for that request, because we'll call them anyway.
+						// Remove any furthe1r hooks for that request, because we'll call them anyway.
```
szmarczak:
```suggestion
						// Remove any further hooks for that request, because we'll call them anyway.
```

source/as-stream.ts
```diff
@@ -115,13 +115,14 @@ export default function asStream<T>(options: NormalizedOptions): ProxyStream<T>
 		pipe(destination, options);
 
 		if (Reflect.has(destination, 'setHeader')) {
```
szmarczak: This should be: `destination instanceof ServerResponse`

source/as-stream.ts
```diff
@@ -115,13 +115,14 @@ export default function asStream<T>(options: NormalizedOptions): ProxyStream<T>
...
+			// @ts-ignore This IS correct, we check for a property but the type-guard broke type infer.
 			piped.add(destination);
 		}
 
 		return destination;
 	};
 
-	proxy.unpipe = stream => {
+	proxy.unpipe = (stream: ServerResponse) => {
```
szmarczak: You do it wrong.

source/as-stream.ts
```diff
@@ -115,13 +115,14 @@ export default function asStream<T>(options: NormalizedOptions): ProxyStream<T>
...
 		piped.delete(stream);
```
szmarczak: `piped.delete(stream as ServerResponse)` is the correct way

pmmmwh:
> LGTM but it'd be nice if you merged my fixes into this PR, I don't expect my PR to get merged before yours.

Do you want me to pull in all changes related to types from there? I am particularly not very confident on pulling in changes in [this file](https://github.com/sindresorhus/got/pull/833/files#diff-02301bc46e8b878f10e9a8339efb7de7), since there is some change in normalization logic too?

szmarczak:
I'll send a PR to your PR (`types-follow-up` branch) ;)

szmarczak:
@pmmmwh Sent you a PR ;)

szmarczak:
If you've got any questions, feel free to ask. This is great work! :D

pmmmwh:
> If you've got any questions, feel free to ask. This is great work! :D

Checked and merged 😺

sindresorhus:
Awesome! Thanks for all your work on this, @pmmmwh! We could not have done this without you. 🙌

PR #953, opened 2019-12-01
Fix `got.mergeOptions(...)` types

scttcper:
Adds test coverage for mergeOptions usage similar to that in the readme.

Not passing because I'm not sure what the fix should be, willing to make the change. Partial?

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -70,7 +70,7 @@ export interface Got extends Record<HTTPAlias, GotFunctions>, GotFunctions {
 
 	extend(...instancesOrOptions: Array<Got | ExtendOptions>): Got;
 	mergeInstances(parent: Got, ...instances: Got[]): Got;
-	mergeOptions<T extends PartialDeep<Options>>(...sources: T[]): T;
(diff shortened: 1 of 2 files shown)


szmarczak:
The fix is to replace

https://github.com/sindresorhus/got/blob/d11dbe40a4e8da094afc000973973fd19ef6ae2a/source/create.ts#L73

with

```ts
	mergeOptions(...sources: Options[]): NormalizedOptions;
```

I forgot to update it in #921

I'm not sure how to test the types though. Should we use [`tsd`](https://github.com/SamVerschueren/tsd)?

/cc @sindresorhus

sindresorhus:
> I'm not sure how to test the types though. Should we use tsd?

`tsd` is meant to test TypeScript definition files for JS projects.

If we really need to test the actual types, we could use something like this:

```ts
type AssertEqual<T, Expected> =
  T extends Expected
  ? (Expected extends T ? true : never)
  : never;
```

But I don't think it's needed in this case.

PR #969, opened 2019-12-06
Fix type problem with `URL` and `URLSearchParams` globals

### LINKED ISSUE #960: Cannot find name 'URLSearchParams'

I've just upgraded to `got@10` and got this TypeScript error:

```
tsc
node_modules/got/dist/source/utils/options-to-url.d.ts:12:71 - error TS2304: Cannot find name 'URLSearchParams'.
```

I figured it's because in my `tsconfig.json` `libs` I don't have `dom`. I'm consciously not using `dom` there to avoid importing/using Browser types that I never need in Node.js project (it caused me issues in the past, e.g global dom `Request` type clashes with Express.js `Request` type).

Is there a way to use `got@10` in a TypeScript project without `dom` in `libs`?

#### Checklist

pmmmwh:
This fixes #960 , basically duplicating the URL global shim to the types file which is exported to user-land.

### DIFF
--- a/source/utils/types.ts
+++ b/source/utils/types.ts
@@ -247,3 +247,44 @@ export interface CancelableRequest<T extends Response | Response['body']> extend
 	buffer(): CancelableRequest<Buffer>;
 	text(): CancelableRequest<string>;
 }
+
+// TODO: Remove this when https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960 is resolved
+declare global {
+	// @ts-ignore TypeScript complains about duplicated identifier.
+	class URL {
+		readonly origin: string;
+		readonly searchParams: URLSearchParams;
+		hash: string;
+		host: string;
+		hostname: string;
+		href: string;
+		password: string;
+		pathname: string;
+		port: string;
+		protocol: string;
+		search: string;
+		username: string;
+
+		constructor(input: string, base?: string | URL);
+		toString(): string;
+		toJSON(): string;
+	}
+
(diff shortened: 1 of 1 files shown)


szmarczak:
Shouldn't be https://github.com/sindresorhus/got/blob/master/source/types/url/index.d.ts removed then?

Also should https://github.com/sindresorhus/got/blob/master/source/types/reflect/index.d.ts be moved also? And https://github.com/sindresorhus/got/blob/master/source/types/mimic-response/index.d.ts ?

pmmmwh:
> Shouldn't be https://github.com/sindresorhus/got/blob/master/source/types/url/index.d.ts removed then?
> 
> Also should https://github.com/sindresorhus/got/blob/master/source/types/reflect/index.d.ts be moved also? And https://github.com/sindresorhus/got/blob/master/source/types/mimic-response/index.d.ts ?

1. I kept it because I think the two files signals two different problems, one about URL missing in the global types, and the other about TSC not able to output non module files. If TSC got updated, we could safely remove the change done in this PR, and if URLs got added, we can remove both. If you think this is unnecessary, I can PR to remove the duplication.
2. They don't have to be added because they are only used internally by the source. Won't bleed type errors into user land.

szmarczak:
Honestly I don't know what's the correct solution. Just asking.

pmmmwh:
Hi

> Honestly I don't know what's the correct solution. Just asking.

Me neither, they are all workarounds for unfortunate limitations to TS, so I just wanted to make it work and move on ...

PR #991, opened 2019-12-13
Retry after certain Cloudflare errors

dguo:
#### Checklist

Many of the [Cloudflare error codes](https://support.cloudflare.com/hc/en-us/articles/115003011431) seem appropriate for retry.

I omitted the SSL ones with the reasoning that those are unlikely to be intermittent. I also omitted 530 because it can refer to many errors, like hitting a rate limit.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -367,7 +367,7 @@ Default:
 - limit: `2`
 - calculateDelay: `(attemptCount, retryOptions, error, computedValue) => computedValue`
 - methods: `GET` `PUT` `HEAD` `DELETE` `OPTIONS` `TRACE`
-- statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`500`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) [`522`](https://support.cloudflare.com/hc/en-us/articles/115003011431#522error) [`524`](https://support.cloudflare.com/hc/en-us/articles/115003011431#524error)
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -367,7 +367,7 @@ Default:
...
+- statusCodes: [`408`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) [`413`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) [`429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) [`500`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) [`502`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`504`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) [`520`](https://support.cloudflare.com/hc/en-us/articles/115003011431#520error) [`521`](https://support.cloudflare.com/hc/en-us/articles/115003011431#521error) [`522`](https://support.cloudflare.com/hc/en-us/articles/115003011431#522error) [`523`](https://support.cloudflare.com/hc/en-us/articles/115003011431#523error) [`524`](https://support.cloudflare.com/hc/en-us/articles/115003011431#524error) [`527`](https://support.cloudflare.com/hc/en-us/articles/115003011431#527error)
```
szmarczak: Only `522` and `524` should be retry-able.

szmarczak: Other `5XX` codes are not.

szmarczak:
Don't forget to include some tests :)

readme.md
dguo: Just so I understand why, is your reasoning that the other ones are unlikely to be intermittent?

dguo:
I didn't add a test because I figured a test would sort of mirror the code (e.g. effectively asserting that the array of status codes is what it is in the code). 

Would you like me to add a test anyway? What do you think would be a useful one? I could add one that loops through the array and makes sure all of those status codes generate a retry.

readme.md
szmarczak: I'm glad you've asked :smiley: Got retries on timeouts by default, so `522` and `524` are OK.

On the other hand, `520` is caused by a server error which is not likely to disappear:

image

`521` is `web server is down`, it makes no sense to retry.

`523` is `origin is unreachable` - CloudFlare cannot find a route to the server, it makes no sense to retry either.

`527` is `Railgun Listener to origin error` and it has these common causes:

image

It isn't safe to retry on `527` as we don't know if the request timed out or not.

szmarczak:
> I didn't add a test because I figured a test would sort of mirror the code (e.g. effectively asserting that the array of status codes is what it is in the code).

Ah, you're right. I don't think more tests are necessary at this point.

readme.md
dguo: Thanks for the explanation. I made the change.

szmarczak: We retry on `ECONNREFUSED` and `ECONNRESET` though, so `521` is appriopriate here: https://community.cloudflare.com/t/community-tip-fixing-error-521-web-server-is-down/42461

Let me finish this for you :)

szmarczak:
Good job! :raised_hands:

PR #995, opened 2019-12-14
Deprecate Electron support

### LINKED ISSUE #899: Fix Electron usage
#### Describe the bug

 - Node.js version: 12.13.0
 - Electron version : 7.0.0 
 - OS & version: linux mint 19.2

When using got in electron, useElectronNet option is not working with electron 7 for POST requests, returns `net::ERR_INVALID_ARGUMENT` error.

#### Actual behavior

HTTP request returns the following error : 
```
GotError [RequestError]: net::ERR_INVALID_ARGUMENT
      at ClientRequest.<anonymous> (/home/vincent/lls/lls-desktop/node_modules/got/source/request-as-event-emitter.js:192:19)
      at Object.onceWrapper (events.js:291:20)
      at ClientRequest.emit (events.js:208:15)
      at ClientRequest.origin.emit (/home/vincent/lls/lls-desktop/node_modules/@szmarczak/http-timer/source/index.js:37:11)
      at URLRequest../lib/browser/api/net.js.URLRequest._emitRequestEvent (electron/js2c/browser_init.js:2284:24) {
    name: 'RequestError',
    host: 'jsonplaceholder.typicode.com',
    hostname: 'jsonplaceholder.typicode.com',
    method: 'POST',
    path: '/posts',
    socketPath: undefined,
    protocol: 'https:',
    url: 'https://jsonplaceholder.typicode.com/posts',
    gotOptions: {
      path: '/posts',
      protocol: 'https:',
      slashes: true,
      auth: null,
      host: 'jsonplaceholder.typicode.com',
      port: null,
      hostname: 'jsonplaceholder.typicode.com',
      hash: null,
      search: null,
      query: null,
      pathname: '/posts',
      href: 'https://jsonplaceholder.typicode.com/posts',
      retry: {
        retries: [Function],
        methods: [Set],
        statusCodes: [Set],
        errorCodes: [Set]
      },
      headers: {
        'user-agent': 'got/9.6.0 (https://github.com/sindresorhus/got)',
        accept: 'application/json',
        'accept-encoding': 'gzip, deflate',
        'content-type': 'application/json',
        'content-length': 39
      },
      hooks: {
        beforeRequest: [],
        beforeRedirect: [],
        beforeRetry: [],
        afterResponse: [],
        beforeError: [],
        init: []
      },
      decompress: true,
      throwHttpErrors: true,
      followRedirect: true,
      stream: false,
      form: false,
      json: true,
      cache: false,
      useElectronNet: true,
      body: '{"title":"foo","body":"bar","userId":1}',
      method: 'POST'
    }
  }

```

#### Expected behavior

Correct HTTP response

#### Code to reproduce

In electron main or renderer process : 

```js
got
    .post('https://jsonplaceholder.typicode.com/posts', {
      body: {
        title: 'foo',
        body: 'bar',
        userId: 1
      },
      json: true,
      useElectronNet: true
    })
    .then(console.log)
    .catch(console.log)
```

### LINKED ISSUE #990: `Error: Premature close` on DELETE requests after upgrading to 10.0.3 (from 9.6.0)
#### Describe the bug

We updated from 9.6.0 to 10.0.3 and as soon as it was deployed, we started getting `Error: Premature close` on most or all of our `DELETE` requests. The `GET` and `POST` requests work fine.

Maybe some cases were missed in 7b2ccb0d69fca9aba27373ab344f97d389d692f2?

 - Node.js version: v12.13.1

#### Actual behavior

```
Error: Premature close
  File "internal/streams/end-of-stream.js", line 61, col 15, in IncomingMessage.onclose
  File "events.js", line 210, col 5, in IncomingMessage.emit
  File "domain.js", line 498, col 23, in IncomingMessage.EventEmitter.emit
  File "/home/node/guardian-website/node_modules/@szmarczak/http-timer/dist/index.js", line 40, col 20, in IncomingMessage.origin.emit
    return emit(event, ...args);
  File "_http_client.js", line 363, col 11, in TLSSocket.socketCloseListener
  File "events.js", line 215, col 7, in TLSSocket.emit
  File "domain.js", line 498, col 23, in TLSSocket.EventEmitter.emit
  File "net.js", line 659, col 12, in null.<anonymous>
  File "_tls_wrap.js", line 481, col 7, in TCP.done
```

#### Expected behavior

We shouldn't get a `Premature close` error on successful `DELETE` requests.

#### Code to reproduce

```js
  const url = "https://example.com";
  const username = "username";
  const password = "password";
  const body = undefined;

  const resp = await got(url, {
    username,
    password,
    method: "DELETE",
    throwHttpErrors: false,
    responseType: "json",
    json: body ? body : undefined,
  });
```

#### Checklist

szmarczak:
Fixes #899 
Fixes #990

#### Checklist

- [ ] I have included some tests. (cannot run `ava` + `electron` yet)


szmarczak:
TODO: fix caching with electron

sindresorhus:
Can you move the changes that fixes #990 into a separate PR?

sindresorhus:
Can you remove the added Electron workarounds? There's no point in fixing Electron issues now that we're deprecating support for it.

szmarczak:
> Can you move the changes that fixes #990 into a separate PR?

Will do, I just did so because without these changes Electron support is fully broken though.

test/error.ts
```diff
@@ -193,7 +193,9 @@ test('errors are thrown directly when options.stream is true', t => {
 	});
 });
 
-test('the old stacktrace is recovered', async t => {
+// Blocked by https://github.com/istanbuljs/nyc/issues/619
+// eslint-disable-next-line ava/no-skip-test
+test.skip('the old stacktrace is recovered', async t => {
```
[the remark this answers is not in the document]
szmarczak: I fails randomly on Node 10...

szmarczak:
> Can you remove the added Electron workarounds?

All the other fixes are required except the `redirect` fix. It's only 29 lines long. Should I remove it?

szmarczak:
Some other unnecessary fixes:
 - `session` uses the value from `electronSession` (by default it's a TLS session) - 4 lines
- it automatically throws if no auth info provided (the promise would never resolve otherwise) - 10 lines

The `content-length` header fix is required as without it POST requests would not be possible anymore.

sindresorhus:
It’s not about the amount of lines, but rather that everything adds complexity, and we have a lot of complexity already. Doesn’t really make sense to add complexity for something we intend to remove. These fixes might even cause problems to other things.

szmarczak:
Summary:

The last known Got version that was Electron-compatible was v9.6.0. At the time the Electron version was `v3.1.1`. A few days ago `v7.1.5` has been released.

https://github.com/electron/electron/commit/2a3793485f71b8ec102a324ab61804c8c093a21e kicked in and broke reading error responses.
https://github.com/electron/electron/commit/b7defaaf6a9b866757b475d0452fbb93de9a39cd kicked in and broke sending CORS headers.
So, Got v9.6.0 has been no longer compatible with Electron v7.0.0 and up,
which has been released on October 22 - about two months ago.

szmarczak:
> These fixes might even cause problems to other things.

I don't think so, I've tested it locally. Anyway, I'll remove those fixes :)

People who switched to Electron 7 two months ago or switched to Got 10 this month - had to disable `useElectronNet`, there was no other solution. But yeah, not everybody keeps their dependencies up-to-date :P

szmarczak:
> Can you move the changes that fixes #990 into a separate PR?

Is that really needed? It's not so complicated. Basically the fix is this: https://github.com/sindresorhus/got/pull/995/files#diff-7bf54112e60a8714e49d3e1e98cf4e2bR32-R36

test/error.ts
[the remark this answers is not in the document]
sindresorhus: That would be good to include in the code comment for posterity.

sindresorhus:
> Is that really needed? It's not so complicated.

It's not about how complicated it is. It's weird having that change in a commit (when merged to master) called "Deprecate Electron support", when it's totally unrelated. This also makes it harder when I do release notes, as I have to remember to mention that change manually instead of `np` including the commit title for me. In addition, it makes it harder to review as I need to sometimes guess whether a change is related to that fix or to the Electron stuff.

> Basically the fix is this: sindresorhus/got/pull/995/files#diff-7bf54112e60a8714e49d3e1e98cf4e2bR32-R36

Looks like it's also: https://github.com/sindresorhus/got/pull/995/files#diff-267e4380ccf0cfebc8b885a948acf226L142-L149 And some other smaller changes.

source/normalize-arguments.ts
```diff
@@ -420,6 +420,8 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 				path,
 				host: ''
 			};
+
+			delete options.url;
```
sindresorhus: What is this? Seems like it should be in a separate commit/PR with a test.

source/normalize-arguments.ts
```diff
@@ -431,14 +433,6 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 		options.lookup = options.dnsCache.lookup;
 	}
 
-	/* istanbul ignore next: electron.net is broken */
-	// No point in typing process.versions correctly, as
-	// `process.version.electron` is used only once, right here.
-	if (options.useElectronNet && (process.versions as any).electron) {
-		const electron = dynamicRequire(module, 'electron') as any; // Trick webpack
-		options.request = electron.net.request ?? electron.remote.net.request;
-	}
```
sindresorhus: I think we should just leave this. No other Electron-related code changes. That's the least risky way to handle it. Just add a deprecation message here.

sindresorhus:
Regarding the fix for #990, feel free to just commit that directly to master. It doesn't have to be a PR, just a commit with a good title.

source/normalize-arguments.ts
szmarczak: Just forgot to delete this when removing electron fixes, it's safe to remove this as `options.url` is deleted later anyway

source/normalize-arguments.ts
szmarczak: Do you mean leave this as is or move code from `electron-request.ts` to `request-as-event-emitter.ts`? Personally I prefer to store it in a separate file so it doesn't clutter the Got core.

If you mean the first, then no, it won't work as Electron needs fixing to support POST etc.

sindresorhus: > If you mean the first, then no, it won't work as Electron needs fixing to support POST etc.

My argument was that we don't add any Electron fixes. Electron will work (or rather not work) like it currently does. It's much easier to deprecate it if it's not even working.

szmarczak: I see. Will do :D

source/request-as-event-emitter.ts
```diff
@@ -56,20 +56,6 @@ export default (options: NormalizedOptions): RequestAsEventEmitter => {
...
-							if (name === 'trailers' || name === 'rawTrailers') {
-								return [];
-							}
-
-							const value = (target as any)[name];
-							return is.function_(value) ? value.bind(target) : value;
-						}
-					});
-				}
-
```
sindresorhus: Can you leave this one? We shouldn't remove any fixes as that would be a breaking change.

source/normalize-arguments.ts
```diff
@@ -436,7 +436,11 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 	// `process.version.electron` is used only once, right here.
 	if (options.useElectronNet && (process.versions as any).electron) {
 		const electron = dynamicRequire(module, 'electron') as any; // Trick webpack
-		options.request = electron.net.request ?? electron.remote.net.request;
+		options.request = deprecate(
+			electron.net.request ?? electron.remote.net.request,
+			'Electron support has been deprecated and will be removed in Got 11',
```
sindresorhus: Can you add a link to https://github.com/sindresorhus/got/issues/899? So people can easily know why we are doing this.

PR #1008, opened 2019-12-27
Make sure `got.mergeOptions(...)` can be assigned to defaults

### LINKED ISSUE #993: Cannot assign `got.mergeOptions(...)` to instance defaults

Even if I set `mutableDefaults` as true, (TS 3.7.3, got 10.0.4), TS still complains that `defaults` is a read-only property. The readme example of the `afterResponse` hook doesn't seem to work when using TS.

According to the definition file, `defaults` may or may not be `ReadOnly`.

If I just invoke `client.mergeOptions(client.defaults.options, updatedOptions);` instead of reassigning `client.defaults.options` it seems to work fine but in that case it works even if I don't set `mutableDefaults` as true.

Any ideas or have I just misunderstood how it works?

#### Checklist

- [ x] I have read the documentation.

szmarczak:
Fixes #993

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -56,7 +56,7 @@ interface GotFunctions {
 
 export interface Got extends Record<HTTPAlias, GotFunctions>, GotFunctions {
 	stream: GotStream;
-	defaults: Defaults | Readonly<Defaults>;
+	defaults: Defaults;
 	GotError: typeof errors.GotError;
 	CacheError: typeof errors.CacheError;
 	RequestError: typeof errors.RequestError;
--- a/source/types.ts
+++ b/source/types.ts
@@ -220,27 +220,6 @@ export interface NormalizedOptions extends Options {
 	path?: string;
 }
 
-/*
-export interface NormalizedOptions extends DefaultOptions {
-	dnsCache: CacheableLookup | false;
-	lookup?: CacheableLookup['lookup'];
-	url: URL;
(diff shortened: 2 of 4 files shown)


source/types.ts
```diff
@@ -202,10 +203,44 @@ export interface NormalizedOptions extends Options {
...
+	followRedirect: boolean;
+	useElectronNet: boolean;
+	methodRewriting: boolean;
+	context: {[key: string]: any};
+
 	// UNIX socket support
 	path?: string;
 }
 
+/*
```
szmarczak: Cannot decide which is better. Both solutions work. Should be `NormalizedOptions` inherited from `Options` or `DefaultOptions`?

sindresorhus: Semantically, it makes more sense for it to inherit from `Options`, but I might be missing some context. Are there any downsides/upsides to either way?

szmarczak: It's just that `DefaultOptions` are already normalized but some properties are **excluded** there. 

I agree. Let's go with `Options`.

source/create.ts
```diff
@@ -56,7 +56,7 @@ interface GotFunctions {
 
 export interface Got extends Record<HTTPAlias, GotFunctions>, GotFunctions {
 	stream: GotStream;
-	defaults: Defaults | Readonly<Defaults>;
+	defaults: Defaults;
```
szmarczak: No point in having `Defaults | Readonly<Defaults>` as in some cases TS misguesses it as `Readonly<Defaults>` and people have to cast it to `Defaults` anyway. The JS runtime will throw anyway if anyone tries to modify frozen defaults.

PR #1017, opened 2020-01-09
Export `GotRequestMethod` TypeScript type

mrhyde:
At this moment to be able to correctly define `got` methods as dependencies we need to import the module into type declaration files
```typescript
import got from 'got'

interface Dependencies {
  readonly get: typeof got.get
}
```

with introduced changes we can simply do as so:
```typescript
import { GotFunctions } from 'got'

interface Dependencies {
  readonly get: GotFunctions
}
```

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1016,7 +1016,7 @@ const addAccessToken = (accessToken: string): BeforeRequestHook => (options) => 
 }
 ```
 
- ## Errors
+## Errors
 
 Each error contains an `options` property which are the options Got used to create a request - just to make debugging easier.
 
--- a/readme.md
+++ b/readme.md
@@ -998,7 +998,7 @@ Got has a set of handy types and interfaces available, so you can integrate Got 
 
 ### Got
 
-Typescript will automatically infer types for Got instances but in case you want to define something like dependencies you can use following: `Got`, `GotStream`, `ReturnStream`, `GotRequestMethod`, `GotReturn`
+Typescript will automatically infer types for Got instances but in case you want to define something like dependencies you can import available types directly from `got`
 
 ```ts
 import {GotRequestMethod} from 'got';
--- a/readme.md
+++ b/readme.md
@@ -39,6 +39,7 @@ For browser usage, we recommend [Ky](https://github.com/sindresorhus/ky) by the 
 - [WHATWG URL support](#url)
 - [Hooks](#hooks)
 - [Instances with custom defaults](#instances)
+- [Types](#types)
 - [Composable](documentation/advanced-creation.md#merging-instances)
 - [Plugins](documentation/lets-make-a-plugin.md)
 - [Used by 3000+ packages and 1.6M+ repos](https://github.com/sindresorhus/got/network/dependents)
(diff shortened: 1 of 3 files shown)


mrhyde:
Failed tests are not related to this PR

sindresorhus:
I don't think `GotFunctions` is a good name to expose (I do realize we named it). I should be singular. Maybe something like `GotRequestMethod`? And it needs a doc comment so users would know what it's for.

mrhyde:
@sindresorhus let me know if readme section is ok

readme.md
```diff
@@ -991,7 +992,31 @@ Default: `false`
...
+### Hooks
+Following types can be used to define isolated hooks and keep their interfaces consistent: `Hooks`, `HookEvent`, `HookType`, `InitHook`, `BeforeRequestHook`, `BeforeRedirectHook`, `BeforeRetryHook`, `BeforeErrorHook`, `AfterResponseHook`
+
+```typescript
+const addAccessToken = (accessToken: string): BeforeRequestHook => (options) => {
+  options.path = `${options.path}?access_token=${accessToken}`
+}
+```
+
+ ## Errors
```
szmarczak: There's an unnecessary space before `## Errors`

mrhyde: fixed

readme.md
```diff
@@ -991,6 +992,32 @@ Default: `false`
 
 A read-only boolean describing whether the defaults are mutable or not. If set to `true`, you can [update headers over time](#hooksafterresponse), for example, update an access token when it expires.
 
+## Types
+
+When used in a TypeScript project, Got has a set of handy types and interfaces available.
+
+### Got
+
+Typescript will automatically infer types for Got instances but in case you want to define something like dependencies you can use following: `Got`, `GotStream`, `ReturnStream`, `GotRequestMethod`, `GotReturn`
```
sindresorhus: I'm not sure the value of just listing these types. The user has no idea what they do or what to use them for. Maybe better to just mention one type with example and refer to the type definition instead. It's also annoying to have to keep this in sync with the source.

Same with hooks.

source/create.ts
```diff
@@ -39,7 +39,7 @@ type OptionsOfJSONResponseBody = Merge<Options, {isStream?: false; resolveBodyOn
 type OptionsOfBufferResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType: 'buffer'}>;
 type ResponseBodyOnly = {resolveBodyOnly: true};
 
-interface GotFunctions {
+export interface GotRequestMethod {
```
sindresorhus: Add a doc comment here describing what it is and how it might be useful to user.s

readme.md
```diff
@@ -994,25 +994,27 @@ A read-only boolean describing whether the defaults are mutable or not. If set t
 
 ## Types
 
-Since Got was rewritten in TypeScript it has a set of handy types and interfaces available
+When used in a TypeScript project, Got has a set of handy types and interfaces available.
```
```suggestion
Got has a set of handy types and interfaces available, so you can integrate Got with your TypeScript project seamlessly.
```

readme.md
mrhyde: I was thinking exactly the same when I was writing it, but I had to start somewhere. I'll try to shorten things up

mrhyde:
@sindresorhus all issues mentioned above should be resolved

readme.md
```diff
@@ -998,7 +998,7 @@ Got has a set of handy types and interfaces available, so you can integrate Got
 
 ### Got
 
-Typescript will automatically infer types for Got instances but in case you want to define something like dependencies you can use following: `Got`, `GotStream`, `ReturnStream`, `GotRequestMethod`, `GotReturn`
+Typescript will automatically infer types for Got instances but in case you want to define something like dependencies you can import available types directly from `got`
```
szmarczak: Missing dot at the end

mrhyde: Fixed

sindresorhus:
Thanks :)

PR #1035, opened 2020-01-22
Reject normalization errors with CancelableRequest

### LINKED ISSUE #1025: Incorrect url causes `.json is not a function` error
#### Describe the bug

 - Node.js version: 12.4.0
 - OS & version: Arch Linux Linux 5.4.11-arch1-1

When a url given to the got.post / got.get functions is incorrect AND the function is followed with a .json(), this error is thrown:

```
    TypeError: got.post(...).json is not a function
```

#### Actual behavior

Error is thrown (see above)

#### Expected behavior

Throw HTTP Error 404, Not found

#### Code to reproduce

```js
function test(){
  return got.get('undefined/https://jsonplaceholder.typicode.com/todos/1').json();
};

test();
```

#### Checklist

callmehiphop:
Closes #1025

In the event of an error being thrown during argument normalization a normal promise was being returned which prevented users from calling any of the CancelableRequest convenience methods. This changes makes it so a rejected CancelableRequest is returned instead.

#### Checklist

### DIFF
--- a/test/error.ts
+++ b/test/error.ts
@@ -194,7 +194,8 @@ test('errors are thrown directly when options.stream is true', t => {
 });
 
 test('normalization errors using convenience methods', async t => {
-	await t.throwsAsync(got('undefined/https://example.com').json());
+	const url = 'undefined/https://example.com';
+	await t.throwsAsync(got(url).json().text().buffer(), `Invalid URL: ${url}`);
 });
 
 // Fails randomly on Node 10:
--- a/test/error.ts
+++ b/test/error.ts
@@ -193,6 +193,12 @@ test('errors are thrown directly when options.stream is true', t => {
 	});
 });
 
+test('normalization errors using convenience methods', async t => {
+	const error = await t.throwsAsync(got('undefined/https://example.com'));
+
+	t.true(error.message.includes('Invalid URL'));
+});
+
 // Fails randomly on Node 10:
 // Blocked by https://github.com/istanbuljs/nyc/issues/619
 // eslint-disable-next-line ava/no-skip-test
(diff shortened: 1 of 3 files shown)


test/error.ts
```diff
@@ -193,6 +193,12 @@ test('errors are thrown directly when options.stream is true', t => {
 	});
 });
 
+test('normalization errors using convenience methods', async t => {
+	const error = await t.throwsAsync(got('undefined/https://example.com'));
```
szmarczak:
```suggestion
	await t.throwsAsync(got('undefined/https://example.com').json().text().buffer(), 'Invalid URL');
```

test/error.ts
```diff
@@ -193,6 +193,12 @@ test('errors are thrown directly when options.stream is true', t => {
...
+
+	t.true(error.message.includes('Invalid URL'));
```
szmarczak: This line is unnecessary

test/error.ts
callmehiphop: Fixed!

test/error.ts
callmehiphop: I went ahead and removed it, PTAL!

test/error.ts
szmarczak: You forgot to test `.text()` and `.buffer()`. Make sure that the error message is `Invalid URL`. As in the suggestion.

callmehiphop: Ah, I must have missed the `Invalid URL` part. Do you want a separate assertion for each method?

szmarczak: No need to

szmarczak:
Looks good, thanks! 🙌

PR #1031, opened 2020-01-20
Add failing test for DELETE with request body

nexdrew:
Here's a minimal failing test to help move #1028 forward.

Currently only tests a JSON request body.

I wasn't sure where to add the test, so I created a new file, similar to `test/post.ts`.

I apologize if I'm not using types as you'd like (e.g. defining an Express `Handler`) - I'm not used to TypeScript.

Let me know what changes you'd like and I'll make them. Thanks.

PR #1037, opened 2020-01-23
Allow request body to be sent with a DELETE request

### LINKED ISSUE #1028: Is request body no longer allowed in DELETE request?

Is there a way to send a request body in a DELETE request?

Similar to #966 but specific to a DELETE request.

This was possible in got 9 like so:

```js
got(url, {
  method: 'DELETE',
  json: true,
  body: { version: 0 }
})
```

But attempting this in got 10 does not transmit a request body:

```js
got(url, {
  method: 'DELETE',
  responseType: 'json',
  json: { version: 0 }
})
```

I attempted to use `json` and `body` (via `JSON.stringify()`), with and without specifying a `Content-Type: application/json` header.

Am I doing it wrong?

I don't see any mention of this change in the [v10 release notes](https://github.com/sindresorhus/got/releases/tag/v10.0.0).

The best I can find about this in any RFC is [here](https://tools.ietf.org/html/rfc7231#section-4.3.5):

> A payload within a DELETE request message has no defined semantics; sending a payload body on a DELETE request might cause some existing implementations to reject the request.

which does not seem to expressly forbid it.

#### Checklist

nexdrew:
Fixes #1028.

Adds DELETE to the methods that apply a `content-length` header when a request body is given, to allow the request body to actually be sent.

Fixes the expected failing test added in #1031 (by using streams directly instead of requiring the body to be parsed) and removes the `failing` modifier.

Thanks to @jrnelson90 for finding the solution!

### DIFF
--- a/test/post.ts
+++ b/test/post.ts
@@ -262,12 +262,8 @@ test('the `form` payload is not touched', withServer, async (t, server, got) => 
 	});
 });
 
-test.failing('DELETE method sends plain objects as JSON', withServer, async (t, server, got) => {
-	server.delete('/', (request, response) => {
-		// Not using streams here to avoid this unhandled rejection:
-		// Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
-		response.json(request.body);
-	});
+test('DELETE method sends plain objects as JSON', withServer, async (t, server, got) => {
+	server.delete('/', defaultEndpoint);
 
 	const {body} = await got.delete({
 		json: {such: 'wow'},
--- a/source/normalize-arguments.ts
+++ b/source/normalize-arguments.ts
@@ -404,7 +404,7 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 	// body.
 	if (is.undefined(headers['content-length']) && is.undefined(headers['transfer-encoding'])) {
 		if (
-			(options.method === 'POST' || options.method === 'PUT' || options.method === 'PATCH') &&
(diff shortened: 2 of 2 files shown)


test/post.ts
```diff
@@ -262,12 +262,8 @@ test('the `form` payload is not touched', withServer, async (t, server, got) =>
...
 });
 
-test.failing('DELETE method sends plain objects as JSON', withServer, async (t, server, got) => {
-	server.delete('/', (request, response) => {
-		// Not using streams here to avoid this unhandled rejection:
-		// Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
-		response.json(request.body);
-	});
+test('DELETE method sends plain objects as JSON', withServer, async (t, server, got) => {
+	server.delete('/', defaultEndpoint);
```
szmarczak: Can you undo this change? This needs to be (not necessarily in this PR) fixed, not covered up.

nexdrew: I can undo it, but it's going to continue to fail. My test wasn't written correctly in the first place.

To fix the test independently, there are a couple options:

1. Use the streaming `defaultEndpoint` and attempt to catch the "premature close" error to avoid an unhandled rejection
2. Use a different test server configuration for this test that supports JSON body parsing

Would you prefer a separate PR that can be merged prior to this one?

sindresorhus: > Would you prefer a separate PR that can be merged prior to this one?

Can be a separate PR. I want to get out a new release now.

PR #1038, opened 2020-01-23
Replace `readonly []` with `ReadonlyArray` for pre-3.4 TS versions

lukelafountaine:
Since `readonly []` syntax was [released in TypeScript 3.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#improvements-for-readonlyarray-and-readonly-tuples), consumers of this library that have an earlier TS compiler are unable to compile. The functionality is the same, but this syntax is at least backwards compatible.

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -77,7 +77,7 @@ export interface GotStream extends Record<HTTPAlias, ReturnStream> {
 	(url: URLOrOptions, options?: Options): ProxyStream;
 }
 
-const aliases: readonly HTTPAlias[] = [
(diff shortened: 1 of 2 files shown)


sindresorhus:
TypeScript 3.4 was released a year ago. I don’t want to have to limit ourselves to such an old compiler. It’s hard enough to support just the latest TS compiler. While this seems like a simple change, it opens the door for us having to deal with other TS 3.4 things, and I’m just not willing to do that, sorry. I’ll add a note in the readme about the minimum required TS version.

PR #1049, opened 2020-02-05
Fix readme typo

Jehong-Ahn:


readme.md
```diff
@@ -255,7 +255,7 @@ Example:
 const body = await got(url).json();
```
sindresorhus: Can you update this one too?

Jehong-Ahn: `json()`, `buffer()`, `text()` returns body only through `shortcut`.
https://github.com/sindresorhus/got/blob/master/source/as-promise.ts#L161

sindresorhus: Ah, duuh. I was confused.

We should make this clearer though:

> The promise has also `.text()`, `.json()` and `.buffer()` methods which set this option automatically.

Here it should say that it also sets `resolveBodyOnly` to true so only the body is returned.

Jehong-Ahn:
@sindresorhus You're right.

sindresorhus:
Thank you! :)

readme.md
szmarczak: > Here it should say that it also sets resolveBodyOnly to true so only the body is returned.

Umm... No? It doesn't (but it still returns just the parsed body):

https://github.com/sindresorhus/got/blob/68fc7ab43d59b5323d406047b2cc7e7b37e05d5e/source/as-promise.ts#L161-L168

sindresorhus: > (but it still returns just the parsed body):

Hm. That's actually worse. We should fix that. It's much more intuitive for the user if we set that option instead of just some magic behavior.

szmarczak: > It's much more intuitive for the user if we set that option instead of just some magic behavior.

I wouldn't name it "magic behavior". It's what you expect. What's magic here? This behavior does not break anything.

OTOH if you were to use just `resolveBodyOnly`:

```js
const promise = got('getSomeJson');

// this can be any async operation
setTimeout(() => {
	// the response has been already received yet
	const json = await promise.json(); // <-- this would return a string
}, 1000);
```

I have even written tests for that...

sindresorhus: > I wouldn't name it "magic behavior". It's what you expect. What's magic here? This behavior does not break anything.

Why does `.json()` even set `responseType` then? It could just do something internally without setting that option. The point is that it makes it easier to explain the behavior on it, if we can just refer to two public options that can reproduce the behavior.

> OTOH if you were to use just resolveBodyOnly:

It's not clear to me what you're trying to show with that example?

szmarczak: > Why does .json() even set responseType then?

It does not either ;)

> It's not clear to me what you're trying to show with that example?

IOW: `promise.json()` wouldn't work if the response has been already received in full.

https://github.com/sindresorhus/got/blob/68fc7ab43d59b5323d406047b2cc7e7b37e05d5e/test/response-parse.ts#L64-L76

szmarczak: > The point is that it makes it easier to explain the behavior on it, if we can just refer to two public options that can reproduce the behavior.

I get it now. Yeah, you can do that but it should be mentioned that these "added" options will be internal and won't be visible in hooks.

The only "magic" here is that all the shortcuts like `.json()` use the original body. So performing `promise.json().text()` will return text. In addition, it allows us to parse the body **after** the response has been received.

szmarczak: Not only that, this also lets us to do this:

```js
console.log(await promise.json()); // object
console.log(await promise); // text
```

If Got was overriding `responseType` on `.json()` then `await promise` would return an object too.

sindresorhus: So what do you recommend we do about this then? Currently the readme is incorrect. We either need to actually make them visible in hooks or fix the readme to not say that it's the same as setting the options, and then we need to explain the behavior of the shortcuts in a different way.

szmarczak: I'll send a PR

Issue #924, opened 2019-11-07
Add test and improve docs for `progress.total`


sindresorhus:
The docs says:

> If it's not possible to retrieve the body size (can happen when streaming), total will be null. - https://github.com/sindresorhus/got#ondownloadprogress-progress

But we have no tests to prove this.

- [ ] Add test
- [ ] Be more specific than "can happen when streaming". In what exact scenarios can it happen?
- [x] Make it `undefined` instead of `null`

PR #1051, opened 2020-02-06
Rewrite Got

### LINKED ISSUE #167: HTTP2 support
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/got/issues/167)

Hi!

Can you consider adding `HTTP2` support to GOT?
Currently the most used client library is this: https://github.com/molnarg/node-http2

Thank you!

> There is a $200.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/got/issues/167).

### LINKED ISSUE #929: Pass error thrown in the `init` hook to `beforeError` hook
Currently that's not possible, the `init` hook is called by `normalizeArguments`, which is catched directly by the Got promise.

### LINKED ISSUE #932: The retry logic should be moved into `as-promise.ts`
Currently it's located at `request-as-event-emitter.ts`

---

Need to also do https://github.com/sindresorhus/got/pull/921#discussion_r346711144 then.

### LINKED ISSUE #954: TypeScript cannot figure out which Got type to use

With `@types/got` we could use type `GotOptions` and be able to pass this to `got.get(options)` however with v10 the types are `string | OptionsOfDefaultResponseBody` and `OptionsOfDefaultResponseBody` and its friends are not available to import at the root of the project. I was wondering if this was on purpose.

What i'm looking for and not sure how to do with v10
```ts
const options: GotOptions = {
	url: 'https://example.com',
};
const response = await got(options);
```

Also, this example from the docs isn't working for me and I'm wondering if it should accept a partial? Opened https://github.com/sindresorhus/got/pull/953 with a demo
```ts
got.mergeOptions(got.defaults.options, {
	responseType: 'json',
});
```

#### Checklist

Really excited about the release of v10 🎉

### LINKED ISSUE #981: Use `error.code` instead of `error.message` to compare errors
https://github.com/sindresorhus/got/commit/7b2ccb0d69fca9aba27373ab344f97d389d692f2#r36353721

### LINKED ISSUE #994: `beforeRequest` hook is not called before the actual request
I had some AWS API integrations using Got v9.6.0 that were working fine until the Got dependency was updated to 10.0.4 by the server when all of a sudden the integrations broke. Despite the updated [example code](https://github.com/sindresorhus/got#aws) for aws4 integration using prefixUrl which I updated to, it still didn't work. Further debugging revealed that not only was aws4 rejecting the v10 options object (complaining it was type object and not string), even if I passed it as a string instead, the resultant options object from aws4 was completely mangled (properties concatenated, etc.) - hence why the integrations broke. As soon as I reverted my code and the Got dependency to 9.6.0 again, everything was back to working fine.

aws4 compatibility may therefore be worth looking at (I'm guessing this is commonly used) as well as the example shown at https://github.com/sindresorhus/got#aws - for now I'll therefore hold at 9.6.0

Thanks,

James

### LINKED ISSUE #1011: mergeOptions isn't merging URLSearchParams instances
#### Describe the bug

 - Node.js version: -
 - OS & version: -

`got.mergeOptions` isn't merging instances of `URLSearchParams`. According to the README is should merge them.

#### Actual behavior
It looks like later instances of `URLSearchParams` just overwrite earlier values. (see code sample)

#### Expected behavior
It should merge the two URLSearchParams instances.

#### Code to reproduce

```js
const got = require('got');
got.mergeOptions({ searchParams: new URLSearchParams({ foo: 1 }) }, { searchParams: new URLSearchParams({ bar: 2 }) });
// => { ... searchParams: URLSearchParams { 'bar' => '2' } }
```

#### Checklist

I think some code got deleted in https://github.com/sindresorhus/got/pull/921. 

`merge.ts` got moved https://github.com/sindresorhus/got/pull/921/files#diff-5e2f9cb0db248d822e1b167d5ccf8e07. Before the move it had code to deal with instances of `URLSearchParams`:
https://github.com/sindresorhus/got/blob/9f4fe33585bf89de37f61ff5614435105be28f64/source/merge.ts#L13-L21

After the move it's missing the code to handle `URLSearchParams`: https://github.com/szmarczak/got/blob/a7e73f2fe89b1eca00d2f8fff71dffb2caf68c16/source/utils/merge.ts

### LINKED ISSUE #1015: Throws if sending body takes longer than full event loop cycle
If you do `got.stream(url, {body: stream, ...options})` or `got.stream(...).pipe(stream)`, it may throw

```
Got\'s stream is not writable when the `body`, `json` or `form` option is used
```

even though you don't do `.write()` on it.

### LINKED ISSUE #1016: The `merge` function is slow

Hi, we are using got heavily in my company to send several thousand requests/s so I'm doing some profiling to see in which places our app spends its time the most.

I've recently upgraded to Got 10 and I've seen that the library spends a significant (though, not alarming) percentage of the time in the normalize arguments phase, which at the same time calls the aforementioned merge function.

Captura de pantalla 2020-01-08 a las 15 50 59

It looks like this single `slice` call to clone the array seems to spend quite a lot of time. So, my question is, do we really need to clone the array or could it be just referenced? If we need a copy, is `slice()` the fastest method? 

I tried https://jsperf.com/cloning-arrays/3 in Chrome 79 and it seems that there're faster (not so much tbh) alternatives.

#### Checklist

### LINKED ISSUE #1021: Using got.post with both body and cache options causes an exception
#### Describe the bug

Using `got.post` with both `body` and `cache` options causes an exception.

 - Node.js version: 13.6.0
 - OS & version: Fedora 30
 - got version: 10.2.2

#### Actual behavior

```
GotError: Cannot read property 'on' of undefined
    at onError (/usr/src/npm/failures/got-test/node_modules/got/dist/source/request-as-event-emitter.js:137:29)
    at EventEmitter.handleRequest (/usr/src/npm/failures/got-test/node_modules/got/dist/source/request-as-event-emitter.js:170:17)
    at destroyer (internal/streams/pipeline.js:27:10)
    at internal/streams/pipeline.js:79:12
    at Array.map (<anonymous>)
    at pipeline (internal/streams/pipeline.js:76:28)
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at pipeline (internal/util.js:296:12)
    at EventEmitter.handleRequest (/usr/src/npm/failures/got-test/node_modules/got/dist/source/request-as-event-emitter.js:162:23)
    at Object.onceWrapper (events.js:428:26)
    at EventEmitter.emit (events.js:321:20) {
  name: 'RequestError'
}
```

#### Expected behavior

No exception.  Unsure if `options.cache` should do anything in this case or be ignored.  For my own use that does not matter as the server I'm really working with never provides cachable responses to POST requests.  Really the goal for me is to be able to use a shared "base options" object and not have to censor the `cache` option when performing posts.

#### Code to reproduce

```js
#!/usr/bin/env node
'use strict';
const got = require('got');

async function main() {
	await got.post('https://httpbin.org/post', {
		body: '',
		cache: new Map()
	});
}

main().catch(error => {
	console.error(error);
	process.exit(1);
});

```

#### Checklist

### LINKED ISSUE #1026: got.stream() swallows errors when used to upload small files with stream.pipeline()
#### Describe the bug

 - Node.js version: v13.5.0
 - OS & version: Linux pc 5.3.18-1-MANJARO #1 SMP PREEMPT Wed Dec 18 18:34:35 UTC 2019 x86_64 GNU/Linux

`got.stream()` stream emits `finish` before `error` in case of `ECONNREFUSED`. The `finish` event should be used to signal that all data has been written/sent once stream.end() has been called (https://nodejs.org/api/stream.html#stream_event_finish).

#### Actual behavior

`stream.pipeline()` promise is resolved.

#### Expected behavior

`stream.pipeline()` should be rejected with `ECONNREFUSED`;

#### Code to reproduce
This code
```js
const fs = require('fs');
const util = require('util');
const stream = require('stream');

const pipeline = util.promisify(stream.pipeline);
const got = require('got');


async function run() {
    await pipeline(
        // Read small file
        fs.createReadStream('/proc/version'),
        // Generates ECONNREFUSED
        got.stream.put('http://localhost:7777')
    )
}

run().then(() => {
    console.log('done');
}).catch((err) => {
    console.log('err', err);
})
```
Prints:
```
done
```
It seems that this is because `stream.pipeline()` considers the data handled if the finish event is received. To isolate the issue I wrote another small test:
```js
const got = require('got');

const u = got.stream.put('http://localhost:7777')
u.on('error', (err) => {
    console.log('err', new Date(), err);
});
u.on('end', () => {
    console.log('end', new Date());
});
u.on('close', () => {
    console.log('close', new Date());
});
u.on('finish', () => {
    console.log('finish', new Date());
});

const writeSomeMore = u.write('test');
console.log('writeSomeMore', writeSomeMore);
u.end();
```
Which prints:
```
writeSomeMore true
finish 2020-01-17T13:50:48.061Z
err 2020-01-17T13:50:48.071Z GotError: connect ECONNREFUSED 127.0.0.1:7777
    at onError (.../node_modules/got/dist/source/request-as-event-emitter.js:137:29)
    at handleRequest (.../node_modules/got/dist/source/request-as-event-emitter.js:170:17)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1136:16) {
  name: 'RequestError',
  code: 'ECONNREFUSED'
}
```
Note that if I read `/dev/urandom` (a bigger file) instead of `/proc/version` the stream.pipeline() promise is rejected.

#### Checklist

- [ x] I have read the documentation.
- [ x] I have tried my code with the latest version of Node.js and Got.

### LINKED ISSUE #1036: Got doesn't pass query-string when calling via unix-socket
#### Describe the bug

 - Node.js version: v10.16.3+ (at least)
 - OS & version: Ubuntu 18.04 Linux 4.15.0-72-generic  (probably any other OS)
- Got: v10.2.2

#### Actual behavior

Making a call via UNIX-socket results in not passing query string in URL. When making a call via normal TCP/IP port it works correctly. For example, for the code below we have received the output:
```
call to /tmp/app.sock, URL = /tmp/app.sock:/http?foo=bar
call to /tmp/app.sock, URL = /got # here should be a query-string
call to 3000, URL = /http?foo=bar
call to 3000, URL = /got?foo=bar
```

#### Expected behavior

Having query string in request in both cases. For example, for the code below it's expected the output:
```
call to /tmp/app.sock, URL = /tmp/app.sock:/http?foo=bar
call to /tmp/app.sock, URL = /tmp/app.sock:/got?foo=bar # here should be a query-string
call to 3000, URL = /http?foo=bar
call to 3000, URL = /got?foo=bar
```
or at least:

```
call to /tmp/app.sock, URL = /tmp/app.sock:/http?foo=bar
call to /tmp/app.sock, URL = /got?foo=bar # here should be a query-string
call to 3000, URL = /http?foo=bar
call to 3000, URL = /got?foo=bar
```

#### Code to reproduce

```js
const fs = require('fs'),
    http = require('http'),
    got = require('got');

const port = 3000,
    socket = '/tmp/app.sock',
    search = '?foo=bar',
    handler = (listen) => (req, res) => {
        console.log(`call to ${listen}, URL = ${req.url}`);

        res.statusCode = 204;
        res.end();
    },
    cb = () => {};

//clean up and establish http server with unix-socket
if (fs.existsSync(socket)) fs.unlinkSync(socket);
http.createServer(handler(port)).listen(port);
http.createServer(handler(socket)).listen(socket);

//reference call with http via TCP/IP port
http.get(
    new URL(`http://localhost:${port}/http${search}`),
    {headers: {authorization: 'Basic aaa'}},
    cb
);

//reference call with got via TCP/IP port
got
    .get(`http://localhost:${port}/got${search}`, {headers: {authorization: 'Basic aaa'}})
    .then(cb);

//reference call with http via unix-socket
http.get(
    new URL(`http://unix:${socket}:/http${search}`),
    {socketPath: '/tmp/app.sock', headers: {authorization: 'Basic aaa'}},
    cb
);

//testing call with got via unix-socket
got
    .get(`http://unix:${socket}:/got${search}`, {headers: {authorization: 'Basic aaa'}})
    .then(cb);
```

#### Checklist

### LINKED ISSUE #1050: Got throws if passed options are already frozen
#### Describe the bug

 - Node.js version: v12.14.1
 - OS & version: Linux AMA-LT-4549 4.15.0-76-generic #86~16.04.1-Ubuntu SMP Mon Jan 20 11:02:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Currently the snippet provided in the `code to reproduce` section fails.

`got` tries to mutate the `options` object. This currently impacts test scenarii where the same `options` object is used between different tests.

#### Actual behavior

```
> node .

(node:12424) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'body' of undefined
    at /home/user/dev/got-issue/index.js:12:32
(node:12424) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:12424) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

#### Expected behavior

No mutation.

#### Code to reproduce

```js
const got = require('got');
const options = {};

Object.freeze(options);

(async () => {
  try {
    const response = await got('https://sindresorhus.com', options);
    console.log(response.body);
    //=> '<!doctype html> ...'
  } catch (error) {
    console.log(error.response.body);
    //=> 'Internal server error ...'
  }
})();
```

#### Checklist

### LINKED ISSUE #1057: Got doesn't throw on leading slashes
As per the readme:

> Note: Leading slashes in input are disallowed when using this option to enforce consistency and avoid confusion. For example, when the prefix URL is https://example.com/foo and the input is /bar, there's ambiguity whether the resulting URL would become https://example.com/foo/bar or https://example.com/bar. The latter is used by browsers.

But it does not throw:

https://github.com/sindresorhus/got/blob/3c6ca29000707dfad1d51edd013748107806404d/source/normalize-arguments.ts#L303-L319

### LINKED ISSUE #1058: A way to customize `cacheable-lookup`
Hello,

I noted `dnsCache` is used only for setup cache adapter at `cacheable-lookup`: 

https://github.com/sindresorhus/got/blob/1f363b976cd23626462314c7cd5670cddd951c73/source/normalize-arguments.ts#L147

Since I'm interested in customize some `cacheable-lookup` capabilities, (such as [resolver](https://github.com/szmarczak/cacheable-lookup#optionsresolver)) I want to ask if you will accept a way for passing a `cacheable-lookup` instance as `dnsCache` to be used 🙂

### LINKED ISSUE #1090: HTTP authentication leak in redirects
#### Describe the bug

A custom authorization header, if present, is passed on redirect. This can leak auth tokens, and in many cases requests with unexpected authorization headers will be rejected.

#### Actual behavior

If a custom authorization header is added to options, and the request is redirected (302), the authorization header will be passed to the redirect location.

In my case, my app was using got to fetch a tarball from npm.pkg.github.com which redirected to pkg.githubusercontent.com, which rejected the request (400).

#### Expected behavior

curl does not forward custom authorization: https://curl.haxx.se/docs/CVE-2018-1000007.html

It is convenient that got follows redirects by default, but it should not forward authorization headers by default.

#### Workaround

Add a `beforeRedirect` hook to remove the authorization header.

Example: https://github.com/joebowbeer/regsync/blob/master/src/util.ts#L54

#### Checklist

### LINKED ISSUE #1113: Spaces should be normalized as `+` in query strings

Some web services like the VLC Web Interface use `%20` to represent spaces in search params. However, Got automatically forces `+` with no way to override it even if put through `query-string`.

#### Describe the feature

The simplest solution would be to allow such overrides using the current syntax:

```js
const got = require("got")
const queryString = require("query-string")

await got(`http://192.168.1.2/requests/status.json`), {
    port,
    password,
    searchParams: queryString.stringify({
        command,
        ...options,
    }).replace(/\+/, "%20")
})
```

The current workaround is to insert it directly into the url:

```js
const got = require("got")
const queryString = require("query-string")

await got(`http://192.168.1.2/requests/status.json?${queryString.stringify({
    command,
    ...options,
}).replace(/\+/, "%20")}`), {
    port,
    password,
})
```

See: https://github.com/Richienb/vlc/blob/62f395cd471cf9cad0c540bad3aade8d2a7faa6f/src/index.ts#L169-L175

#### Checklist

### LINKED ISSUE #1118: Cannot reuse an options object
#### Describe the bug

 - Node.js version: v12.14.0
 - OS & version: OSx Catalina

When trying to make an API call, I am getting the following error: `(node:73533) UnhandledPromiseRejectionWarning: TypeError: The `url` option cannot be used if the input is a valid URL.` . When I copy and paste the exact URL being tried into Postman, it works fine

#### Actual behavior
The error is being thrown and the API call is not being made.
...

#### Expected behavior
The API would be made.
...

#### Code to reproduce
Unfortunately, I cannot give an entirely repoduceable code snippet, because it requires a personal API key, but I believe the problem happens before the request is actually made. This is the code used to made the call:
```
     getSummonerBySummonerId(summonerId: string): Promise<any> {
    return got(
      encodeURI(
        'https://na1.api.riotgames.com/lol/summoner/v4/summoners/xWheQXcH0U1J20HleXa5dFJ137QleMldaJtSCxHK9ljsnKAv',
      ),
      lolApiConstants.httpOptions,
    );
  }

const lolApiConstants = {
  baseURL: 'https://na1.api.riotgames.com',
  httpOptions: {
    headers: {
      'content-type': 'application/json',
      'X-Riot-Token': process.env.LOL_API_KEY,
    },
    responseType: 'json',
    resolveBodyOnly: true,
  },
};
```
```
(node:73533) UnhandledPromiseRejectionWarning: TypeError: The `url` option cannot be used if the input is a valid URL.
    at Object.exports.normalizeArguments (/Users/blubberbo/Web Projects/Lol-Pickem/lol-pickem-api/node_modules/got/dist/source/normalize-arguments.js:241:23)
    at got (/Users/blubberbo/Web Projects/Lol-Pickem/lol-pickem-api/node_modules/got/dist/source/create.js:52:58)
    at LolApiService.getSummonerBySummonerId (/Users/blubberbo/Web Projects/Lol-Pickem/lol-pickem-api/src/services/lol-api.service.ts:54:12)
    at GameService.<anonymous> (/Users/blubberbo/Web Projects/Lol-Pickem/lol-pickem-api/src/services/game.service.ts:31:8)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/blubberbo/Web Projects/Lol-Pickem/lol-pickem-api/dist/services/game.service.js:5:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:73533) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:73533) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

#### Checklist

- [x ] I have read the documentation.
- [x ] I have tried my code with the latest version of Node.js and Got.

### LINKED ISSUE #1129: Modify response headers while using `got.stream(...)`
So you can `got.stream(...).pipe(serverResponse)` with the modified ones.

### LINKED ISSUE #1130: got actually need nodejs >= 10.9, not >= 10
#### Describe the bug

 - Node.js version: 10.2
 - OS & version: MacOS

got actually need nodejs >= 10.9, not >= 10

the request function prototype used in GOT was appeared in NodeJS 10.9:

https://github.com/sindresorhus/got/blob/master/source/request-as-event-emitter.ts#L246

https://nodejs.org/docs/latest-v10.x/api/http.html#http_http_request_options_callback

image

#### Checklist

### LINKED ISSUE #1138: Errors  have undefined body when using streams
#### Describe the bug

 - Node.js version: 12.6.1
 - OS & version: Macos 10.14

#### Actual behavior
When `isStream: true` is set and request results in HTTPError, the stream emits `error` event. The emit handler is only called with `error` argument and `body` and `response` arguments are `undefined`.

...

#### Expected behavior
As per the [docs](https://github.com/sindresorhus/got#onerror-error-body-response), the `body` and `response` arguments should be set with `error` event.
...

#### Code to reproduce

```js
const got = require('got');

let stream = got("https://www.google.com/non-existant-path", {
  isStream: true,
  method: "GET",
  headers: {}
});

stream.on("response", (response) => {
  console.log(response.headers);
  console.log(response.statusCode)
});

stream.on("error", (err, body, response) => {
  // err is set
  console.log(err);

  // body should have been set but it's undefined
  console.log(body);

  // response should have been set but it's undefined.
  console.log(response);
});

```

#### Checklist

### LINKED ISSUE #1140: Pagination should ignore `resolveBodyOnly` option
https://github.com/sindresorhus/got/issues/1052#issuecomment-606696546

szmarczak:
### Bug fixes

1. Fixes #1036 (tested)
2. Fixes #994 (tested)
3. Fixes #1026 (tested)
4. Fixes #1021 (tested)
5. Fixes #1057 (tested)
6. Fixes #1050 (tested)
7. Fixes #954 (tested)
8. Fixes #1011 (tested)
9. Fixes #1090 (tested)
10. Fixes #1140 (tested)
11. Fixes #1118 (tested)

#### API

- `options.encoding` will no longer affect streams (reverting back to the Got 9 behavior), use `got.stream(...).setEncoding(encoding)` instead (to prevent confusion)
- renamed `GotError` to `RequestError`
- `options.agent` can no longer be an instance of `http.Agent` - instead you need to pass:

```js
{
	http: new http.Agent(...),
	https: new https.Agent(...),
	http2: new http2wrapper.Agent(...)
}
```

- `options.useElectronNet` has been removed
- `dnsCache` is now enabled by default
- `dnsCache` no longer can be a Map or a Keyv instance, instead you need to pass a [CacheableLookup](https://github.com/szmarczak/cacheable-lookup) instance
- errors thrown in `init` hooks will be converted to `RequestError`s (why: `RequestError` provides much more useful information (e.g. Got options) than the usual TypeError etc.)
- `options._pagination` has been renamed to `options.pagination`
- the `options.url` property will no longer be visible in init hooks (design limitation)
- the `error.request` property is no longer a `ClientRequest` instance - instead, it gives a Got stream

#### Types

- renamed the `Defaults` type to `InstanceDefaults`
- renamed the `DefaultOptions` type to `Defaults`
- renamed the `DefaultRetryOptions` type to `RequiredRetryOptions`
- renamed the `GotOptions` type to `Options`
- the `ResponseObject` type has been merged into the `Response` type
- renamed the `GotRequestMethod` type to `GotRequestFunction`

### Enhancements

1. Fixes #1016 (benchmarked)
2. Fixes #981 (refactored)
3. Fixes #932 (refactored)
4. Fixes #929 (tested)
5. Fixes #1058 (tested)
6. Fixes #167 (tested, but there's only 1 test as of now, blocked by #1093)
7. Fixes #1138 (tested)
8. Fixes #1113 (tested)
9. Fixes #1129 (tested)
10. The errors now have a `request` property (it's a Got stream).

### Miscellaneous

1. Fixes #1130

### Known bugs

1. The `timings` indicate that the request was successful (even though it errored).
2. The `downloadProgress` object may show incorrect data if the request has errored.

## TODO

- [x] remove unused deps
- [x] remove unused dev deps if any
- [x] #924 (PR: #1061)
- [x] #167
- [x] #722 (PR: #833)
- [x] #1046
- [x] merge #1081
- [x] as-promise cleanup
- [x] ~core cleanup~ (can't extract the types into a different file because `Options`, `RequestError`s and `Request` need each other)
  - [ ] optional: extract `normalizeArguments` into a different file
- [x] revert back hard deprecation of legacy URL options
- [x] `create.ts` cleanup (will have a follow-up PR)
- [x] benchmark Got
- [x] test breaking changes
- [x] changelog
- [x] Rename `options._pagination` to `options.pagination`
- [x] Port #1099
- [x] docs

#### FEEL FREE TO TEST THIS

Note that this has ~yet un~documented **breaking** changes, ~but~ everything should work as expected - all tests pass. TypeScript recommended.

```bash
git clone https://github.com/szmarczak/got.git
git checkout rework
npm install
npm run build
```

```js
const got = require('./dist/source');

(async () => {
	console.log(await got('https://httpbin.org/anything'));
})();
```


sindresorhus:
I got chills down my spine when I read the PR title 😅

I hope you plan to add lots of tests. I'm very paranoid that we'll end up breaking a lot of stuff that we're not currently testing.

szmarczak:
> I hope you plan to add lots of tests.

Yup. Every fix & every change in behavior is gonna be tested. 

> I'm very paranoid that we'll end up breaking a lot of stuff that we're not currently testing.

Don't worry! There's gonna be a **very** detailed changelog. I'll explain why things are done like that and not in another way.

benchmark/index.ts
```diff
@@ -0,0 +1,200 @@
...
+
+	internalSuite.run();
+};
+
+// Results (i7-7700k, CPU governor: performance):
+// got - promise                          x 3,092 ops/sec ±5.25% (73 runs sampled)
+// got - stream                           x 4,313 ops/sec ±5.61% (72 runs sampled)
+// got - promise core                     x 6,756 ops/sec ±5.32% (80 runs sampled)
+// got - stream core                      x 6,863 ops/sec ±4.68% (76 runs sampled)
+// got - stream core - normalized options x 7,960 ops/sec ±3.83% (81 runs sampled)
```
szmarczak: Normalizing cuts out ~1k req/s but the time is needed to duplicate the options, so the original ones won't be altered if you use hooks to modify the request.

benchmark/index.ts
```diff
@@ -0,0 +1,200 @@
...
+	}).on('cycle', (event: Benchmark.Event) => {
+		console.log(String(event.target));
+	});
+
+	internalSuite.run();
+};
+
+// Results (i7-7700k, CPU governor: performance):
+// got - promise                          x 3,092 ops/sec ±5.25% (73 runs sampled)
+// got - stream                           x 4,313 ops/sec ±5.61% (72 runs sampled)
```
szmarczak: I don't know why, but the `got` function cuts out 3k-4k req/s.

package.json
```diff
@@ -10,7 +10,7 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"test": "xo && tsc --noEmit && nyc ava",
+		"test": "xo && tsc --noEmit && nyc --reporter=html --reporter=text ava",
```
szmarczak: Very useful to debug later. Generates both HTML output and text output as usual.

source/as-promise/core.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+	Options,
+	NormalizedOptions,
+	Defaults,
+	ResponseType
+} from './types';
+import Request, {knownHookEvents, RequestError, HTTPError} from '../core';
+
+if (!knownHookEvents.includes('beforeRetry' as any)) {
+	knownHookEvents.push('beforeRetry' as any, 'afterResponse' as any);
```
szmarczak: The normalizing will be done via `Request.normalizeArguments` so we just tell `Request` that it needs to normalize `beforeRetry` too.

source/as-promise/core.ts
```diff
@@ -0,0 +1,139 @@
...
+		} catch (error_) {
+			this.destroy(error_);
+			return;
+		}
+
+		if (this._throwHttpErrors && !isHTTPError) {
+			this.destroy(error);
+		} else {
+			this.emit('error', error);
+		}
```
szmarczak: We cannot destroy this if the `retry` functionality is enabled. This is the responsibility of the Got promise.

source/as-promise/index.ts
```diff
@@ -0,0 +1,242 @@
...
+				const isOk = (): boolean => {
+					const {statusCode} = response;
+					const limitStatusCode = options.followRedirect ? 299 : 399;
+
+					return (statusCode >= 200 && statusCode <= limitStatusCode) || statusCode === 304;
+				};
+
+				// Download body
+				try {
+					body = await getStream.buffer(request);
```
szmarczak: Note that this is `request`, not `response`. If we were downloading directly from the `response` there would be no progress events emitted.

source/as-promise/types.ts
```diff
@@ -0,0 +1,149 @@
+import PCancelable = require('p-cancelable');
```
szmarczak: This file just adds promise-specific options.

source/as-stream.ts
```diff
@@ -1,144 +0,0 @@
-import duplexer3 = require('duplexer3');
```
szmarczak: Got no longer uses EventEmitter to emit the request and the response. It is based directly on Duplex streams.

source/as-stream.ts
```diff
@@ -1,144 +0,0 @@
...
-			proxy._read = (...args) => {
-				isFinished = true;
-
-				proxy._read = read;
-				return read.apply(proxy, args);
-			};
-		}
-
-		if (options.encoding) {
-			proxy.setEncoding(options.encoding);
```
szmarczak: Breaking change: `options.encoding` will no longer affect streams (it's the same behavior as in Got 9), people should do `got.stream(...).setEncoding(encoding)` instead. It should be mentioned in the docs that this option is for promises only.

source/as-stream.ts
```diff
@@ -1,144 +0,0 @@
...
-		}
-
-		// We cannot use `stream.pipeline(...)` here,
-		// because if we did then `output` would throw
-		// the original error before throwing `ReadError`.
-		response.pipe(output);
```
szmarczak: The new implementation doesn't do this. It just proxies the `_read` calls (basically a Duplex stream).

source/core/utils/options-to-url.ts
```diff
@@ -0,0 +1,73 @@
+/* istanbul ignore file: deprecated */
```
szmarczak: Note: `searchParams` is handled via the `Request` class.

source/errors.ts
```diff
@@ -1,125 +0,0 @@
-import is from '@sindresorhus/is';
-import {Timings} from '@szmarczak/http-timer';
-import {TimeoutError as TimedOutError} from './utils/timed-out';
-import {Response, NormalizedOptions} from './types';
-
-export class GotError extends Error {
```
szmarczak: Breaking: renamed `GotError` to `RequestError`

source/normalize-arguments.ts
```diff
@@ -1,536 +0,0 @@
...
-	options.throwHttpErrors = Boolean(options.throwHttpErrors);
-	options.ignoreInvalidCookies = Boolean(options.ignoreInvalidCookies);
-	options.cache = options.cache ?? false;
-	options.responseType = options.responseType ?? 'text';
-	options.resolveBodyOnly = Boolean(options.resolveBodyOnly);
-	options.followRedirect = Boolean(options.followRedirect);
-	options.dnsCache = options.dnsCache ?? false;
-	options.useElectronNet = Boolean(options.useElectronNet);
-	options.methodRewriting = Boolean(options.methodRewriting);
-	options.allowGetBody = Boolean(options.allowGetBody);
```
szmarczak: This has been dropped to get a 1.5k req/s increase. It will be inherited from defaults so it's unnecessary either because the defaults define these booleans.

source/index.ts
```diff
@@ -52,21 +53,22 @@ const defaults: Defaults = {
 			beforeError: [],
 			afterResponse: []
 		},
+		cache: undefined,
```
szmarczak: `false` -> `undefined`. Objects should either objects or undefined. People can do `instance.extend({cache: false})` to disable caching later (if the instance has them enabled).

source/index.ts
```diff
@@ -52,21 +53,22 @@ const defaults: Defaults = {
...
+		dnsCache: new CacheableLookup(),
```
szmarczak: Breaking change: DNS cache is on by default.

source/progress.ts
```diff
@@ -1,48 +0,0 @@
-import EventEmitter = require('events');
-import {Transform as TransformStream} from 'stream';
-import is from '@sindresorhus/is';
-
-export function createProgressStream(name: 'downloadProgress' | 'uploadProgress', emitter: EventEmitter, totalBytes?: number | string): TransformStream {
```
szmarczak: This has been removed in favor of the Duplex stream (yeah, we can do things like this there).

source/request-as-event-emitter.ts
```diff
@@ -1,345 +0,0 @@
...
-	const requestUrl = options.url.toString();
-	const redirects: string[] = [];
-	let retryCount = 0;
-
-	let currentRequest: http.ClientRequest;
-
-	// `request.aborted` is a boolean since v11.0.0: https://github.com/nodejs/node/commit/4b00c4fafaa2ae8c41c1f78823c0feb810ae4723#diff-e3bc37430eb078ccbafe3aa3b570c91a
-	const isAborted = (): boolean => typeof currentRequest.aborted === 'number' || (currentRequest.aborted as unknown as boolean);
-
-	const emitError = async (error: GeneralError): Promise<void> => {
```
szmarczak: This has been renamed to `request._beforeError`.

source/request-as-event-emitter.ts
```diff
@@ -1,345 +0,0 @@
...
-				timer(request);
-				timedOut(request, options.timeout, options.url);
-
-				emitter.emit('request', request);
-
-				const uploadStream = createProgressStream('uploadProgress', emitter, httpOptions.headers!['content-length'] as string);
-
-				isPiped = true;
-
-				await pipeline(
```
szmarczak: We don't need to use `pipeline` if we do `Duplex`.

source/request-as-event-emitter.ts
```diff
@@ -1,345 +0,0 @@
...
-			// Catches errors thrown by calling `requestFn(…)`
-			try {
-				handleRequest(httpOptions[requestSymbol](options.url, httpOptions, handleResponse));
-			} catch (error) {
-				emitError(new RequestError(error, options));
-			}
-		}
-	};
-
-	emitter.retry = error => {
```
szmarczak: This has been moved to `as-promise/index.ts`.

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-export interface Response<BodyType = unknown> extends IncomingMessageWithTimings {
-	body: BodyType;
-	statusCode: number;
-
-	/**
-	The remote IP address.
-
-	Note: Not available when the response is cached. This is hopefully a temporary limitation, see [lukechilds/cacheable-request#86](https://github.com/lukechilds/cacheable-request/issues/86).
-	*/
-	ip: string;
```
szmarczak: This has been corrected to `ip?: string;`

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	methods: Method[];
-	statusCodes: number[];
-	errorCodes: string[];
-	calculateDelay: RetryFunction;
-	maxRetryAfter?: number;
-}
-
-export interface RetryOptions extends Partial<DefaultRetryOptions> {
-	retries?: number;
-}
```
szmarczak: Why is it even here? It has been removed in Got 10...

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-
-export type RequestFunction = typeof http.request;
-
-export interface AgentByProtocol {
-	http?: http.Agent;
-	https?: https.Agent;
-}
```
szmarczak: Breaking change: `options.agent` can no longer be an instance of `http.Agent`. Instead, you need to pass an object like:

```
{
	http: new http.Agent(...),
	https: new https.Agent(...),
	http2: new http2wrapper.Agent(...)`
}
```

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-			'url' |
-			typeof requestSymbol
-		>
-	>,
-	{
-		hooks: Required<Hooks>;
-		retry: DefaultRetryOptions;
-		timeout: Delays;
-		context: {[key: string]: any};
-		_pagination?: PaginationOptions<unknown>['_pagination'];
```
szmarczak: We should make it public in Got 12 if there are no people giving feedback rn.

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	isStream?: boolean;
-	encoding?: BufferEncoding;
-	method?: Method;
-	retry?: RetryOptions | number;
-	throwHttpErrors?: boolean;
-	cookieJar?: ToughCookieJar | PromiseCookieJar;
-	ignoreInvalidCookies?: boolean;
-	request?: RequestFunction;
-	agent?: http.Agent | https.Agent | boolean | AgentByProtocol;
-	cache?: string | CacheableRequest.StorageAdapter | false;
```
szmarczak: Breaking change:

```diff
-instance('https://example.com', {cache: false})
+const newInstance = instance.extend({cache: false});
+newInstance('https://example.com');
```

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	headers?: Headers;
-	responseType?: ResponseType;
-	resolveBodyOnly?: boolean;
-	followRedirect?: boolean;
-	prefixUrl?: URL | string;
-	timeout?: number | Delays;
-	dnsCache?: CacheableLookup | Map<string, string> | Keyv | false;
-	useElectronNet?: boolean;
```
szmarczak: Breaking change: this has been removed.

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
+	// `asStream` usage
+	(url: string | URL, options?: Options & {isStream: true}): Request;
 
-export interface ExtendOptions extends Options {
-	handlers?: HandlerFunction[];
-	mutableDefaults?: boolean;
-}
+	(options: Options & {isStream: true}): Request;
 
-export interface Defaults {
```
szmarczak: `Defaults` -> `InstanceDefaults`

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	options: DefaultOptions;
```
szmarczak: `DefaultOptions` -> `Defaults`

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	};
+	// A bug.
+	// eslint-disable-next-line @typescript-eslint/adjacent-overload-signatures
+	<T>(options?: Options & PaginationOptions<T>): AsyncIterableIterator<T>;
+	// A bug.
+	// eslint-disable-next-line @typescript-eslint/adjacent-overload-signatures
+	all<T>(options?: Options & PaginationOptions<T>): Promise<T[]>;
 }
 
-export interface GotOptions extends PaginationOptions<unknown> {
```
szmarczak: `GotOptions` -> `Options`

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
+	mutableDefaults: boolean;
+	_rawHandlers?: HandlerFunction[];
 }
 
-export type RetryFunction = (retryObject: RetryObject) => number;
-
+export type GotReturn = Request | CancelableRequest;
 export type HandlerFunction = <T extends GotReturn>(options: NormalizedOptions, next: (options: NormalizedOptions) => T) => T | Promise<T>;
 
-export interface DefaultRetryOptions {
```
szmarczak: `DefaultRetryOptions` -> `RequiredRetryOptions`

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
-	redirectUrls: string[];
-	request: {
-		options: NormalizedOptions;
-	};
-	url: string;
-}
-
-// TODO: The `ResponseLike` type should be properly fixed instead:
-// https://github.com/sindresorhus/got/pull/827/files#r323633794
-export interface ResponseObject extends Partial<ResponseLike> {
```
szmarczak: No `ResponseObject`. Use `Response`.

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
+	(url: string | URL, options?: Options): CancelableRequest | Request;
 
-export interface Progress {
-	percent: number;
-	transferred: number;
-	total?: number;
+	(options: Options): CancelableRequest | Request;
 }
 
-export interface GotEvents<T> {
```
szmarczak: `GotEvents` -> `RequestEvents`

source/utils/merge.ts
```diff
@@ -1,32 +0,0 @@
-import {URL} from 'url';
```
szmarczak: This is no longer needed because all the necessary options are merged by hand (I mean they are merged still automatically but these are explicit Got options - people are responsible for merging their own options).

source/as-promise/create-rejection.ts
```diff
@@ -0,0 +1,31 @@
+import {CancelableRequest, BeforeErrorHook, RequestError} from './types';
+
+export default function createRejection(error: Error, ...beforeErrorGroups: Array<BeforeErrorHook[] | undefined>): CancelableRequest<never> {
+	const promise = (async () => {
+		if (error instanceof RequestError) {
```
szmarczak: Init hooks will throw `RequestError`s if anybody throws inside that hook.

szmarczak: `RequestError` provides much more useful information (Got options etc.) that the usual `TypeError` etc.

source/as-promise/index.ts
```diff
@@ -0,0 +1,244 @@
...
+	const emitter = new EventEmitter();
+
+	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
+		const makeRequest = (): void => {
+			if (options.responseType === 'json' && options.headers.accept === undefined) {
+				options.headers.accept = 'application/json';
+			}
+
+			// Support retries
+			const {throwHttpErrors} = options;
```
szmarczak: `Request` needs to throw so https://github.com/sindresorhus/got/pull/1051/files#diff-6be2971b2bb8dbf48d15ff680dd898b0R132 gets called.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+	rejectUnauthorized: boolean;
+	lookup?: CacheableLookup['lookup'];
+	methodRewriting: boolean;
+	username: string;
+	password: string;
+	[kRequest]: HttpRequestFunction;
+	[kIsNormalizedAlready]?: boolean;
+}
+
+export interface Defaults {
```
szmarczak: We shouldn't extend `NormalizedOptions` nor `Options` to require strict (aka "full", "complete") values.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+	const onError = (error: Error): void => {
+		reject(error);
+	};
+
+	file.once('error', onError);
+	file.once('open', () => {
+		file.off('error', onError);
+		resolve();
+	});
+});
```
szmarczak: This is just 275 bytes long, there's no need to include `p-event` (34.1kB) in the dependencies.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+			source.on('end', lockWrite);
+		});
+
+		this.on('unpipe', (source: Writable) => {
+			source.off('data', unlockWrite);
+			source.off('data', lockWrite);
+
+			source.off('end', unlockWrite);
+			source.off('end', lockWrite);
+		});
```
szmarczak: This is actually needed when people pass custom `options.body` (stream), so

```
got.post('https://example.com', {body: readableStream}).write('asdf');
```

throws (the current Got implementation allows this behavior).

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+		}
+
+		(async (nonNormalizedOptions: Options) => {
+			try {
+				if (nonNormalizedOptions.body instanceof ReadStream) {
+					await waitForOpenFile(nonNormalizedOptions.body);
+				}
+
+				if (kIsNormalizedAlready in nonNormalizedOptions) {
+					this.options = nonNormalizedOptions as NormalizedOptions;
```
szmarczak: The options may be normalized in the `got` function already.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+			for (const key in rawOptions) {
+				// @ts-ignore Dear TypeScript, all object keys are strings (or symbols which are NOT enumerable).
+				if (is.undefined(rawOptions[key]) && !is.undefined(defaults[key])) {
+					// @ts-ignore See the note above
+					options[key] = defaults[key];
+				}
+			}
+		}
+
+		// TODO: Deprecate URL options in Got 12.
```
szmarczak: People should pass a URL instance instead.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+					host: ''
+				});
+			}
+		}
+
+		const isHttps = url.protocol === 'https:';
+
+		let fallbackFn: HttpRequestFunction;
+		if (options.http2) {
+			fallbackFn = http2wrapper.auto;
```
szmarczak: `options.http2` is `false` by default because big companies set their own agents, and many of them use HTTP1 Agents... Let's turn it on by default in Got 12.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+		}
+
+		options[kRequest] = realFn as HttpRequestFunction;
+		delete options.request;
+		delete options.timeout;
+
+		let requestOrResponse: ReturnType<RequestFunction>;
+
+		try {
+			requestOrResponse = await fn(url, options as unknown as RequestOptions);
```
szmarczak: The `options.request` can be now a promise. This is to make HTTP2 support more easy.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+
+			if (is.undefined(requestOrResponse)) {
+				requestOrResponse = fallbackFn(url, options as unknown as RequestOptions);
```
szmarczak: Future behavior: if `options.request` returns `undefined` (e.g. cache mechanism: no cache found), fallback to native HTTP requests.

source/core/index.ts
```diff
@@ -0,0 +1,1472 @@
...
+				this._progressCallbacks.shift()!();
+			}
+
+			callback(error);
+		});
+	}
+
+	_final(callback: (error?: Error | null) => void): void {
+		const endRequest = (): void => {
+			// FIX: Node.js 10 calls the write callback AFTER the end callback!
```
szmarczak: This is a workaround for Node.js 10.

source/types.ts
```diff
@@ -1,274 +1,121 @@
...
+	// A bug.
+	// eslint-disable-next-line @typescript-eslint/adjacent-overload-signatures
+	<T>(options?: Options & PaginationOptions<T>): AsyncIterableIterator<T>;
+	// A bug.
+	// eslint-disable-next-line @typescript-eslint/adjacent-overload-signatures
+	all<T>(options?: Options & PaginationOptions<T>): Promise<T[]>;
 }
 
-export type Options = Merge<https.RequestOptions, Merge<GotOptions, URLOptions>>;
+export interface GotRequest {
```
szmarczak: `GotRequestMethod` -> `GotRequest`

szmarczak:
Also we shouldn't enable DNS cache by default because of https://github.com/szmarczak/cacheable-lookup/issues/8

source/as-stream.ts
sindresorhus: Should also be documented as a doc comment on the `options.encoding` type.

source/types.ts
sindresorhus: Can you mention all the breaking changes in the PR description?

source/types.ts
sindresorhus: Yup. Feel free to make it public now.

source/core/index.ts
sindresorhus: > Let's turn it on by default in Got 12.

Can you add a TODO comment?

sindresorhus:
> Also we shouldn't enable DNS cache by default because of szmarczak/cacheable-lookup#8

Can you add some more details to the issue? Why does it need to respect `hosts` file manually? I assumed Node.js would do that by default.

szmarczak:
> I assumed Node.js would do that by default.

It does when you use `dns.lookup`, but `cacheable-lookup` makes a **real** DNS query.

szmarczak:
> Can you add some more details to the issue? Why does it need to respect hosts file manually? I assumed Node.js would do that by default.

Fixed that (not accepting IPv6 yet). The `dnsCache` is on by default from now.

source/types.ts
szmarczak: Will do! :D

szmarczak:
Finally, full `hosts` file support! :D

source/core/index.ts
szmarczak: Done, put in `index.ts`

source/types.ts
szmarczak: Renamed `GotRequest` to `GotRequestFunction`

szmarczak:
> ### Breaking changes
> 
> #### API
> 
> - renamed `GotError` to `RequestError`

@sindresorhus Are you okay with this? I can revert that if necessary. Alternatively I can reexport `RequestError` as `GotError`, so this wouldn't be such a breaking change.

sindresorhus:
> renamed GotError to RequestError

What if we need to throw a generic Got error that is not a `RequestError`? Maybe all our errors should inherit from `GotError`?

szmarczak:
> Maybe all our errors should inherit from GotError?

[It's done like that already](https://github.com/sindresorhus/got/blob/master/source/errors.ts). This PR doesn't change that but adds some additional useful information (`request`, `response`). I just renamed `GotError` to `RequestError` for better readability (I think). I'm not sure about this decision so I'm asking you :>

sindresorhus:
Sounds good to me.

szmarczak:
Finished coding. I just need to update the changelog and the docs.

sindresorhus:
Can you double-check that you have ported all commits landed on master since you opened this PR?

szmarczak:
> Can you double-check that you have ported all commits landed on master since you opened this PR into this branch?

Done. Nothing's missing as of now.

szmarczak:
I'll finish the changelog today and will do the docs tomorrow.

szmarczak:
> > Can you double-check that you have ported all commits landed on master since you opened this PR into this branch?
> 
> Done. Nothing's missing as of now.

Also double-checked the types and the logic in general. Didn't notice any bugs other than those in the PR description (just nitpicks, nothing breaking).

sindresorhus:
Wow! This includes a lot of fixes and improvements. Really nice work, @szmarczak! 🎉

joebowbeer:
You might want to list the fix for #1090 as a breaking change in case someone was depending on forwarded authorization headers.

package.json
```diff
@@ -7,10 +7,10 @@
 	"funding": "https://github.com/sindresorhus/got?sponsor=1",
 	"main": "dist/source",
 	"engines": {
-		"node": ">=10"
+		"node": ">=10.19.0"
```
NMinhNguyen, line 10: @szmarczak Sorry to be commenting on a merged PR, but is there a reason you had to bump this? This should probably be mentioned as a breaking change?

PR #1054, opened 2020-02-06
Wait until body file descriptor is open

### LINKED ISSUE #1046: Body stream error is not caught by Got

szmarczak:
Fixes #1046

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -1,3 +1,4 @@
+import {Readable} from 'stream';
 import {Merge} from 'type-fest';
 import is from '@sindresorhus/is';
 import asPromise, {createRejection} from './as-promise';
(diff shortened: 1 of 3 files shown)


source/create.ts
```diff
@@ -99,7 +100,29 @@ const aliases: readonly HTTPAlias[] = [
...
+				(result as ProxyStream).destroy(error);
+				return;
+			}
+
+			(result as CancelableRequest<unknown>).cancel(error.message);
+		};
+
+		options.body.once('error', errorHandler);
+
+		// @ts-ignore Each member of the union type '...' has signatures, but none of those signatures are compatible with each other
```
szmarczak: I think it just says that result can be either or a stream or a promise and it doesn't know what types to choose

szmarczak:
Or it could be done via a promise that waits until `options.body` is writable...

source/request-as-event-emitter.ts
```diff
@@ -1,3 +1,4 @@
+import {ReadStream} from 'fs';
```
sindresorhus: Generally, it's nicer to rename generic imports like this as it's not immediately clear further down in the code that it's an **fs** read stream.

```
import {ReadStream: FsReadStream} from 'fs';
```

szmarczak: Will do :)

PR #1064, opened 2020-02-12
Fix `responseType` docs

szmarczak:


readme.md
```diff
@@ -248,7 +248,35 @@ Default: `'text'`
 
 The parsing method. Can be `'text'`, `'json'` or `'buffer'`.
 
-The promise also has `.text()`, `.json()` and `.buffer()` methods which sets this and the `resolveBodyOnly` option automatically.
+The promise also has `.text()`, `.json()` and `.buffer()` methods which return another Got promise returning the parsed body.\
```
```suggestion
The promise also has `.text()`, `.json()` and `.buffer()` methods which return another Got promise for the parsed body.\
```

readme.md
```diff
@@ -248,7 +248,35 @@ Default: `'text'`
...
+It's like setting the options to `{responseType, resolveBodyOnly: true}` but without affecting the main Got promise.\
```
```suggestion
It's like setting the options to `{responseType: true, resolveBodyOnly: true}` but without affecting the main Got promise.

```

readme.md
```diff
@@ -248,7 +248,35 @@ Default: `'text'`
...
+	// `json` is an object
+})();
+```
+
+Or if you don't care about the Response object:
+
+```js
+(async () => {
+	const bufferPromise = got(url).buffer();
+	const jsonPromise = bufferPromise.json();
```
sindresorhus: Why are you calling `.json()` on the `bufferPromise`? That makes no sense to me.

readme.md
sindresorhus: > affecting the main Got promise

Affecting the main Got promise in what way? That needs to be clearer.

szmarczak: > It's like setting the options to `{responseType: true

`responseType` is a string: `text` | `json` | `buffer`

> Affecting the main Got promise in what way? That needs to be clearer.

It doesn't change the request options. Instead, it clones the promise and instead of returning `response.body` it returns `JSON.parse(response.body)` for example.

readme.md
szmarczak: It's just an example. The example above contains a Request instance, this one does not. 

> Why

`jsonPromise` may throw, so people may want to save the buffer somewhere for debugging purposes.

readme.md
szmarczak: > It's like setting the options to `{responseType: true

`responseType` is a string: `text` | `json` | `buffer`

> Affecting the main Got promise in what way? That needs to be clearer.

It doesn't change the request options. Instead, it clones the promise and instead of returning `response.body` it returns `JSON.parse(response.body)` for example.

readme.md
sindresorhus: I feel like the above example is clear enough and this example only adds confusion and it's not really needed.

readme.md
```diff
@@ -248,7 +248,35 @@ Default: `'text'`
...
+
+	const [buffer, json] = Promise.all([bufferPromise, jsonPromise]);
+	// `buffer` is an instance of Buffer
+	// `json` is an object
+})();
+```
 
 Example:
```
sindresorhus: > // is the same as this

This should probably say:

> // is semantically the same as this

readme.md
sindresorhus: Ok

readme.md
szmarczak: Sure, will remove this.

readme.md
```diff
@@ -249,7 +249,8 @@ Default: `'text'`
 The parsing method. Can be `'text'`, `'json'` or `'buffer'`.
 
 The promise also has `.text()`, `.json()` and `.buffer()` methods which return another Got promise for the parsed body.\
-It's like setting the options to `{responseType, resolveBodyOnly: true}` but without affecting the main Got promise.\
+It's like setting the options to `{responseType: true, resolveBodyOnly: true}` but without affecting the main Got promise.
```
szmarczak: The `responseType` is not a boolean. It's a string. I think this change should be reverted.

sindresorhus: Yeah, not sure what I was doing there. 🤷‍♂️ https://github.com/sindresorhus/got/pull/1064/commits/fe1c6a8f2ea87a841d45e294081fe336405dbea2

PR #1073, opened 2020-02-17
Add fair warning about rfc7231 support

### LINKED ISSUE #966: The GET method cannot be used with a body
#### Describe the bug

 - Node.js version: 12
 - OS & version: Ubuntu

```
TypeError {
  message: 'The `GET` method cannot be used with a body',
}
Object.exports.normalizeRequestArguments (node_modules/got/dist/source/normalize-arguments.js:262:19)
get (node_modules/got/dist/source/request-as-event-emitter.js:39:55)
node_modules/got/dist/source/request-as-event-emitter.js:244:19
```

I ran into this when upgrading to got@10.0.1.

I realize this is up for debate, but to my best knowledge, a lot of HTTP servers _do_ support GET requests with a body. Elasticsearch, for example, heavily uses this. (They also provide an escape hatch with POST though.)

I don't think `got` should be the limiting factor, given that the underlying modules allow it.

#### Checklist

dcharbonnier:
I think many people assume GOT support or will support 100% of the rfc and anything that is not compliant is considered as a bug, with different priority of development but a bug.
We should add a warning so developers can check if this library feed their needs or if they should use a different one.

Close #966

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -26,6 +26,7 @@ For browser usage, we recommend [Ky](https://github.com/sindresorhus/ky) by the 
 
 ## Highlights
 
+- [Partial support of the RFC] (#partial)
 - [Promise API](#api)
 - [Stream API](#streams)
 - [Pagination API (experimental)](#pagination)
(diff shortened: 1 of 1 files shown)


szmarczak:
This already has been discussed

Issue #1074, opened 2020-02-17
Add an option to allow GET with body


szmarczak:
Currently Got throws <code>TypeError: The `GET` method cannot be used with a body</code> when you try to upload some data using the GET method. [The RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3.1) doesn't specify any particular behavior regarding this situation. Undefined behavior is considered to be an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern).

Although this has been discussed in some issues, there are many people who are still interested in this. Sometimes the HTTP servers don't follow the RFC. We have already added an option to [ignore invalid cookies](https://github.com/sindresorhus/got/pull/826). The proposed solution is to disable the error via another option, for example ~`ignoreStrictPayload`~ `allowGetBody`.

sindresorhus:
Wouldn't it be clearer with an option called `allowGetBody`?

szmarczak:
Sure.

sindresorhus:
In addition, to anyone that decides to work on this. The docs should strongly advice against using this option and explain that it's an anti-pattern and that HTTP2 doesn't allow a GET with body. That it's only meant to interact with non-compliant servers when you have no other choice.

dcharbonnier:
Sorry but just in case someone find this thread I need to give some information to avoid the spread of disinformation, people trust what they read on the web.
 - If a server does not accept a body with GET it's up to the server, like a server that would refuse you to do a POST on a certain url.
- If a proxy / a firewall does not allow you to pass a body with a GET request it's a non compliant server, traefik, nginx for example are perfectly compliant, no issue here.
- Sending a body with a GET request is not "considered" an anti-pattern (the link point to  "anti-pattern is on wikipedia  that obviously does not mention the body with GET). It's considered by some as a good pattern for some specific cases, carrying more sense than having an  "artificial" POST for the development of REST api especially (exemple elastic search, facebook at some point), even made it recently in openapi 3.1 (https://github.com/OAI/OpenAPI-Specification/pull/1937) and by others as an anti-pattern with probably good points too.
[...]

PR #1081, opened 2020-02-19
Add option `allowGetBody` to allow GET requests with payload

dopecodez:
#### Checklist

Fix for #1074

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -188,7 +188,7 @@ Type: `string | Buffer | stream.Readable` or [`form-data` instance](https://gith
 
 **Note #2:** If you provide this option, `got.stream()` will be read-only.
 
-**Note #3:** If you provide a payload with the `GET` or `HEAD` method, it will throw a `TypeError`.
+**Note #3:** If you provide a payload with the `HEAD` method, it will throw a `TypeError`. If you provide a a payload with a 'GET' method, without specifying `allowGetBody` as true, it will also throw a `TypeError`. See [here](https://github.com/dopecodez/got#allowgetbody) for more information regarding this.
 
 The `content-length` header will be automatically set if `body` is a `string` / `Buffer` / `fs.createReadStream` instance / [`form-data` instance](https://github.com/form-data/form-data), and `content-length` and `transfer-encoding` are not manually set in `options.headers`.
 
@@ -415,7 +415,7 @@ By default, redirects will use [method rewriting](https://tools.ietf.org/html/rf
 Type: `boolean`\
 Default: `false`
 
-By default, GET methods are not allowed to have a body. This option is to force GET requests to have a body. Please note that this is an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and NOT RECOMMENDED. In many servers, the body in a GET request could be ignored or even lead to errors. Please read the [RF 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) docs for more information. Please note this option is only enabled for interacting with non-compliant servers when you have no other choice.
+By default, GET methods are not allowed to have a payload. This option is to force GET requests to have a payload. Please note that this is an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and NOT RECOMMENDED. In many servers, the payload in a GET request could be ignored or even lead to errors. Please read the [RF 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) docs for more information. Please note this option is only enabled for interacting with non-compliant servers when you have no other choice.
 
 **Note:**  **Not Reccomended**
 
--- a/readme.md
+++ b/readme.md
@@ -188,7 +188,7 @@ Type: `string | Buffer | stream.Readable` or [`form-data` instance](https://gith
 
 **Note #2:** If you provide this option, `got.stream()` will be read-only.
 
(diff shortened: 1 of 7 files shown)


test/arguments.ts
```diff
@@ -158,10 +158,17 @@ test('throws when passing body with a non payload method', async t => {
 	// @ts-ignore Error tests
 	await t.throwsAsync(got('https://example.com', {body: 'asdf'}), {
 		instanceOf: TypeError,
-		message: 'The `GET` method cannot be used with a body'
+		message: 'The `GET` method cannot be used with a body unless `allowGetBody` option is set to true(Not reccomended)'
```
szmarczak:
```suggestion
		message: 'The `GET` method cannot be used with a body'
```

readme.md
```diff
@@ -410,6 +410,15 @@ Default: `true`
 
 By default, redirects will use [method rewriting](https://tools.ietf.org/html/rfc7231#section-6.4). For example, when sending a POST request and receiving a `302`, it will resend the body to the new location using the same HTTP method (`POST` in this case).
 
+###### allowGetBody
+
+Type: `boolean`\
+Default: `false`
+
+By default, GET methods are not allowed to have a payload. This option is to force GET requests to have a payload. Please note that this is an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and NOT RECOMMENDED. In many servers, the payload in a GET request could be ignored or even lead to errors. Please read the [RF 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) docs for more information. Please note this option is only enabled for interacting with non-compliant servers when you have no other choice.
```
```suggestion
**Note:** The [RFC 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) doesn't specify any particular behavior for the GET method having a payload, therefore **it's considered an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern)**.

Set this to `true` to allow sending body for the `GET` method. However, the [HTTP/2 specification](https://tools.ietf.org/html/rfc7540#section-8.1.3) says that `An HTTP GET request includes request header fields and no payload body`, therefore when using the HTTP/2 protocol this option will have no effect.
```

readme.md
```diff
@@ -410,6 +410,15 @@ Default: `true`
...
+
+**Note:**  **Not Reccomended**
```
```suggestion
```

source/normalize-arguments.ts
```diff
@@ -386,6 +388,10 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 			throw new TypeError(`The \`${options.method}\` method cannot be used with a body`);
 		}
 
+		if (!options.allowGetBody && (isBody || isForm || isJson) && withoutBodyUnlessSpecified === options.method) {
+			throw new TypeError(`The \`${options.method}\` method cannot be used with a body unless \`allowGetBody\` option is set to true(Not reccomended)`);
```
szmarczak:
```suggestion
			throw new TypeError(`The \`${options.method}\` method cannot be used with a body`);
```

readme.md
sindresorhus: > This option is to force GET requests to have a payload

This makes it sound like Got is actually "forcing" it to have a body. However, the user is still free to not provide one. It just "allows" the user to supply a body.

test/arguments.ts
```diff
@@ -158,10 +158,17 @@ test('throws when passing body with a non payload method', async t => {
...
 	});
 });
 
+test('passes when passing option allowGetBody with a GET method', withServer, async (t, server, got) => {
```
sindresorhus:
```suggestion
test('option `allowGetBody` with a GET method', withServer, async (t, server, got) => {
```

test/post.ts
```diff
@@ -20,10 +20,16 @@ const echoHeaders: Handler = (request, response) => {
 	response.end(JSON.stringify(request.headers));
 };
 
-test('GET cannot have body', withServer, async (t, server, got) => {
+test('GET cannot have body without option allowGetBody', withServer, async (t, server, got) => {
```
sindresorhus:
```suggestion
test('GET cannot have body without the `allowGetBody` option', withServer, async (t, server, got) => {
```

readme.md
dopecodez:
```suggestion
By default, GET methods are not allowed to have a payload. This option is to allow GET requests to have a payload. Please note that this is an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and NOT RECOMMENDED. In many servers, the payload in a GET request could be ignored or even lead to errors. Please read the [RF 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) docs for more information. Please note this option is only enabled for interacting with non-compliant servers when you have no other choice.
```

Changing the force to allow should make more sense then.

dopecodez:
```suggestion
By default, GET methods are not allowed to have a payload. This option is to allow GET requests to have a payload. The [RFC 7321](https://tools.ietf.org/html/rfc7231#section-4.3.1) doesn't specify any particular behavior for the GET method having a payload, therefore it's considered an [anti-pattern](https://en.wikipedia.org/wiki/Anti-pattern) and NOT RECOMMENDED. Please note this option is only enabled for interacting with non-compliant servers when you have no other choice.
```

sindresorhus:
Nice work, @dopecodez  👍

PR #1099, opened 2020-03-03
Fix `got.paginate(...)` typings

jaulz:
Currently, if you use `got.paginate` or `got.paginate.all` you cannot properly use the pagination options because `GotOptions` extend `PaginationOptions<unknown>`: https://github.com/sindresorhus/got/blob/526b4bb03eed5196a3c6375beeb7699e8f512d4f/source/types.ts#L173

This would result in unknown types:
```ts
got.paginate.all<Item>('/test', { 
  _pagination: {
    // item is of type unknown instead of Item
    filter: (item) => true
  }
})
```

#### Checklist

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -62,10 +62,11 @@ export interface GotRequestMethod {
 }
 
 export type GotPaginateOptions<T> = Except<Options, keyof PaginationOptions<unknown>> & PaginationOptions<T>;
+export type URLOrGotPaginateOptions<T> = string | GotPaginateOptions<T>;
 
 export interface GotPaginate {
-	<T>(url: string | GotPaginateOptions<T>, options?: GotPaginateOptions<T>): AsyncIterableIterator<T>;
-	all<T>(url: string | GotPaginateOptions<T>, options?: GotPaginateOptions<T>): Promise<T[]>;
+	<T>(url: URLOrGotPaginateOptions<T>, options?: GotPaginateOptions<T>): AsyncIterableIterator<T>;
+	all<T>(url: URLOrGotPaginateOptions<T>, options?: GotPaginateOptions<T>): Promise<T[]>;
 }
 
 export interface Got extends Record<HTTPAlias, GotRequestMethod>, GotRequestMethod {
@@ -200,7 +201,7 @@ const create = (defaults: Defaults): Got => {
 	}
 
 	// @ts-ignore The missing property is added below
-	got.paginate = async function * <T>(url: string | GotPaginateOptions<T>, options?: GotPaginateOptions<T>) {
+	got.paginate = async function * <T>(url: URLOrGotPaginateOptions<T>, options?: GotPaginateOptions<T>) {
 		let normalizedOptions = normalizeArguments(url as URLOrOptions, options as Options, defaults);
 
 		const pagination = normalizedOptions._pagination!;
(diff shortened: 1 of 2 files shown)


jaulz:
Hm, not sure if I got your point. Currently, you can pass the type of the item as a generic to the pagination function and you have to explicitly type the response type for `transform`.

szmarczak:
Yeah, but instead of passing the options every time you can extend the instance, right?

jaulz:
Got it. For `extend` it's of course necessary to explicitly add typings but I still don't understand why it's not a good approach? Why do we have it as a generic at all then?

https://github.com/sindresorhus/got/blob/cf4fdad553f4bec82a9972dda40f5762f8c70c6d/source/types.ts#L163-L171

szmarczak:
> Why do we have it as a generic at all then?

So you can define your own options as `PaginationOptions<Blah>` (`const options = ...`) and cast it to `PaginationOptions<unknown>` later :)

One of the "solutions" would be to use `any` instead of `unknown` but it's not a solution as it doesn't require you to type the items properly.

@sindresorhus What do you think?

jaulz:
@sindresorhus any opinion? 😊

jaulz:
@szmarczak yep, I replaced it...

sindresorhus:
I'm fine with this.

jaulz:
@szmarczak could you please review this PR then again? 😊

source/create.ts
```diff
@@ -61,9 +61,11 @@ export interface GotRequestMethod {
 	<T>(url: string | Merge<Options, {isStream: true}>, options?: Merge<Options, {isStream: true}>): ProxyStream<T>;
 }
 
+export type GotPaginateOptions<T> = Except<Options, keyof PaginationOptions<unknown>> & PaginationOptions<T>;
+
 export interface GotPaginate {
-	<T>(url: URLOrOptions & PaginationOptions<T>, options?: Options & PaginationOptions<T>): AsyncIterableIterator<T>;
-	all<T>(url: URLOrOptions & PaginationOptions<T>, options?: Options & PaginationOptions<T>): Promise<T[]>;
+	<T>(url: string | GotPaginateOptions<T>, options?: GotPaginateOptions<T>): AsyncIterableIterator<T>;
```
szmarczak: I think `string | GotPaginateOptions<T>` should be `URLOrOptions & GotPaginateOptions<T>`

jaulz: That would duplicate work but also duplicate the options because `GotPaginateOptions` is defined like this:
```ts
export type GotPaginateOptions<T> = Except<Options, keyof PaginationOptions<unknown>> & PaginationOptions<T>;
```

Though, I just pushed a commit which adds the `URLOrGotPaginateOptions` type so we have that definition in one place.

szmarczak:
It's only missing a test

jaulz:
@szmarczak shouldn't the existing unit tests cover it?

szmarczak:
> shouldn't the existing unit tests cover it?

What do you mean? We don't test **your** changes, we only tests the current code.

jaulz:
> > shouldn't the existing unit tests cover it?
> 
> What do you mean? We don't test **your** changes, we only test the current code.

I assumed that the changes are only about types and the build would anyway fail if there would be any typing issues. Anyway, now I added some explicit types that should cover the new behaviour. If you have something else in mind, let me know.

szmarczak:
Looks good!

jaulz:
Thanks for the inputs!

sindresorhus:
Thanks :)

PR #1102, opened 2020-03-03
Avoid parsing JSON twice in `_pagination.transform`

jaulz:
If the `responseType` is `json` we do not need to parse the response again in the default transform function.

#### Checklist

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -69,6 +69,10 @@ const defaults: Defaults = {
 		context: {},
 		_pagination: {
 			transform: (response: Response) => {
+				if (response.request.optinos.responseType === 'json') {
+					return response.body;
+				}
+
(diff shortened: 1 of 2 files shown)


szmarczak:
This needs a test :+1:

szmarczak:
Instead of the check I'd just put `normalizedOptions.responseType = 'text';` in https://github.com/sindresorhus/got/blob/62f3f3840f892fa62758773addc86188375c0c71/source/create.ts#L203

jaulz:
I would rather use the passed options and not change them in between. That might lead to an unexpected behavior in the future.

szmarczak:
Hmmm. You're right.

szmarczak:
Good job! :raised_hands:

Issue #945, opened 2019-11-26
Gzip seems not work in my project


grapehunter:
#### What would you like to discuss?
Enviroment: 
Running in render process with Electron 7 (But not use Electron.net.)

Code: 
`got('http://www.example.com', { decompress: true }).then(r => console.log(r.body));`

Result:
![image](https://user-images.githubusercontent.com/3354458/69620507-73044100-1078-11ea-9d5e-d8b023e145ce.png)

It looks like the gzip data is not being decoded correctly

#### Checklist

- [x] I have read the documentation.
- [x] I have removed node_modules and reinstalled all dependencies by `yarn install`.
- [x] I have removed .yarnclean and reinstalled all dependencies by `yarn install`.

szmarczak:
What Got version are you running?

grapehunter:
> What Got version are you running?

The latest release version with `yarn install got`, I think it is 9.6.0

grapehunter:
Is it possible that the relevant code was cleared by yarn clean?
Here is my .yarnclean
```
# test directories
__tests__
test
tests
powered-test

# asset directories
docs
doc
website
images
assets

# examples
example
examples

# code coverage directories
coverage
.nyc_output

# build scripts
Makefile
Gulpfile.js
Gruntfile.js

# configs
appveyor.yml
circle.yml
codeship-services.yml
codeship-steps.yml
wercker.yml
.tern-project
.gitattributes
.editorconfig
.*ignore
.eslintrc
.jshintrc
.flowconfig
.documentup.json
.yarn-metadata.json
.travis.yml

# misc
# *.md
*.gz
```

grapehunter:
I have found out the reason.
In get-response.js
```
const decompressResponse = require('decompress-response');
```
decompressResponse is not a function but a empty object. I don't know why so far

szmarczak:
Glad you've solved the issue! PR welcome :)

szmarczak:
It used to work some while ago. Now electron support is broken.

adityapk00:
Was this fixed? It seems to be broken with electron, and fails with the same error.

szmarczak:
Electron support has been deprecated, as the `electron.net`API differs very much from the Node js `http` API.

ci010:
> I have found out the reason.
> In get-response.js
> 
> ```
> const decompressResponse = require('decompress-response');
> ```
> 
> decompressResponse is not a function but a empty object.
> decompress-response was disabled in package.json
> 
[...]

PR #1105, opened 2020-03-04
Do not ignore `decompress-response` when bundling

ci010:
Currently, using got with webpack to build target `electron-renderer` will throw an error from `get-response.js` like:

```
decompressResponse is not a Function
```

The root cause is webpack ignoring the `decompress-response` module in got, according the `browser` setting in package.json.

Like the issue #945 described, after removing this field, the webpack will bundle this module and `got` works correctly.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,6 @@
 	"types": "dist/source",
 	"sideEffects": false,
 	"browser": {
-		"decompress-response": false,
 		"electron": false
 	},
 	"ava": {


szmarczak:
Electron support has been deprecated and will be completely removed in Got 11

ci010:
> Electron support has been deprecated and will be completely removed in Got 11

I'm not sure if I understand this right. The removal of "browser" fields is already planned in Got 11, and I can just wait the Got 11 release. 

From my understanding, the deprecation of electron means Got does not support for electron `net` module, but it should still works in `electron-renderer` with `nodeIntegration` enabled (using node http/https modules).

This field just makes the webpack bundle wrongly, and Got will not work even if the `nodeIntegration` is enabled in `electron-renderer`.

Issue #1120, opened 2020-03-11
On `afterResponse` token renewal, chained `.text()`|`.json()`|`.buffer()` method returns the wrong body


PopGoesTheWza:
#### Describe the bug

 - Node.js version: 12.16.1
 - OS & version: MacOS Catalina

Given a `client` Got instance with a `afterResponse` hook to set/renew OAuth2 token upon 401 `response.statusCode`

```typescript
const unchained = await client(url, options)
// unchained.body holds the correct (as a string) json result (i.e. after token properly set)

const chained = await client(url, options).json()
// chained is the json result of the 401 response (i.e. before token is properly set)
```

#### Actual behavior

When `.json()` is chained with a got instance with an `afterResponse` hook, it returns the json result of the first request and ignores (or overwrites?) the result of the second (and valid) request. 

#### Expected behavior

When `.json()` is chained with a got instance with an `afterResponse` hook, it should return the json result of the last request. 

#### Code to reproduce

Add the following to `test\hooks.ts` and run `npx tsc && npx ava test/hooks.ts`

```typescript
test('afterResponse with retry as correct .json()', withServer, async (t, server, got) => {
	server.get('/', (request, response) => {
		if (request.headers.token !== 'unicorn') {
			response.statusCode = 401;
			response.end(JSON.stringify({hello: 'nasty'}));
		} else {
			response.end(JSON.stringify({hello: 'world'}));
		}
	});

	const body = await got({
		hooks: {
			afterResponse: [
				(response, retryWithMergedOptions) => {
					if (response.statusCode === 401) {
						return retryWithMergedOptions({
							headers: {
								token: 'unicorn'
							}
						});
					}

					return response;
				}
			]
		}
	}).json() as any;
	t.is(body.hello, 'world');
});
```

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

PopGoesTheWza:
I opened a PR with the (currently failing) test to reproduce this issue.

szmarczak:
Thank you for the report, I'll get back to this ASAP.

PopGoesTheWza:
I added test cases for `.text()` and `.buffer()` methods for they are also affected.

szmarczak:
Fixed in 99d70df02f209ffac80a6529001e63b8684056af

PR #1121, opened 2020-03-12
test: test case for issue #1120

PopGoesTheWza:
Sample test to reproduce issue #1120 

#### Checklist

### DIFF
--- a/source/as-promise.ts
+++ b/source/as-promise.ts
@@ -7,17 +7,17 @@ import {normalizeArguments, mergeOptions} from './normalize-arguments';
 import requestAsEventEmitter, {proxyEvents} from './request-as-event-emitter';
 import {CancelableRequest, GeneralError, NormalizedOptions, Response} from './types';
 
-const parseBody = (kludge: {body: Buffer}, responseType: NormalizedOptions['responseType'], encoding: NormalizedOptions['encoding']): unknown => {
+const parseBody = (body: Buffer, responseType: NormalizedOptions['responseType'], encoding: NormalizedOptions['encoding']): unknown => {
 	if (responseType === 'json') {
-		return kludge.body.length === 0 ? '' : JSON.parse(kludge.body.toString());
+		return body.length === 0 ? '' : JSON.parse(body.toString());
 	}
 
 	if (responseType === 'buffer') {
-		return Buffer.from(kludge.body);
+		return Buffer.from(body);
 	}
 
 	if (responseType === 'text') {
-		return kludge.body.toString(encoding);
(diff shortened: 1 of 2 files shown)


PopGoesTheWza:
@szmarczak @sindresorhus please comment this (very) kludgy fix.

In short, `.buffer()`, `.json()` and `.text()` shortcuts consume the `body` Buffer bound to the initial `asPromise()` call and cannot "see" the other `body` bound to the `asPromise()` which is called during the `afterResponse` hook.

There is very likely a cleaner way to handle this issue.

source/as-promise.ts
```diff
@@ -7,17 +7,17 @@ import {normalizeArguments, mergeOptions} from './normalize-arguments';
 import requestAsEventEmitter, {proxyEvents} from './request-as-event-emitter';
 import {CancelableRequest, GeneralError, NormalizedOptions, Response} from './types';
 
-const parseBody = (body: Buffer, responseType: NormalizedOptions['responseType'], encoding: NormalizedOptions['encoding']): unknown => {
+const parseBody = (kludge: {body: Buffer}, responseType: NormalizedOptions['responseType'], encoding: NormalizedOptions['encoding']): unknown => {
```
szmarczak: don't do this

PopGoesTheWza: @szmarczak please elaborate. My intent wasn't to commit a proper fix but to hint at what and how this issue occurs.

I never expected this kludge to pollute the codebase and couldn't decide on a better approach.

PopGoesTheWza: Perhaps a better resolution would be to:

- call the `shortcut()` factory function with a reference to the `promise` variable
- then from within `shortcut()` to bind the resulting buffer()/json()/text() function to said `promise` reference
- finally when invoked, buffer()/json()/text() function will `await` its `this` in order to access the `response`, next the `body`

I can give it a try if you think that approach is acceptable.

PopGoesTheWza:
@szmarczak here is a cleaner fix (type casting is somewhat ugly though)

szmarczak:
If you don't mind waiting until I finish up the Rewrite PR, it will be much easier to look into this later...

PopGoesTheWza:
Sure.

sindresorhus:
This can continue now. https://github.com/sindresorhus/got/pull/1051 was merged.

PopGoesTheWza:
Let me now if and how I may assist

szmarczak:
Thanks for the reproducible example, fixed in 99d70df02f209ffac80a6529001e63b8684056af

PR #1153, opened 2020-04-10
Make a fix regarding UNIX domain socket support

gwn:
I have no idea why this is happening, but for some reason the `body` property (which is a ReadableStream) of the `options` object gets lost when the `options` object is extended using the object spread syntax. This doesn't happen when we make use of `Object.assign`.

#### Checklist

### DIFF
--- a/source/normalize-arguments.ts
+++ b/source/normalize-arguments.ts
@@ -485,12 +485,11 @@ export const normalizeRequestArguments = async (options: NormalizedOptions): Pro
 		if (matches?.groups) {
 			const {socketPath, path} = matches.groups;
 
-			options = {
-				...options,
+			Object.assign(options, {
 				socketPath,
 				path,
 				host: ''
-			};
+			});
 		}
 	}
 


gwn:
My node version is `v13.7.0`

sindresorhus:
This will need a test.

gwn:
I'm not adding a feature, I'm just fixing a simple bug. And I'm not breaking any existing tests.

If you are absolutely sure we need a test before this can get passed, please offer some guidance on what kind of test we need. I don't know what and how to test.

sindresorhus:
You're saying you're fixing a bug. I'd like to see a test that would fail in `master` and pass in this PR, to prove that the problem is real and that the fix actually fixes the problem.

gwn:
All right, will do, thank you.

gwn:
Ok, I won't be able to submit a complete test because of my insufficient command of Typescript, lack of a deeper understanding of Got's internals, and lack of time. I gave it a few hours but couldn't make it work.

Basically what I'm trying to do is making sure that the body stream still exists in the options object after normalization.

I'm suggesting adding a second test like the following into the `test/normalize-arguments.ts` (one that actually works :) :

```
diff --git a/test/normalize-arguments.ts b/test/normalize-arguments.ts
index 4f60b12..0352091 100644
--- a/test/normalize-arguments.ts
+++ b/test/normalize-arguments.ts
@@ -1,5 +1,10 @@
 import test from 'ava';
 import got from '../source';
+import {Readable} from 'stream'
+import {
+	normalizeArguments,
+	normalizeRequestArguments
+} from '../source/normalize-arguments.ts';
 
 test('should merge options replacing responseType', t => {
 	const responseType = 'json';
@@ -9,3 +14,16 @@ test('should merge options replacing responseType', t => {
 
 	t.is(options.responseType, responseType);
 });
+
+
+test('should preserve the body stream when connecting to a UNIX socket', t => {
+	const options = {
+		normalizeRequestArguments(normalizeArguments({
+			url: 'unix:/asdf.sock',
+			method: 'POST',
+			json: {foo: 'bar'}
+		}))
+	}
+
+	t.is(options.body.constructor, Readable)
+});

```

Now I can only hope you or one of the maintainers can find the time to complete this.

I also apologize for my tone earlier. Of course you need a test to prove that this is a legit bug. I was kind of in a hurry. I thank you for your explanation.

gwn:
Edit: Fixed a problem in the test code above.

gwn:
Good news for me is that this is no longer a problem after @szmarczak 's rewrite (https://github.com/sindresorhus/got/commit/286e31e0655f4aa98704ed3c9d65554c2354cae7)

szmarczak:
Glad to hear this :) I'll close this then.

PR #1181, opened 2020-04-24
Limit number of requests in pagination

jaulz:
Currently, it's easily possible to trigger an infinite amount of requests during development because we use a `while(true)` construct. This PR adds a new option to limit the number of requests.

#### Checklist

- [X] I have included some tests.
- [X] If it's a new feature, I have included documentation updates.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -765,7 +765,7 @@ The maximum amount of items that should be emitted.
 Type: `number`\
 Default: `10000`
 
-The maximum amount of request that should be triggered.
+The maximum amount of request that should be triggered. [Retries on failure](#retry) are not count towards this limit.
 
 For example, if you need the first 10 pages but don't know how many items there are per page. It can also be helpful during development to avoid an infinite number of requests.
 
--- a/readme.md
+++ b/readme.md
@@ -760,6 +760,13 @@ Default: `Infinity`
 
 The maximum amount of items that should be emitted.
 
+###### pagination.requestLimit
+
+Type: `number`\
+Default: `100`
+
+The maximum amount of request that should be triggered.
+
 ##### localAddress
 
 Type: `string`
--- a/readme.md
+++ b/readme.md
@@ -763,9 +763,11 @@ The maximum amount of items that should be emitted.
 ###### pagination.requestLimit
 
 Type: `number`\
-Default: `100`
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -760,6 +760,13 @@ Default: `Infinity`
 
 The maximum amount of items that should be emitted.
 
+###### pagination.requestLimit
+
+Type: `number`\
+Default: `100`
```
sindresorhus, line 766: Why `100`? If this is just meant to prevent a mistake in development, I think the default should be higher. Maybe just `Infinity`. Then you just set it if you need it.

readme.md
```diff
@@ -760,6 +760,13 @@ Default: `Infinity`
...
+
+The maximum amount of request that should be triggered.
```
sindresorhus, line 768: The text should include the intended use-case for this option.

readme.md
jaulz: Okay, I increased it to 10000. Defaulting to `Infinity` is too much I think as also at runtime in production you could trigger accidentally an infinite amount.

readme.md
jaulz: I just updated the paragraph with an example.

szmarczak: It should be noted that retry requests does not count.

jaulz: Just added that note as well.

readme.md
```diff
@@ -760,6 +760,15 @@ Default: `Infinity`
...
 The maximum amount of items that should be emitted.
 
+###### pagination.requestLimit
+
+Type: `number`\
+Default: `10000`
+
+The maximum amount of request that should be triggered. [Retries on failure](#retry) are not count towards this limit.
+
+For example, if you need the first 10 pages but don't know how many items there are per page. It can also be helpful during development to avoid an infinite number of requests.
```
sindresorhus, line 770: > For example, if you need the first 10 pages but don't know how many items there are per page.

It's not clear how setting a request limit helps with this? It seems like setting a hard-coded limit goes against not knowing how many items there are.

szmarczak: `are not count` -> `are not counted`

jaulz: @szmarczak it's fixed.

Issue #1186, opened 2020-04-24
allowGetBody payload is dropped by .paginate()


PopGoesTheWza:
this is a spinoff of #1171 

 - Node.js version: 12.16.1
 - OS & version: latest MacOQ

since got@11.0.0, `.paginate()` no longer send the body (json payload) of an `allowGetBody` query

#### Actual behavior

the json payload (a filter expression) is not sent to the server which returns the whole set of records

#### Expected behavior

Like in got@10.x the server should only return the records matching the filter (i.e. the json payload)

#### Code to reproduce

I have two simple allowGetBody test cases. In order to run, `test/helpers/with-server.ts` needs to be changed in lines 16-20:

```typescript
	const server = await createTestServer({
		bodyParser: {
			// type: () => false
		}
	}) as ExtendedTestServer;
```

This enable the bodyParser middleware to properly populate `request.body`. Because I do not fathom the possible impact of above change, I won't issue a PR.

First a simple and successful test

```typescript
test('allowGetBody sends json payload', withServer, async (t, server, got) => {
	server.get('/', (request, response) => {
		if (request.body.hello !== 'world') {
			response.statusCode = 400;
		}

		response.end();
	});

	const {statusCode} = await got({
		allowGetBody: true,
		json: {hello: 'world'},
		retry: 0,
		throwHttpErrors: false
	});
	t.is(statusCode, 200);
});
```

Now with .paginate() (it fails)

```typescript
test('allowGetBody sends json payload with .paginate()', withServer, async (t, server, got) => {
	server.get('/', (request, response) => {
		if (request.body.hello !== 'world') {
			response.statusCode = 400;
		}

		response.end(JSON.stringify([1, 2, 3]));
	});

	const iterator = got.paginate({
		allowGetBody: true,
		json: {hello: 'world'},
		retry: 0
	});

	const result = await iterator.next();

	t.is(result.value, 1);
});
```

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

szmarczak:
Good news: I have located the source of the bug

PopGoesTheWza:
About time! ;)

Thanks a lot.

szmarczak:
It was a typo in the code :P

PR #1190, opened 2020-04-26
Add more tests for `options.allowGetBody`

PopGoesTheWza:
To help investigate #1186 this PR adds:

- `withBodyParsingServer` helper to allow the use of `request.body` in test cases
- two test cases for using `allowGetBody` with a json payload

Adding more allowGetBody with non-json payloads might be relevant but not in the current scope of originating issue.

#### Checklist

### DIFF
--- a/test/pagination.ts
+++ b/test/pagination.ts
@@ -1,7 +1,7 @@
 import {URL} from 'url';
 import test from 'ava';
 import got, {Response} from '../source';
-import withServer from './helpers/with-server';
+import withServer, {withBodyParsingServer} from './helpers/with-server';
 import {ExtendedTestServer} from './helpers/types';
 
 const thrower = (): any => {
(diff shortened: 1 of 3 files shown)


test/pagination.ts
```diff
@@ -308,3 +308,23 @@ test('ignores the `resolveBodyOnly` option', withServer, async (t, server, got)
...
+
+	const iterator = got.paginate({
+		allowGetBody: true,
+		json: {hello: 'world'},
+		retry: 0
+	});
+
+	const result = await iterator.next();
+
+	t.is(result.value, 1);
```
szmarczak, line 329:
```suggestion
	t.deepEqual(result.value, [1, 2, 3]);
```

szmarczak: otherwise looks good

PopGoesTheWza: Done. The hack for withBodyParsingServer isn't very nice but I had little time, less inspiration for a better approach.

szmarczak:
Thanks for submitting! :smiley:

PR #1204, opened 2020-04-27
Add `pagination.stackAllItems` option

### LINKED ISSUE #1200: Add an option to disable stacking paginated items

Use `.paginate()` over huge set of data without fear of memory exhaustion.

#### Describe the feature

When using `.paginate()` over a large/huge set of data, the `allItems` parameter will grow and eat up all memory. Personally, I have no use cases where `currentItems` less `allItems` variables are useful.

In addition to the memory usage issue, the aggregation of `allItems` can cause CPU consumption to grow exponentially (at least with got@10.x.)

Hence the request for a goldfish/dumb/forgetful mode for `.paginate()`

I would go as far as suggesting this to be the default mode (but you surely have a better knowledge of the overall user base needs)

One obvious issue I can forsee is of course the signature of `.pagination` callback functions which currently looks like `(response, allItems, currentItems) => {...}` for the second parameter becomes literaly unusable while the third (`currentItems`) is still possibly useful to some.

#### Checklist

PopGoesTheWza:
Fixes #1200 

#### Checklist

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -856,7 +856,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		options.maxRedirects = options.maxRedirects ?? 0;
 
 		// Set non-enumerable properties
-		setNonEnumerableProperties([defaults, options], options);
+		setNonEnumerableProperties([defaults, rawOptions], options);
 
 		return options as NormalizedOptions;
 	}
--- a/test/normalize-arguments.ts
+++ b/test/normalize-arguments.ts
@@ -9,3 +9,14 @@ test('should merge options replacing responseType', t => {
 
 	t.is(options.responseType, responseType);
 });
+
+test('should copy non-numerable properties', t => {
+	const options = {
+		json: {hello: '123'}
+	};
+
+	const merged = got.mergeOptions(got.defaults.options, options);
+	const mergedTwice = got.mergeOptions(got.defaults.options, merged);
+
+	t.is(mergedTwice.json, options.json);
+});
--- a/test/pagination.ts
+++ b/test/pagination.ts
@@ -309,7 +309,7 @@ test('ignores the `resolveBodyOnly` option', withServer, async (t, server, got) 
 	t.deepEqual(items, [1, 2]);
 });
 
-test.failing('allowGetBody sends json payload with .paginate()', withBodyParsingServer, async (t, server, got) => {
+test('allowGetBody sends json payload with .paginate()', withBodyParsingServer, async (t, server, got) => {
 	server.get('/', (request, response) => {
 		if (request.body.hello !== 'world') {
 			response.statusCode = 400;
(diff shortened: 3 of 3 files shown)


source/create.ts
```diff
@@ -224,10 +225,13 @@ const create = (defaults: InstanceDefaults): Got => {
...
 
-					all.push(item as T);
+					if (pagination.stackAllItems) {
+						all.push(item as T);
+					}
+
 					current.push(item as T);
 
-					if (all.length === pagination.countLimit) {
+					if ((countLimit -= 1) === 0) {
```
szmarczak, line 234:
```js
(--countLimit) === 0
```

It's a bit quicker ;)

PopGoesTheWza:
Awaiting comments/feedback to proceed further.

Designing tests is not my strong point so guidelines/directions would be appreciated.

PopGoesTheWza:
Perhaps there should be a better name than stackAllItems? This one does appeal to me…

szmarczak:
> Perhaps there should be a better name than stackAllItems? This one does appeal to me…

Seems good to me. Sorry I haven't looked at this PR yet, I'm busy with getting `http2-wrapper` to work on Node.js 14. Actually I'm almost finished.

szmarczak:
Basically you need to write just two test:

```ts
test('`stackAllItems` set to false', withServer, async (t, server, got) => {
	attachHandler(server, 3);

	const result = await got.paginate.all<number>({
		stackAllItems: false,
		pagination: {
			filter: (element: number, allItems: number[], currentItems: number[]) => {
				// You need to check this on every method that has `allItems`
				t.is(allItems.length, 0);

				return true;
			},
			...
		}
	});

	t.deepEqual(result, [1, 2, 3]);
});
```

and the other one would be `stackAllItems set to true`.

szmarczak:
> is there already a test which check Got's default values?

No. We test every option. It's pointless to check defaults.

PopGoesTheWza:
> It's pointless to check defaults.

When a default options changes in the master, it could be considered a breaking change. Testing defaults can act as a canary if such changes happens but isn't properly tracked and documented... Just my two cents ;)

PopGoesTheWza:
Oops. I am unsure how I made the two merge above and if it did not polute anything.

PopGoesTheWza:
@szmarczak tests and documentation added

szmarczak:
```
$ git cherry-pick c3d5a5f51ec9eb8fa16189869478874a9d0e6c43
Segmentation fault
```

I don't know how but your merge commits broke my git :((((

even force push failed for some reason

szmarczak:
```
fatal: Unable to create '/home/szm/Desktop/got/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```

wtf I didn't do anything that would lock the file, just git checkout

szmarczak:
Don't know how I fixed it but I fixed it somehow: #1214

szmarczak:
btw I did not close this PR, it was automatically closed by the force push. I tried to push again and got `permission denied` because it has been closed.

szmarczak:
IMO GitHub shouldn't do that

PR #1208, opened 2020-04-30
Add a test for stringified `searchParams` in merge

jaulz:
While using the pagination API I see that the `searchParams` option is a string and thus it fails at this line.

#### Checklist

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -674,9 +674,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			options.searchParams = new URLSearchParams(options.searchParams as Record<string, string>);
 
 			// `normalizeArguments()` is also used to merge options
-			const defaultSearchParams = defaults?.searchParams;
-			if (defaultSearchParams && defaultSearchParams instanceof URLSearchParams) {
-				defaultSearchParams.forEach((value, key) => {
+			const defaultSearchParameters = defaults?.searchParams;
+			if (defaultSearchParameters && defaultSearchParameters instanceof URLSearchParams) {
+				defaultSearchParameters.forEach((value, key) => {
 					(options!.searchParams as URLSearchParams).append(key, value);
 				});
 			}
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -674,8 +674,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			options.searchParams = new URLSearchParams(options.searchParams as Record<string, string>);
 
 			// `normalizeArguments()` is also used to merge options
-			if (defaults?.searchParams? instanceof URLSearchParams) {
-				defaults.searchParams.forEach((value, key) => {
+			const defaultSearchParams = defaults?.searchParams;
+			if (defaultSearchParams && defaultSearchParams instanceof URLSearchParams) {
+				defaultSearchParams.forEach((value, key) => {
 					(options!.searchParams as URLSearchParams).append(key, value);
 				});
 			}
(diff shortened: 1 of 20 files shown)


test/normalize-arguments.ts
```diff
@@ -9,3 +10,16 @@ test('should merge options replacing responseType', t => {
 
 	t.is(options.responseType, responseType);
 });
+
+test('should merge searchParams', t => {
```
szmarczak, line 14: This test passes without your changes

szmarczak: Duplicate: merges `searchParams` instances in arguments.ts

szmarczak:
Fixed in 157e02bcb613ab506829326e7af96e8155b4e186

jaulz:
Great, thanks!

szmarczak:
Actually I'm wrong. e23b7f9b975cd1d70254847b5348685af315163b

szmarczak:
You just need to revert your commit and `git pull origin master --rebase`. Then you can start working on the fix :)

jaulz:
@szmarczak I just pushed another change which 1) handles `searchParams` that are passed `string`, `undefined` or instances of `URLSearchParams` via defaults and 2) prevents duplications of search params. Though, I'm not sure if the second will cause regressions because it changes the behaviour how `searchParams` are merged. However, in the context of paging there would be no other way to force setting only one key in the URL as it would result in a URL like this: `?page=1&page=2&page=3&page=4...`

source/core/index.ts
```diff
@@ -647,9 +647,14 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
-					(options!.searchParams as URLSearchParams).append(key, value);
-				});
+				if (defaults?.searchParams) {
+					new URLSearchParams(defaults.searchParams).forEach((value, key) => {
+						const searchParameters = options!.searchParams as URLSearchParams;
+						if (!searchParameters.get(key)) {
+							searchParameters.append(key, value);
+						}
+					});
+				}
```
szmarczak, line 657: This is exactly the same as:

```ts
				defaults?.searchParams?.forEach((value, key) => {
					(options!.searchParams as URLSearchParams).set(key, value);
				});
```

szmarczak:
I don't think you fixed 1) because as mentioned before it worked even without your changes. You need to provide another test if you think otherwise ;)

source/core/index.ts
jaulz: Not exactly. If you pass a `string` via the init hook then it will dump for example.

test/normalize-arguments.ts
```diff
@@ -11,6 +11,21 @@ test('should merge options replacing responseType', t => {
 	t.is(options.responseType, responseType);
 });
 
+test('should merge searchParams', t => {
```
szmarczak, line 14:
```suggestion
test('no duplicated searchParams values', t => {
```

source/core/index.ts
szmarczak: You would need to change the docs then

szmarczak: > Not exactly. If you pass a string via the init hook then it will dump for example.

A test would be awesome :)

jaulz: Sure... though, is there any magic about the tests? I just noticed that I can even use `t.is(true, false);` and the test does not fail locally when using `yarn test test/hooks.ts`.

jaulz: And is there any way to output debug information during tests?

szmarczak:
This PR partially is covered by #1229 Fixed the merge conflict. Thanks! :raised_hands:

PR #1212, opened 2020-04-30
Allow response body to be typed for pagination API

fiznool:
When using got with TypeScript, this PR allows the `Response` generic body type to be set when using the pagination API, meaning that the `transform()` and `paginate()` functions will work without any typecasting necessary.

For example we can now do:

``` ts
interface Account {
  id: string;
  name: string;
}

interface ResponseBody {
  data: Account[];
}

const accounts = await got.paginate.all<Account, ResponseBody>(url, {
  pagination: {
    transform: response => response.body.data  // No typecasting necessary: response.body is typed as ResponseBody
  }
});
```

#### Checklist

### DIFF
--- a/source/as-promise/types.ts
+++ b/source/as-promise/types.ts
@@ -70,18 +70,18 @@ export interface Hooks extends RequestHooks {
 	afterResponse?: AfterResponseHook[];
 }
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
Very nice!

PR #1214, opened 2020-05-02
Add `pagination.stackAllItems` option

### LINKED ISSUE #1200: Add an option to disable stacking paginated items

szmarczak:
Moved from #1204
Fixes #1200

Had to open this PR because git was complaining that the @PopGoesTheWza branch was in middle of merging even though it clearly was not. Dunno what happend.

/cc @PopGoesTheWza

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -791,6 +791,15 @@ The maximum amount of request that should be triggered. [Retries on failure](#re
 
 For example, it can be helpful during development to avoid an infinite number of requests.
 
+###### pagination.stackAllItems
+
+Type: `boolean`\
+Default: `true`
+
+Defines how the parameter `allItems` in [pagination.paginate](#pagination.paginate), [pagination.filter](#pagination.filter) and [pagination.shouldContinue](#pagination.shouldContinue) is managed. When set to `false`, the parameter `allItems` is always an empty array.
+
+This option can be helpful to save on memory usage when working with a large dataset.
+
 ##### localAddress
 
 Type: `string`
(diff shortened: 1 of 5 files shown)


PopGoesTheWza:
@szmarczak the fuss with previous PR is likely due to me wrongfully pull from `upstream/master` two days ago

PopGoesTheWza:
@sindresorhus @szmarczak I would humbly suggest the following 2 breaking changes:

- Switch order of `allItems` and `currentItems` parameters

My reasoning about parameters order is that (hopefully/probably) most use cases don't need the `allItems` parameters, with `currentItems` typically being more used. As per the default, the computation of `allItems` when unused implies a waste of CPU and RAM resources (and a crash in worst case scenarios)

Anyhow, let me know if/how I can assist more on this PR.

Regards,

sindresorhus:
We cannot do breaking changes for a while, but open a new issue and we’ll consider it for the next major release.

PopGoesTheWza:
Sure. I will open an enhancement request after `stackAllItems` option is released.

szmarczak:
> Switch order of allItems and currentItems parameters

I don't think there's a need to. It doesn't change how the code works.

> Default value for stackAllItem to false

I don't have a particular opinion on this one. I'm good with either.

PR #1243, opened 2020-05-10
Prevent URL pollution

### LINKED ISSUE #1235: got 11.x strips username and password from WHATWG URLs
#### Describe the bug

 - Node.js version: 12.16, 14.2.0
 - OS & version: Debian 10

The code in question is here: https://github.com/sindresorhus/got/blob/50ef99a9856ccd37a75b42f94d023a491728a92d/source/core/index.ts#L656-L658

#### Actual behavior

Got sends an unauthenticated request when credentials are specified within the URL.

#### Expected behavior

`normalizeArguments` should use the `username` and `password` options passed in the `url` parameter given that it's a WHATWG URL and that they are set. This was supported in got 10.x.

#### Code to reproduce

```js
const url = new URL('http://user:pass@localhost:3000');
console.log(url);
got.post(url, {});
```

#### Checklist

Giotino:
#### Checklist

Keeping in mind that
> Properties from options will override properties in the parsed url.

URL's `username` ans `password` are overwritten only if they're not empty (or null) in options.

I'm going to add some tests.

Fixes #1235

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - '14'
   - '13'
   - '12'
   - '10'
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "got",
-	"version": "11.1.1",
+	"version": "11.1.2",
 	"description": "Human-friendly and powerful HTTP request library for Node.js",
 	"license": "MIT",
 	"repository": "sindresorhus/got",
--- a/readme.md
+++ b/readme.md
@@ -921,7 +921,7 @@ Returns a [duplex stream](https://nodejs.org/api/stream.html#stream_class_stream
 
 ```js
 got.stream('https://github.com')
-	.on('request', request => setTimeout(() => request.abort(), 50));
+	.on('request', request => setTimeout(() => request.destroy(), 50));
 ```
 
 ##### .on('response', response)
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -727,8 +727,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			}
 
 			// Update `username` & `password`
-			options.url.username = options.username;
-			options.url.password = options.password;
+			if (options.username !== '')
+				options.url.username = options.username;
+			if (options.password !== '')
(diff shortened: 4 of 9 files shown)


Giotino:
Added some tests

Giotino:
There's no documentation about if `username` and `password` should be replaced (or merged) together or separately.

There's a test `response contains got options` which fails when the `username` and `password` are not replaced together.
The problem is that the URL (only if an object, not a string) from the given `ExtendedGot` is modified by each call and, as it's an object (reference), it's the same object for each call.

Is this an expected behavior? If not the URL should be deep-copied on each call to prevent pollution.

https://github.com/sindresorhus/got/blob/master/source/core/index.ts#L583

Is only a 1 level deep copy, url is copied as a reference that is not enough.

My suggestion is to add
```
if (is.urlInstance(url)) {
  options.url = {...url} as any;
} else {
  options.url = url;
}
```
after line 587

szmarczak:
> There's no documentation about if username and password should be replaced (or merged) together or separately.

Seperately. E.g. a user can have an empty password.

szmarczak:
> If not the URL should be deep-copied on each call to prevent pollution.

Not exactly object copy, but `new URL(url)`.

szmarczak:
This also needs a separate test.

Giotino:
> > If not the URL should be deep-copied on each call to prevent pollution.
> 
> Not exactly object copy, but `new URL(url)`.

Yes, I'm doing it

Giotino:
Check this last commit

szmarczak:
Hint: `git pull origin master --rebase`. So you don't need to commit the `Merge branch 'master' into ...`.

szmarczak:
This just needs a separate test to check against the pollution and we're good to go! :)

Giotino:
> Hint: `git pull origin master --rebase`. So you don't need to commit the `Merge branch 'master' into ...`.

Let's say i need to get used to git... :)

szmarczak:
> Let's say I need to get used to git... :)

I've been there too ;)

> BTW, I added a test, it's like the one which was failing, but simpler

Let me see it.

szmarczak:
Good job! Thanks :raised_hands:

PR #1244, opened 2020-05-10
Clarify docs for got.HTTPError

### LINKED ISSUE #1213: Note that 304 doesn't throw `HTTPError`

The PR #252 updated `got` not to throw `HTTPErrors`. The ReadMe was updated that in case of 304 response status code the body would be empty. However the documentation on `HTTPError` still says that they are thrown for non-20x codes. 

I read the documentation on `HTTPError` and expected an exception for 304. It didn’t work this way. Would it make sense to update this part?

#### Checklist

chinesedfan:
#### Checklist

Closed #1213. The preview is as following, and note that the anchor is a relative link so it doesn't work here but doesn't matter in fact.

----

Determines if a [`got.HTTPError`](#gothttperror) is thrown for error responses.

#### got.ParseError

> *ok*: 2xx or 304, and includes other 3xx if `options.followRedirect` is set

When server response code is ***ok***, and parsing body fails. Includes a `response` property.

#### got.HTTPError

When the server response code is not ***ok***. Includes a `response` property.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1298,13 +1298,11 @@ When reading from response stream fails.
 
 #### got.ParseError
 
-> *ok*: 2xx or 304, and includes other 3xx if `options.followRedirect` is set
-
-When server response code is ***ok***, and parsing body fails. Includes a `response` property.
+When server response code is 2xx, and parsing body fails. Includes a `response` property.
 
 #### got.HTTPError
 
-When the server response code is not ***ok***. Includes a `response` property.
+When the server response code is not 2xx nor 3xx if `options.followRedirect` is `true`, but always except for 304. Includes a `response` property.
 
 #### got.MaxRedirectsError
 
--- a/readme.md
+++ b/readme.md
@@ -523,7 +523,7 @@ const got = require('got');
 Type: `boolean`\
 Default: `true`
 
-Determines if a [`got.HTTPError`](#gothttperror) is thrown for error responses.
+Determines if a [`got.HTTPError`](#gothttperror) is thrown for unsuccessful responses.
 
 If this is disabled, requests that encounter an error status code will be resolved with the `response` instead of throwing. This may be useful if you are checking for resource availability and are expecting error responses.
 
--- a/readme.md
+++ b/readme.md
@@ -523,7 +523,7 @@ const got = require('got');
 Type: `boolean`\
 Default: `true`
 
-Determines if a `got.HTTPError` is thrown for error responses (non-2xx status codes).
+Determines if a [`got.HTTPError`](#gothttperror) is thrown for error responses.
 
 If this is disabled, requests that encounter an error status code will be resolved with the `response` instead of throwing. This may be useful if you are checking for resource availability and are expecting error responses.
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -523,7 +523,7 @@ const got = require('got');
 Type: `boolean`\
 Default: `true`
 
-Determines if a `got.HTTPError` is thrown for error responses (non-2xx status codes).
+Determines if a [`got.HTTPError`](#gothttperror) is thrown for error responses.
```
szmarczak, line 526:
```suggestion
Determines if a [`got.HTTPError`](#gothttperror) is thrown for unsuccessful responses.
```

readme.md
```diff
@@ -1298,11 +1298,13 @@ When reading from response stream fails.
 
 #### got.ParseError
 
-When server response code is 2xx, and parsing body fails. Includes a `response` property.
+> *ok*: 2xx or 304, and includes other 3xx if `options.followRedirect` is set
+
+When server response code is ***ok***, and parsing body fails. Includes a `response` property.
```
szmarczak, line 1303: This actually does not affect `ParseError`.

readme.md
szmarczak, line 526:
```suggestion
Determines if a [`got.HTTPError`](#gothttperror) is thrown for unsuccessful responses.
```

readme.md
```diff
@@ -1298,11 +1298,13 @@ When reading from response stream fails.
...
 
 #### got.HTTPError
 
-When the server response code is not 2xx. Includes a `response` property.
+When the server response code is not ***ok***. Includes a `response` property.
```
szmarczak, line 1307: You need to replace `not ok` with a more descriptive information.

readme.md
chinesedfan: @szmarczak Not really. See source/as-promise/index.ts.

```
// Parse body
try {
    response.body = parseBody(response, options.responseType, options.encoding);
} catch (error) {
    // Fallback to `utf8`
    response.body = rawBody.toString();

    if (isOk()) {
        // TODO: Call `request._beforeError`, see https://github.com/nodejs/node/issues/32995
        reject(error);
        return;
    }
}
```

readme.md
chinesedfan: Same with above. Looks like a duplicated suggestion.

readme.md
chinesedfan: I want to make `ok` as a term, as well as used in `ParseError`.

readme.md
szmarczak: It doesn't have any effect. This code is **not** executed on redirects. `304` are empty responses, so even with `responseType: 'json'` it never throws:

https://github.com/sindresorhus/got/blob/ece94edec31bc8d004f93d74e218d8fd3b442c15/source/as-promise/core.ts#L28-L30

There is no `response` event emitted on redirect:

https://github.com/sindresorhus/got/blob/ece94edec31bc8d004f93d74e218d8fd3b442c15/source/core/index.ts#L1032-L1114

readme.md
szmarczak: Keep things simple. Don't redefine things. Simply:

> When the server response code is not 2xx and 3xx if `options.followRedirect` is `true`.

szmarczak:
Looks good. Thanks for submitting! 🙌

PR #1246, opened 2020-05-11
Don't force query string normalization

### LINKED ISSUE #1234: Don't force query string normalization

In an url like `http://example.org/random?param=SOMETHING~SOMETHING` the special character `~` is percent-encoded before the request, resulting in `http://example.org/random?param=SOMETHING%7ESOMETHING` which is not supported (decoded) by some HTTP servers.

As described by [RFC 3986 in section 2.3](https://tools.ietf.org/html/rfc3986#section-2.3)
"
URI comparison implementations do not always perform normalization prior to comparison. For consistency, percent-encoded octets in the ranges of ALPHA (%41-%5A and %61-%7A), DIGIT (%30-%39), hyphen (%2D), period (%2E), underscore (%5F), or tilde (%7E) should not be created by URI producers and, when found in a URI, should be decoded to their corresponding unreserved characters by URI normalizers.
"

Also in [RFC 3986, section 6.2.2.2](https://tools.ietf.org/html/rfc3986#section-6.2.2.2)
"
The percent-encoding mechanism is a frequent source of variance among otherwise identical URIs. In addition to the case normalization issue noted above, some URI producers percent-encode octets that do not require percent-encoding, resulting in URIs that are equivalent to their non-encoded counterparts. These URIs should be normalized by decoding any percent-encoded octet that corresponds to an unreserved character, as described in Section 2.3.
"

The percent-encoding of `~` by got happen because NodeJS follows the "WHATWG URL API" (https://nodejs.org/api/url.html#url_the_whatwg_url_api) which misses `~` from the unreserved characters (https://url.spec.whatwg.org/#interface-urlsearchparams, the Note below the example, and https://url.spec.whatwg.org/#urlencoded-serializing) and, by the way, includes `*`.

#### Describe the feature

My proposal is to add a flag to the options to prevent the normalization by skipping the append and delete of "_GOT_INTERNAL_TRIGGER_NORMALIZATION"

#### Checklist

szmarczak:
It assumes that the query speicified in the input is not `URLSearchParams` unless the `searchParams` option is present.

~The search params are correctly normalized via `decodeURIComponent` (I spent 30 mins reading stackoverflow and exeprimenting by myself, I'm pretty sure that's the correct one).~ The WHATWG URL handles it, what I need is just `decodeURI` before parsing the `location` header.

Fixes #1234

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -139,6 +139,13 @@ Properties from `options` will override properties in the parsed `url`.
 
 If no protocol is specified, it will throw a `TypeError`.
 
+**Note:** The query string is **not** parsed as search params. Example:
+
+```
+got('https://example.com/?query=a b'); //=> https://example.com/?query=a%20b
+got('https://example.com/', {searchParams: {query: 'a b'}}); //=> https://example.com/?query=a+b
+```
+
 ##### options
 
 Type: `object`
--- a/readme.md
+++ b/readme.md
@@ -144,6 +144,9 @@ If no protocol is specified, it will throw a `TypeError`.
 ```
 got('https://example.com/?query=a b'); //=> https://example.com/?query=a%20b
 got('https://example.com/', {searchParams: {query: 'a b'}}); //=> https://example.com/?query=a+b
+
+// The query string is overridden by `searchParams`
+got('https://example.com/?query=a b', {searchParams: {query: 'a b'}}); //=> https://example.com/?query=a+b
 ```
 
 ##### options
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -726,11 +726,6 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				options.url.search = options.searchParams.toString();
 			}
 
-			// Trigger search params normalization
-			if (options.url.search) {
-				options.url.search = decodeURIComponent(options.url.search.toString());
-			}
-
 			// Protocol check
 			if (protocol !== 'http:' && protocol !== 'https:') {
 				throw new UnsupportedProtocolError(options as NormalizedOptions);
@@ -1069,9 +1064,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				const redirectBuffer = Buffer.from(response.headers.location, 'binary').toString();
 
 				// Handles invalid URLs. See https://github.com/sindresorhus/got/issues/604
-				const redirectUrl = new URL(redirectBuffer, url);
+				const redirectUrl = new URL(decodeURI(redirectBuffer), url);
 				const redirectString = redirectUrl.toString();
-				decodeURI(redirectString);
 
 				// Redirecting to a different site, clear sensitive data.
 				if (redirectUrl.hostname !== url.hostname) {
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1067,7 +1067,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				const redirectUrl = new URL(redirectBuffer, url);
 				const redirectString = redirectUrl.toString();
 				decodeURI(redirectString);
-				
+
 				// Redirecting to a different site, clear sensitive data.
 				if (redirectUrl.hostname !== url.hostname) {
 					if ('host' in options.headers) {
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -728,10 +728,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 			// Trigger search params normalization
 			if (options.url.search) {
-				const triggerSearchParameters = '_GOT_INTERNAL_TRIGGER_NORMALIZATION';
-
-				options.url.searchParams.append(triggerSearchParameters, '');
-				options.url.searchParams.delete(triggerSearchParameters);
+				options.url.search = decodeURIComponent(options.url.search.toString());
 			}
 
 			// Protocol check
@@ -1068,8 +1065,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			}
 
 			try {
-				// Handles invalid URLs. See https://github.com/sindresorhus/got/issues/604
+				// Do not remove. See https://github.com/sindresorhus/got/pull/214
 				const redirectBuffer = Buffer.from(response.headers.location, 'binary').toString();
+
+				// Handles invalid URLs. See https://github.com/sindresorhus/got/issues/604
 				const redirectUrl = new URL(redirectBuffer, url);
 				const redirectString = redirectUrl.toString();
 				decodeURI(redirectString);
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1064,9 +1064,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				const redirectBuffer = Buffer.from(response.headers.location, 'binary').toString();
 
 				// Handles invalid URLs. See https://github.com/sindresorhus/got/issues/604
-				const redirectUrl = new URL(decodeURI(redirectBuffer), url);
+				const redirectUrl = new URL(redirectBuffer, url);
 				const redirectString = redirectUrl.toString();
-
+				decodeURI(redirectString);
+				
 				// Redirecting to a different site, clear sensitive data.
 				if (redirectUrl.hostname !== url.hostname) {
 					if ('host' in options.headers) {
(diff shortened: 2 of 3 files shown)


source/core/index.ts
```diff
@@ -728,10 +728,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 			// Trigger search params normalization
 			if (options.url.search) {
-				const triggerSearchParameters = '_GOT_INTERNAL_TRIGGER_NORMALIZATION';
-
-				options.url.searchParams.append(triggerSearchParameters, '');
-				options.url.searchParams.delete(triggerSearchParameters);
+				options.url.search = decodeURIComponent(options.url.search.toString());
```
sindresorhus, line 731:
```suggestion
				options.url.search = decodeURIComponent(options.url.search);
```

`.search` is already a string.

test/arguments.ts
```diff
@@ -55,12 +55,12 @@ test('throws an error if the protocol is not specified', async t => {
 	});
 });
 
-test('string url with searchParams is preserved', withServer, async (t, server, got) => {
+test('properly encodes search params', withServer, async (t, server, got) => {
```
szmarczak, line 58:
```suggestion
test('properly encodes query string', withServer, async (t, server, got) => {
```

Giotino:
I would suggest to simply remove the if block (`if (options.url.search)`) which you're modifying.

There's nothing wrong about using `decodeURIComponent()`, this behavior is even suggested by the [RFC 3986 in section 2.3](https://tools.ietf.org/html/rfc3986#section-2.3), but I think that it can cause problems with some server.

As the RFC use the word "should" I think you should consider to not apply that policy in order to increase compatibility.

source/core/index.ts
```diff
@@ -728,10 +728,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 			// Trigger search params normalization
 			if (options.url.search) {
-				const triggerSearchParameters = '_GOT_INTERNAL_TRIGGER_NORMALIZATION';
-
-				options.url.searchParams.append(triggerSearchParameters, '');
-				options.url.searchParams.delete(triggerSearchParameters);
+				options.url.search = decodeURIComponent(options.url.search);
 			}
```
szmarczak, line 732:
```suggestion
			options.url.search = decodeURIComponent(options.url.search);
```

Giotino:
I don't want to be pedantic (you probably hate me now), but this is going to break `x-www-form-urlencoded`

Example: `http://127.0.0.1:8123/?FIRST%3DSECOND=THIRD`
Correct: `{ 'FIRST=SECOND': 'THIRD' }`
This PR: `{ FIRST: 'SECOND=THIRD' }`

Giotino:
https://github.com/Giotino/got/tree/issue-1234

This is what I'm proposing

szmarczak:
> became: http://example.org/?a=b

No, this is a breaking change and we won't release Got 12 in the nearest future. Got 11 is buggy along with Node.js 14...

szmarczak:
Actually I misunderstood you. This is working as expected. If you provide `searchParams`, it overrides the input. [It's in the docs.](https://github.com/sindresorhus/got/blob/master/readme.md#searchparams)

szmarczak:
https://github.com/sindresorhus/got/pull/1246#issuecomment-626846622 Well, we gotta sacrifice something! This is an edge case anyway... :P

source/core/index.ts
```diff
@@ -726,13 +726,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
 
-			// Trigger search params normalization
-			if (options.url.search) {
-				const triggerSearchParameters = '_GOT_INTERNAL_TRIGGER_NORMALIZATION';
-
-				options.url.searchParams.append(triggerSearchParameters, '');
-				options.url.searchParams.delete(triggerSearchParameters);
-			}
+			// Properly normalize query string
+			options.url.search = decodeURIComponent(options.url.search);
```
szmarczak, line 730: Maybe this should be put in the `if` above?

Giotino:
> [#1246 (comment)](https://github.com/sindresorhus/got/pull/1246#issuecomment-626846622) Well, we gotta sacrifice something! This is an edge case anyway... :P

Removing `decodeURIComponent` does not cause any side effect and does not break any format.

Giotino:
https://github.com/sindresorhus/got/pull/1246/commits/4724a0d091ed2af05c6693a76bff423ff53f4106#diff-8eef38c70c64f535c3b8046b5ce40d9eR1067

Using `decodeURI` there give problems.
Example:

`Location: http://example.org/?something%25` (valid)

`http://example.org/?something%25` => `http://example.org/?something%`
which is invalid, error `RequestError: URI malformed`

Giotino:
> error `RequestError: URI malformed`

That maybe wrong, but the behavior is even stranger.
What's happening on my end is:
First request to: `http://example.org/a_url_that_answer_with_redirect`
New request following the redirect to: `http://example.org/?something%25`
If this request fails (eg. socket hang up) a new request is made, but it fails with `RequestError: URI malformed`

szmarczak:
@sindresorhus So the solution seems to be just remove the stress on the `search` normalization. I think it's good to merge now.

Giotino:
The last thing that's missing is to remove `decodeURI` from here
https://github.com/sindresorhus/got/commit/4724a0d091ed2af05c6693a76bff423ff53f4106#diff-8eef38c70c64f535c3b8046b5ce40d9eR1067

szmarczak:
> The last thing that's missing is to remove decodeURI from here

The server may send incorrect URL. If it's already correct, then `decodeURI` won't break it. But I don't say I'm right. Feel free to prove me wrong :)

readme.md
```diff
@@ -139,6 +139,13 @@ Properties from `options` will override properties in the parsed `url`.
 
 If no protocol is specified, it will throw a `TypeError`.
 
+**Note:** The query string is **not** parsed as search params. Example:
+
+```
+got('https://example.com/?query=a b'); //=> https://example.com/?query=a%20b
+got('https://example.com/', {searchParams: {query: 'a b'}}); //=> https://example.com/?query=a+b
```
sindresorhus, line 146: This is not clear enough. It should also show an example with both `?query` and `{searchParams..`

Giotino:
> > The last thing that's missing is to remove decodeURI from here
> 
> The server may send incorrect URL. If it's already correct, then `decodeURI` won't break it. But I don't say I'm right. Feel free to prove me wrong :)

Let's say I request a page that answer with `Location: http://example.org/?some%25thing` header (the URL is valid).
Got has to follow the redirect using that url (`http://example.org/?something%25`).
The expected behavior is to request this page `http://example.org/?some%25thing`, but with `decodeURI` it request `http://example.org/?some%thing` which is wrong (and, in this case, invalid).

readme.md
szmarczak: Fixed

szmarczak:
https://tools.ietf.org/html/rfc3986#section-2.4

>    Under normal circumstances, the only time when octets within a URI
>    are percent-encoded is during the process of producing the URI from
>    its component parts.  This is when an implementation determines which
>    of the reserved characters are to be used as subcomponent delimiters
>    and which can be safely used as data.  Once produced, a URI is always
>    in its percent-encoded form.
> 
>    When a URI is dereferenced, the components and subcomponents
>    significant to the scheme-specific dereferencing process (if any)
>    must be parsed and separated before the percent-encoded octets within
>    those components can be safely decoded, as otherwise the data may be
>    mistaken for component delimiters.  The only exception is for
>    percent-encoded octets corresponding to characters in the unreserved
>    set, which can be decoded at any time.  For example, the octet
>    corresponding to the tilde ("\~") character is often encoded as "%7E"
>    by older URI processing implementations; the "%7E" can be replaced by
>    "~" without changing its interpretation.
> 
>    Because the percent ("%") character serves as the indicator for
>    percent-encoded octets, it must be percent-encoded as "%25" for that
>    octet to be used as data within a URI.  Implementations must not
>    percent-encode or decode the same string more than once, as decoding
>    an already decoded string might lead to misinterpreting a percent
>    data octet as the beginning of a percent-encoding, or vice versa in
>    the case of percent-encoding an already percent-encoded string.

szmarczak:
image

Giotino:
While `~` is an edge case (that I suggest to leave as it is for compatibility), `%` is not.
Try to do the same thing you've done but with `%25`.

szmarczak:
Welp. My example is as valid as yours. But `decodeURI` is indeed potentially breaking. Reverting.

PR #1248, opened 2020-05-11
Do not destroy completed requests

### LINKED ISSUE #1245: Socket hang up with parallel requests and maxSockets agent options with node >= 14.1.0
#### Describe the bug

 - Node.js version: 14.1.0 and 14.2.0
 - OS & version: Ubuntu 19
 - Got: 11.1.3

Starting with node 14.1.0 (not 14.0.0), there's socket hang ups when doing parallel requests and limiting the sockets number with the `maxSockets` option from Agent class. If it could be a clue, I was incapable to reproduce this bug with an independent test including an http server on my localhost. Node's http/https libs are not affected but it's clear there's something related to a node change as the following example runs fine with node 14.0.0 but not above.

#### Code to reproduce

```js
'use strict';
const https = require('https');
const got = require('got');
(async () => {
	try {
		const promises = [];
		const options = {
			url: 'https://github.com',
			agent: {
				https: new https.Agent({ maxSockets: 1 })
			},
			retry: 0
		};

		for (let i = 0; i < 5; i++) {
			promises.push(
				got(options).then(() => {
					console.log('end');
				})
			);
		}

		await Promise.all(promises);
		console.log('all done');
	} catch (e) {
		console.log(e);
	}
})();
```

#### Node's native functional code
```js
'use strict';
const https = require('https');
(async () => {
	try {
		const promises = [];
		const options = {
			hostname: 'github.com',
			agent: new https.Agent({ maxSockets: 1, })
		};

		for (let i = 0; i < 5; i++) {
			promises.push(new Promise((resolve, reject) => {
				const req = https.request(options, (res) => {
					res.on('data', () => {});
					res.on('end', () => {
						console.log('end')
						resolve();
					});
					res.on('error', reject);
					req.on('error', reject);
				});
				req.end();
			}));

		}

		await Promise.all(promises);
		console.log('all done');
	} catch (e) {
		console.log(e);
	}
})();
```

#### Checklist

szmarczak:
Fixes #1245

#### Checklist

### DIFF
--- a/source/core/utils/decompress-response.ts
+++ b/source/core/utils/decompress-response.ts
@@ -1,0 +1,94 @@
+'use strict';
+import {IncomingMessage} from 'http';
+import {Transform, PassThrough} from 'stream';
+const zlib = require('zlib');
+
+const knownProperties = [
+	'aborted',
+	'complete',
+	'destroy',
+	'headers',
+	'httpVersion',
+	'httpVersionMinor',
+	'httpVersionMajor',
+	'method',
+	'rawHeaders',
+	'rawTrailers',
+	'setTimeout',
+	'socket',
+	'statusCode',
+	'statusMessage',
+	'trailers',
+	'url'
+];
+
+const decompressResponse = (response: IncomingMessage): IncomingMessage => {
+	const contentEncoding = (response.headers['content-encoding'] || '').toLowerCase();
+
+	if (!['gzip', 'deflate', 'br'].includes(contentEncoding)) {
+		return response;
+	}
+
+	// TODO: Remove this when targeting Node.js 12.
+	const isBrotli = contentEncoding === 'br';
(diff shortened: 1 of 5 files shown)


source/core/utils/decompress-response.ts
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// TODO: Remove this when targeting Node.js 12.
+	const isBrotli = contentEncoding === 'br';
+	if (isBrotli && typeof zlib.createBrotliDecompress !== 'function') {
+		response.destroy(new Error('Brotli is not supported on Node.js < 12'));
+		return response;
+	}
+
+	let empty = true;
```
sindresorhus, line 39:
```suggestion
	let isEmpty = true;
```

sindresorhus:
Is the plan to upstream these changes? If so, maybe add a TODO comment.

Issue #1072, opened 2020-02-17
The RunKit example is outdated


Eomm:
#### Describe the bug

The runkit example is not updated for got v10:
https://npm.runkit.com/got

#### Actual behavior

```
(await got(ISS, { json: true })).body.iss_position;
```

> TypeError: The `GET` method cannot be used with a body

The code should be updated with:

```
(await got(ISS, { responseType: 'json' })).body.iss_position;
```

I didn't find where this example is stored in this repo, I suppose it is not.

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

sindresorhus:
We don't have any control over that example as it's made by npm, but it seems like we can override it: https://blog.runkit.com/2015/10/28/npm-plus-runkit/

szmarczak:
Fixed in #1253 48bbb369d715b8ec3c36e0af1ec88f47e2299172

szmarczak:
Unfortunately it still doesn't have any effect:

![image](https://user-images.githubusercontent.com/36894700/82125106-e96e3a00-97a3-11ea-86b8-a2ffce208910.png)

Giotino:
Maybe it has to be in `files`...

szmarczak:
Let's wait a while a let's see. Maybe it should be moved to root directory, idk.

PR #1253, opened 2020-05-12
Add a Runkit example

Giotino:
#### Checklist

Similar to the already existing example
```javascript
const got = require("got");
const ISS = "http://api.open-notify.org/iss-now.json";

(await got(ISS, { json: true })).body.iss_position;
```
But compatible with the latest version of Got

#1072


documentation/examples/runkit-example.js
```diff
@@ -0,0 +1,4 @@
+const got = require("got");
+const ISS = "http://api.open-notify.org/iss-now.json";
+
+(await got(ISS).json()).iss_position;
```
szmarczak, line 4:
```suggestion
(await got(ISS).json()).iss_position;

```

Giotino: I've not included the empty line at the end of the file because on the Runkit editor there would have been an empty line.

chrome_4FgDZvgLlq

szmarczak: Do it anyway :) Just to be consistent with other files ;)

szmarczak: Does the file need to be included in `files` in `package.json`?

Giotino:
> Does the file need to be included in `files` in `package.json`?

I don't know and I can't find any info about that...

szmarczak:
Let's give it a try :)

sindresorhus:
I guess this is better than a broken example, but the existing example was not really a good one. I was hoping we could add something better.

Giotino:
> I guess this is better than a broken example, but the existing example was not really a good one. I was hoping we could add something better.

Let's say that what we have done here is only a test in order to check if Runkit load our script, now we can work on a better example.

Issue #1191, opened 2020-04-26
document: make an HTTP request over TLS and disable server certificate verification 


jgehrcke:
For controlled environments like test suites one wants to be able to make the HTTP client do communication over TLS but instruct it to _not_ perform server certificate verification.

Looks like currently this can be done with
```
got('https://local.dev', { rejectUnauthorized: false });
```
see https://github.com/sindresorhus/got/issues/477#issuecomment-582450042.

I think this (or something like this) should be an official/documented part of the API. At least discussed / easy to find in the documentation.

Thanks!

sindresorhus:
I agree. We should add a section about this. It's a common use-case. We can add it after this section: https://github.com/sindresorhus/got#custom-endpoints

Giotino:
On the same topic there's also `checkServerIdentity` which enables custom certificate validation (useful for certificate pinning).

```
await got("https://google.com", {
  checkServerIdentity: (hostname: string, certificate: DetailedPeerCertificate) => {
    return;
  }
});
// I'm not 100% sure on the "DetailedPeerCertificate" type
```

It works, but on the Node.JS I can't find any documentation applied to `http.Request`.

>The following additional options from tls.connect() are also accepted: ca, cert, ciphers, clientCertEngine, crl, dhparam, ecdhCurve, honorCipherOrder, key, passphrase, pfx, rejectUnauthorized, secureOptions, secureProtocol, servername, sessionIdContext, highWaterMark.

But, some lines after, there's an example with it...
"Example pinning on certificate fingerprint, or the public key (similar to pin-sha256):"

This is just a guess, but:
core/index.ts:118
```
export interface Options extends URLOptions, SecureContextOptions
```
maybe should extend something like `ConnectionOptions` (from `tls`) and not `SecureContextOptions`

szmarczak:
> On the same topic there's also checkServerIdentity which enables custom certificate validation (useful for certificate pinning).

Can you send a PR for that? Thanks.

szmarczak:
I'll open an issue for that.

PR #1255, opened 2020-05-12
General improvements to HTTPS API

### LINKED ISSUE #1254: Document `checkServerIdentity` option
https://github.com/sindresorhus/got/issues/1191#issuecomment-627255748

/cc @Giotino

Giotino:
#### Checklist

Fixes #1254 

`checkServerIdentity` [signature](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L722): 
```typescript
(hostname: string, certificate: PeerCertificate) => Error | undefined
```

According to DefinitelyTyped `certificate` is [`PeerCertificate`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L37), but after testing I discovered that certificate has the property `issuerCertificate` that's mean it's [`DetailedPeerCertificate`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L53)

I can't find any documentation about using this property with `https.Request` on Node.JS, I think this is a side effect of the implementation of `https.Request`. But, on the Node.JS documentation there's an example using it ["Example pinning on certificate fingerprint, or the public key (similar to pin-sha256)"](https://nodejs.org/api/https.html#https_https_request_url_options_callback)

Node.JS even have a test about it:
https://github.com/nodejs/node/blob/8b4af64f50c5e41ce0155716f294c24ccdecad03/test/parallel/test-https-client-checkServerIdentity.js


Giotino:
With this PR i also wanted to add the documentation described in #1191

Giotino:
2 out of 3 current HTTPS test are broken, the first 2 make plain HTTP requests.

szmarczak:
> You should have used got.secure instead of got

Could you fix that please? :pray:

Giotino:
> > 2 out of the 3 current HTTPS test are broken, the first 2 make plain HTTP requests.
> > You should have used got.secure instead of got
> 
> Could you fix that please? 🙏

Yeah, I'm working on it

Giotino:
I was working on a test for `checkServerIdentity` and i discovered that it's not called when connecting to 127.0.0.1 (and other localhost addresses)...

```javascript
const tls = require('tls');

tls.connect({
  host: '127.0.0.1',
  port: 49637,
  checkServerIdentity: (hostname, certificate) => {
    // This is never called for localhost addresses
  }
})
```

Of course there's no documentation by Node.JS, there's only a comment in an example:
```javascript
// Necessary only if the server's cert isn't for "localhost".
checkServerIdentity: () => { return null; },
```

I've to keep in mind this when writing the documentation, also the test need to be performed using an external service, we can use `https://badssl.com/`

szmarczak:
Maybe there's a `servername` option missing? Try setting it to `localhost`.

szmarczak:
Or maybe try setting `host` to `localhost` instead of `127.0.0.1`...

Giotino:
> Maybe there's a `servername` option missing? Try setting it to `localhost`.
> Or maybe try setting `host` to `localhost` instead of `127.0.0.1`...

I've tried all of them, same result

Giotino:
Seems like I made a mistake, it seems that `checkServerIdentity` is only called if the certificate is valid. That's why localhost was not working (I was testing without the test CA). I'm going to add some tests.

Giotino:
While `Options` extednds `SecureContextOptions` these features are not documented.

For example `key`, `cert` and `ca` useful for using a client certificate, or other params to tweak the TLS options.

szmarczak:
I agree.

Giotino:
Since `key`, `cert` and `passphrase` are always used together I grouped the examples in a single block.

readme.md
```diff
@@ -40,6 +40,7 @@ For browser usage, we recommend [Ky](https://github.com/sindresorhus/ky) by the
 - [Errors with metadata](#errors)
 - [JSON mode](#json-mode)
 - [WHATWG URL support](#url)
+- [HTTPS](#https)
```
szmarczak, line 43:
```suggestion
- [HTTPS API](#https)
```

szmarczak: Not sure if `HTTPS certificate API` would be better...

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
 
 The IP address used to send the request from.
 
+### HTTPS
```
sindresorhus, line 823: I don't think it's clear enough what this means. Does it mean that these options will be ignored when not using HTTPS or will it throw an error.

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+
+##### ca
```
sindresorhus, line 825: I would prefer to use the full word for this. (I do realize this is just passed through to Node.js, but we can do better).

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+
+Type: `string` | `string[]` | `Buffer` | `Buffer[]`
```
sindresorhus, line 827:
```suggestion
Type: `string | string[] | Buffer | Buffer[]`
```

Does this mean that you cannot have an array of string **and** buffer?

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+
+Override the default CAs ([from Mozilla](https://ccadb-public.secure.force.com/mozilla/IncludedCACertificateReport))
```
sindresorhus, line 829: Don't use acronyms.

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+		fs.readFileSync('./my_ca2.pem')
+	]
+});
+```
+
+##### key
+
+Type: `string` | `string[]` | `Buffer` | `Buffer[]` | `Object[]`
+
+Private keys in PEM format.\
```
sindresorhus, line 848: Linkify `PEM` to some place that explains it.

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+PEM allows the option of private keys being encrypted. Encrypted keys will be decrypted with `options.passphrase`.\
+Multiple keys with different passphrases can be provided as an array of `{pem: <string|Buffer>, passphrase: <string>}`
+
+##### cert
```
sindresorhus, line 852:
```suggestion
##### certificate
```

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+});
+```
+
+##### key
+
+Type: `string` | `string[]` | `Buffer` | `Buffer[]` | `Object[]`
+
+Private keys in PEM format.\
+PEM allows the option of private keys being encrypted. Encrypted keys will be decrypted with `options.passphrase`.\
+Multiple keys with different passphrases can be provided as an array of `{pem: <string|Buffer>, passphrase: <string>}`
```
sindresorhus, line 850:
```suggestion
Multiple keys with different passphrases can be provided as an array of `{pem: <string | Buffer>, passphrase: <string>}`
```

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+
+##### cert
+
+Type: `string` | `string[]` | `Buffer` | `Buffer[]`
+
+Cert chains in PEM format.\
```
sindresorhus, line 856: "Cert"

Use the full word.

Also linkify to a place that explains "certificate chains".

readme.md
```diff
@@ -844,6 +931,31 @@ const got = require('got');
...
 ```
 
+##### checkServerIdentity
+
+Type: `Function`\
+Signature: `(hostname: string, certificate: DetailedPeerCertificate) => Error | undefined`\
+Default: `tls.checkServerIdentity` (from the `tls` module)
+
+This function enable a custom check of the certificate.\
+N.B. In order to have the function called the certificate must not be `expired`, `self-signed` or with an `untrusted-root`.\
```
sindresorhus, line 941: "N.B"

Don't use acronyms

readme.md
```diff
@@ -819,6 +820,92 @@ Type: `string`
...
+One cert chain should be provided per private key (`options.key`).\
+When providing multiple cert chains, they do not have to be in the same order as their private keys in `options.key`.\
+If the intermediate certificates are not provided, the peer will not be able to validate the certificate, and the handshake will fail.
+
+##### passphrase
+
+Type: `string`
+
+The passphrase to decrypt the `key` (if different keys have different passphrases refer to `key` documentation).
+
```
sindresorhus, line 866:
```suggestion
```

readme.md
szmarczak:
```suggestion
### Advanced HTTPS API
```

readme.md
Giotino: Is [wikipedia](https://en.wikipedia.org/wiki/X.509#Certificate_chains_and_cross-certification) fine?

readme.md
Giotino: > Does it mean that these options will be ignored when not using HTTPS or will it throw an error.

I think it should just ignore them.
I can write a line right after the "HTTPS" (or whatever we are going to call the section) title like:
`Note: If the request is not HTTPS these options will be ignored.`

readme.md
szmarczak: MDN is better. Do they have it?

Giotino: I can only find a generic page about certificates
https://developer.mozilla.org/en-US/docs/Mozilla/Security/x509_Certificates
And a small paragraph about Certificate Authorities
https://developer.mozilla.org/en-US/docs/Glossary/Certificate_authority

Giotino:
@sindresorhus If you want to have different names (from the tls module) for the properties related to HTTPS I think that `Options` should stop extending `SecureContextOptions` and start defining its HTTPS properties and then translate them before the request.
I'm working on a commit with an example of that.

readme.md
szmarczak: https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning

Giotino:
We have 2 options:
- Removing old HTTPS related variables (given by Node.JS)
- Or simply add an "alias" as I done for `certificate` (alias to `cert` and only add it to the documentation)

The reason why I'm suggesting the second is because there might be someone that was already using the property names given by Node.JS. They were undocumented (in the readme), but `Options` extended `SecureContextOptions` so they were "documented".

readme.md
Giotino: > https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning

That's about HPKP, that's a browser thing. The Certificate Pinning I'm talking about is nearly the same thing, but it's performed voluntarily by the client (application), the client specify which CAs and/or certificate to trust.

szmarczak: Then Wikipedia it is :)

szmarczak:
Let's go with the first one.

Giotino:
> Let's go with the first one.

Just to be sure we are thinking the same think:
If we go with the first one all of the implementation already using `cert` or `ca` will break.

If we go with the second one we keep compatibility with that implementations. (I implemented this is the last commit, I was already working on it)

szmarczak:
@sindresorhus Giotino's right. Maybe let's make it a breaking change for Got 12 and for now just document `ca` and `cert`?

Giotino:
> @sindresorhus Giotino's right. Maybe let's make it a breaking change for Got 12 and for now just document `ca` and `cert`?

I was proposing another way.
Leave `ca` and `cert` undocumented (options are passed to the TLS object anyways) and document `certificateAuthority` and `certificate` (as in may PR).
Maybe give a warning when using `ca` and `cert` as they're "deprected" (never documented, but they're going to stop working in the future).

szmarczak:
I don't like issues like this. I'm not feeling comfortable with renaming, so I'm not voting on this one :sweat_smile:

sindresorhus:
> I was proposing another way.
Leave ca and cert undocumented (options are passed to the TLS object anyway) and document certificateAuthority and certificate (as in may PR).
Maybe give a warning when using ca and cert as they're "deprected" (never documented, but they're going to stop working in the future).

👍

Giotino:
> > I was proposing another way.
> > Leave ca and cert undocumented (options are passed to the TLS object anyway) and document certificateAuthority and certificate (as in may PR).
> > Maybe give a warning when using ca and cert as they're "deprected" (never documented, but they're going to stop working in the future).
> 
> 👍

This PR branch is actually doing it. If a user use `ca` or `cert`.
```js
console.warn('WARNING: "options.cert" was never documented, please use "options.certificate"');
```
Is called. (obviously slightly different for `ca`).

`certificateAuthority` and `certificate` (if presents) will override `ca` and `cert` right before the request, but the warning is displayed while merging the options.

sindresorhus:
There's an unanswered question in: https://github.com/sindresorhus/got/pull/1255#discussion_r424455674

readme.md
```diff
@@ -819,6 +820,93 @@ Type: `string`
...
+##### certificateAuthority
+
+Type: `string | string[] | Buffer | Buffer[]`
+
+Override the default Certificate Authorities ([from Mozilla](https://ccadb-public.secure.force.com/mozilla/IncludedCACertificateReport))
+
+```js
+// Single CA
+got('https://example.com', {ca: fs.readFileSync('./my_ca.pem')});
```
sindresorhus, line 835: Needs to be updated.

readme.md
```diff
@@ -819,6 +820,93 @@ Type: `string`
...
+
+// Multiple CAs
+got('https://example.com', {
+	ca: [
+		fs.readFileSync('./my_ca1.pem'),
+		fs.readFileSync('./my_ca2.pem')
+	]
+});
```
sindresorhus, line 843: I don't think we need two examples for this. Users should know how to use arrays.

readme.md
```diff
@@ -819,6 +820,93 @@ Type: `string`
...
+```
+
+##### key
```
sindresorhus, line 846: I feel like `key` is too general of an option name.

readme.md
```diff
@@ -819,6 +820,93 @@ Type: `string`
...
+When providing multiple cert chains, they do not have to be in the same order as their private keys in `options.key`.\
+If the intermediate certificates are not provided, the peer will not be able to validate the certificate, and the handshake will fail.
+
+##### passphrase
+
+Type: `string`
+
+The passphrase to decrypt the `key` (if different keys have different passphrases refer to `key` documentation).
+
+##### Examples for key, cert and passphrase
```
sindresorhus, line 869: Needs to be updated.

readme.md
```diff
@@ -819,6 +820,93 @@ Type: `string`
...
+
+```js
+// Single key with cert
+got('https://example.com', {
+	key: fs.readFileSync('./client_key.pem'),
+	cert: fs.readFileSync('./client_cert.pem'),
```
sindresorhus, line 875: Needs to be updated.

readme.md
```diff
@@ -844,6 +932,31 @@ const got = require('got');
...
+Signature: `(hostname: string, certificate: DetailedPeerCertificate) => Error | undefined`\
+Default: `tls.checkServerIdentity` (from the `tls` module)
+
+This function enable a custom check of the certificate.\
+Note: In order to have the function called the certificate must not be `expired`, `self-signed` or with an `untrusted-root`.\
+The function parameters are
+- `hostname`: the server hostname (used when connecting)
+- `certificate`: the server certificate
+
+The function must return `undefined` if the check succeeded or and `Error` if it failed.
```
sindresorhus, line 947: `or and`?

readme.md
```diff
@@ -844,6 +932,31 @@ const got = require('got');
...
 
+##### checkServerIdentity
+
+Type: `Function`\
+Signature: `(hostname: string, certificate: DetailedPeerCertificate) => Error | undefined`\
+Default: `tls.checkServerIdentity` (from the `tls` module)
+
+This function enable a custom check of the certificate.\
+Note: In order to have the function called the certificate must not be `expired`, `self-signed` or with an `untrusted-root`.\
+The function parameters are
```
sindresorhus, line 943:
```suggestion
The function parameters are:
```

readme.md
```diff
@@ -844,6 +932,31 @@ const got = require('got');
...
+- `hostname`: the server hostname (used when connecting)
```
sindresorhus, line 944:
```suggestion
- `hostname`: The server hostname (used when connecting)
```

readme.md
```diff
@@ -844,6 +932,31 @@ const got = require('got');
...
+- `certificate`: the server certificate
```
sindresorhus, line 945:
```suggestion
- `certificate`: The server certificate
```

readme.md
```diff
@@ -844,6 +932,31 @@ const got = require('got');
...
+
+The function must return `undefined` if the check succeeded or and `Error` if it failed.
+
+```js
+await got('https://example.com', {
+	checkServerIdentity: (hostname, certificate) => {
+		if (hostname === 'example.com')
```
sindresorhus, line 952: Use braces

source/core/index.ts
```diff
@@ -150,6 +149,15 @@ export interface Options extends URLOptions, SecureContextOptions {
 	socketPath?: string;
 	method?: Method;
 	createConnection?: (options: http.RequestOptions, oncreate: (error: Error, socket: Socket) => void) => Socket;
+
+	// From tls.ConnectionOptions
```
sindresorhus, line 153:
```suggestion
	// From `tls.ConnectionOptions`
```

source/core/index.ts
```diff
@@ -833,6 +844,15 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			}
 		}
 
+		// HTTPS options
+		if ('ca' in options) {
+			console.warn('WARNING: "options.ca" was never documented, please use "options.certificateAuthority"');
```
sindresorhus, line 849: Use `util.deprecate`

readme.md
Giotino: > Does this mean that you cannot have an array of string and buffer?

Sorry, I missed this line.

Node.JS documentation.
> or an Array of strings and/or Buffers

It's ambiguous, but, after some testing, I can say that it accepts both mixed in the same array.
The type should be `string | Buffer | (string | Buffer)[]`

source/core/index.ts
Giotino: According to Node.JS documentation `util.deprecrate` is made for functions.
https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

Giotino: According to the Node.JS documentation `util.deprecrate` is made for functions.
https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

source/core/index.ts
Giotino: Some line before there's `// From http.RequestOptions` which was already there, shoud I also change it?

Giotino:
Considering this comment
> I feel like `key` is too general of an option name.

Maybe we should have something like this for all this HTTPS options
```javascript
got('http://google.com', {
  https: {
    key: ...,
    certificate: ...,
    certificateAuthority: ...,
    rejectUnauthorized: ...,
    checkServerIdentity: ...
  },
  rejectUnauthorized: ... //I'm also here for backwards compatibility (but I'm overwritten by the other)
});
```

source/core/index.ts
sindresorhus: Yup

sindresorhus:
> Maybe we should have something like this for all this HTTPS options

I like that. @szmarczak Are you ok with this?

szmarczak:
Yep. Looks more elegant 👌

readme.md
Giotino: Typo: `or an`

source/core/index.ts
```diff
@@ -24,6 +24,7 @@ import urlToOptions from './utils/url-to-options';
 import optionsToUrl, {URLOptions} from './utils/options-to-url';
 import WeakableMap from './utils/weakable-map';
 import decompressResponse from './utils/decompress-response';
+import deprecationWarning from '../utils/deprecation-warning';
```
szmarczak, line 27: Use [`util.deprecate`](https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code). You can pass a noop as the function.

source/core/index.ts
```diff
@@ -845,12 +862,28 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		}
 
 		// HTTPS options
+		if ('rejectUnauthorized' in options) {
+			deprecationWarning('options.rejectUnauthorized', '"options.rejectUnauthorized" is now deprecated, please use "options.https.rejectUnauthorized"');
```
szmarczak, line 870: Can't we just pass the message here? I don't think the first argument is needed at all :)

source/core/index.ts
```diff
@@ -1350,14 +1383,33 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
 
-		const requestOptions = options as unknown as https.RequestOptions;
+		const requestOptions = options as unknown as RealRequestOptions;
 
-		if (options.certificateAuthority) {
-			requestOptions.ca = options.certificateAuthority;
-		}
+		// HTTPS options remapping
+		if (options.https) {
+			if ('rejectUnauthorized' in options.https) {
```
szmarczak, line 1394:
```suggestion
			if (options.https.rejectUnauthorized) {
```

source/index.ts
```diff
@@ -68,7 +68,7 @@ const defaults: InstanceDefaults = {
 		// TODO: Set this to `true` when Got 12 gets released
 		http2: false,
 		allowGetBody: false,
-		rejectUnauthorized: true,
+		https: undefined,
```
szmarczak, line 71: Nope. Please define this as object with `rejectUnauthorized` set to `true` :)

source/core/index.ts
Giotino: I wrote that line like that because
`if(options.https.rejectUnauthorized)` is not taken if `https.rejectUnauthorized` is `false`, so `https.rejectUnauthorized` is unable to override `rejectUnauthorized` if `false`.

source/index.ts
Giotino: The problem is that `https.rejectUnauthorized` is going to override `rejectUnauthorized` if a used is using `rejectUnauthorized`

szmarczak: I'm fine with that. Show 'em a deprecation warning then :)

source/core/index.ts
```diff
@@ -623,8 +646,17 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		assert.any([is.boolean, is.undefined], options.throwHttpErrors);
 		assert.any([is.boolean, is.undefined], options.http2);
 		assert.any([is.boolean, is.undefined], options.allowGetBody);
-		assert.any([is.boolean, is.undefined], options.rejectUnauthorized);
 		assert.any([is.string, is.undefined], options.localAddress);
+		assert.any([is.object, is.undefined], options.https);
+		assert.any([is.boolean, is.undefined], options.rejectUnauthorized);
+		if (options.https) {
```
szmarczak, line 652: Can you remove this if?

szmarczak: No, you're right, sorry :sweat_smile:

source/utils/deprecation-warning.ts
```diff
@@ -0,0 +1,14 @@
...
+export default (message: string) => {
+	if (alreadyWarned.includes(message)) {
+		return;
+	}
+
+	alreadyWarned.push(message);
+
+	// @ts-ignore
+	process.emitWarning(message, {
+		type: 'DeprecationWarning'
```
szmarczak, line 12:
```suggestion
		type: 'GotDeprecationWarning'
```

source/utils/deprecation-warning.ts
```diff
@@ -0,0 +1,14 @@
+const alreadyWarned: string[] = [];
+
+export default (message: string) => {
+	if (alreadyWarned.includes(message)) {
+		return;
+	}
+
+	alreadyWarned.push(message);
+
+	// @ts-ignore
```
szmarczak, line 10: Please include a comment. If the types are missing, then:

```suggestion
	// @ts-ignore Missing types.
```

source/core/index.ts
```diff
@@ -3,7 +3,7 @@ import {Duplex, Writable, Readable} from 'stream';
 import {ReadStream} from 'fs';
 import {URL, URLSearchParams} from 'url';
 import {Socket} from 'net';
-import {SecureContextOptions} from 'tls';
+import {SecureContextOptions, DetailedPeerCertificate} from 'tls';
```
szmarczak, line 6:
```suggestion
import {ConnectionOptions, SecureContextOptions, DetailedPeerCertificate} from 'tls';
```

source/core/index.ts
```diff
@@ -116,7 +117,13 @@ type CacheableRequestFn = (
 	cb?: (response: ServerResponse | ResponseLike) => void
 ) => CacheableRequest.Emitter;
 
-export interface Options extends URLOptions, SecureContextOptions {
+type CheckServerIdentityFn = (hostname: string, certificate: DetailedPeerCertificate) => Error | undefined | void;
```
szmarczak, line 120:
```suggestion
```

source/core/index.ts
```diff
@@ -116,7 +117,13 @@ type CacheableRequestFn = (
...
+
+interface RealRequestOptions extends https.RequestOptions {
+	checkServerIdentity: CheckServerIdentityFn;
```
szmarczak, line 123:
```suggestion
	checkServerIdentity: ConnectionOptions['checkServerIdentity'];
```

source/core/index.ts
```diff
@@ -141,15 +148,31 @@ export interface Options extends URLOptions, SecureContextOptions {
...
+
+	https?: HTTPSOptions;
+}
+
+export interface HTTPSOptions {
+	// From `http.RequestOptions` and `tls.CommonConnectionOptions`
+	rejectUnauthorized?: https.RequestOptions['rejectUnauthorized'];
+
+	// From `tls.ConnectionOptions`
+	checkServerIdentity?: CheckServerIdentityFn;
```
szmarczak, line 169:
```suggestion
	checkServerIdentity?: ConnectionOptions['checkServerIdentity'];
```

source/core/index.ts
Giotino: The problem is that `DefinitelyTyped ` describe the `certificate` argument as `PeerCertificate`, but it's a `DetailedPeerCertificate` (the same thing, but 1 more property)

source/core/index.ts
Giotino: The problem is that `DefinitelyTyped ` describe the `certificate` argument as `PeerCertificate`, but it's a `DetailedPeerCertificate` (the same thing, but 1 more property)

Giotino:
"Deprecation Warning" tests must be serial or else they'll steal each other `process` events.

source/utils/deprecation-warning.ts
szmarczak: ?

Giotino: By emitting a warn with type: `DeprecationWarning` Node.JS sees it as a deprecation.
Therefore you can use `--trace-deprecation`.

szmarczak: Or maybe `Got/DeprecationWarning`?

Giotino: It only works with exactly "DeprecationWarning".

source/utils/deprecation-warning.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default (message: string) => {
+	if (alreadyWarned.includes(message)) {
+		return;
+	}
+
+	alreadyWarned.push(message);
+
+	// @ts-ignore Missing types.
+	process.emitWarning(message, {
```
szmarczak, line 11:
```suggestion
	process.emitWarning(`Got: ${message}`, {
```

maybe this?

Giotino: The output will be
`(node:29572) DeprecationWarning: Got: "options.rejectUnauthorized" is now deprecated, please use "options.https.rejectUnauthorized"`

szmarczak: Yup. So people know it's from Got without enabling `--trace-warnings`.

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
 
 The IP address used to send the request from.
 
-##### rejectUnauthorized
+### Advanced HTTPS API
+
+Note: If the request is not HTTPS these options will be ignored.
```
sindresorhus, line 825:
```suggestion
Note: If the request is not HTTPS, these options will be ignored.
```

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
...
+
+##### https.certificateAuthority
+
+Type: `string | Buffer | (string | Buffer)[]`
```
sindresorhus, line 829:
```suggestion
Type: `string | Buffer | Array<string | Buffer>`
```

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
...
+got('https://example.com', {
+	https: {
+		certificateAuthority: fs.readFileSync('./my_ca.pem')
+	}
+});
+```
+
+##### https.key
+
+Type: `string | Buffer | (string | Buffer)[] | Object[]`
```
sindresorhus, line 844:
```suggestion
Type: `string | Buffer | Array<string | Buffer> | object[]`
```

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
...
+When providing multiple cert chains, they do not have to be in the same order as their private keys in `options.https.key`.\
+If the intermediate certificates are not provided, the peer will not be able to validate the certificate, and the handshake will fail.
+
+##### https.passphrase
+
+Type: `string`
+
+The passphrase to decrypt the `options.https.key` (if different keys have different passphrases refer to `options.https.key` documentation).
+
+##### Examples for https.key, https.certificate and https.passphrase
```
sindresorhus, line 865:
```suggestion
##### Examples for `https.key`, `https.certificate` and `https.passphrase`
```

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
...
+
+```js
+// Single key with cert
```
sindresorhus, line 868:
```suggestion
// Single key with certificate
```

readme.md
```diff
@@ -819,7 +820,98 @@ Type: `string`
...
+got('https://example.com', {
+	https: {
+		key: fs.readFileSync('./client_key.pem'),
+		certificate: fs.readFileSync('./client_cert.pem')
+	}
+});
+
+// Multiple keys with certs (out of order)
```
sindresorhus, line 876:
```suggestion
// Multiple keys with certificates (out of order)
```

source/core/index.ts
sindresorhus, line 120:
```suggestion
type CheckServerIdentityFn = (hostname: string, certificate: DetailedPeerCertificate) => Error | void;
```

?

source/core/index.ts
```diff
@@ -141,15 +148,31 @@ export interface Options extends URLOptions, SecureContextOptions {
...
 	headers?: Headers;
 	methodRewriting?: boolean;
 
-	// From http.RequestOptions
+	// From `http.RequestOptions`
 	localAddress?: string;
 	socketPath?: string;
 	method?: Method;
 	createConnection?: (options: http.RequestOptions, oncreate: (error: Error, socket: Socket) => void) => Socket;
+	rejectUnauthorized?: boolean; // Here for backwards compatibility
```
sindresorhus, line 159: Can you add a TODO comment about removing this in Got 12?

source/utils/deprecation-warning.ts
```diff
@@ -0,0 +1,14 @@
+const alreadyWarned: string[] = [];
```
sindresorhus, line 1: A `Set` would be more appropriate.

source/utils/deprecation-warning.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default (message: string) => {
+	if (alreadyWarned.includes(message)) {
+		return;
+	}
+
+	alreadyWarned.push(message);
+
+	// @ts-ignore Missing types.
```
sindresorhus, line 10:
```suggestion
	// @ts-expect-error Missing types.
```

test/https.ts
```diff
@@ -1,27 +1,100 @@
...
+		https: {
+			certificateAuthority: server.caCert
+		},
+		headers: {host: 'example.com'}
 	});
+
+	t.is(body, 'ok');
+});
+
+test('https request with checkServerIdentity OK', withServer, async (t, server, got) => {
```
sindresorhus, line 33:
```suggestion
test('https request with `checkServerIdentity` OK', withServer, async (t, server, got) => {
```

test/https.ts
```diff
@@ -1,27 +1,100 @@
...
+				t.is(certificate.issuer.CN, 'localhost');
+			}
+		},
+		headers: {host: 'example.com'}
+	});
+
 	t.is(body, 'ok');
 });
 
+test('https request with checkServerIdentity NOT OK', withServer, async (t, server, got) => {
```
sindresorhus, line 53:
```suggestion
test('https request with `checkServerIdentity` NOT OK', withServer, async (t, server, got) => {
```

test/https.ts
```diff
@@ -34,3 +107,67 @@ test('http2', async t => {
 	t.is(headers[':status'], 200);
 	t.is(typeof body, 'string');
 });
+
+test.serial('deprecated rejectUnauthorized', withServer, async (t, server, got) => {
```
sindresorhus, line 111:
```suggestion
test.serial('deprecated `rejectUnauthorized` option', withServer, async (t, server, got) => {
```

test/https.ts
```diff
@@ -34,3 +107,67 @@ test('http2', async t => {
...
+	server.get('/', (_request, response) => {
+		response.end('ok');
+	});
+
+	await new Promise(resolve => {
+		process.once('warning', warn => {
```
sindresorhus, line 117: Would be better to use `p-event`.

test/https.ts
```diff
@@ -34,3 +107,67 @@ test('http2', async t => {
...
+				rejectUnauthorized: false
+			});
+
+			t.fail();
+			resolve();
+		})();
+	});
+});
+
+test.serial('non deprecated rejectUnauthorized', withServer, async (t, server, got) => {
```
sindresorhus, line 133:
```suggestion
test.serial('non-deprecated `rejectUnauthorized` option', withServer, async (t, server, got) => {
```

test/https.ts
```diff
@@ -34,3 +107,67 @@ test('http2', async t => {
...
+
+	process.once('warning', warn => {
+		t.is(warn.name, 'DeprecationWarning');
+	});
+
+	await got.secure({
+		rejectUnauthorized: false
+	});
+
+	process.once('warning', warn => {
```
sindresorhus, line 164: Use `p-event`

test/https.ts
Giotino:
```js
pEvent(process, 'warning').then(warn => {
  t.is(warn.name, 'DeprecationWarning');
  resolve();
});
```
That's the best I can do without breaking the logic of the test. I can't remove that promise.

I can achieve something by using `Promise.race`, but it's going to be harder to read.

sindresorhus: At least do:

```js
(async () => {
	const warning = await pEvent(process, 'warning');
	t.is(warning.name, 'DeprecationWarning');
	resolve();
})();
```

sindresorhus:
There are a few more unresolved inline comments.

test/https.ts
szmarczak: The `async () => ` function wrapper is not required as the test itself is async.

Giotino: > The `async () => ` function wrapper is not required as the test itself is async.

It's required because the test must continue in order to make the request, then when the event received is checked.

source/utils/deprecation-warning.ts
Giotino: Since this is as Typescript 3.9 feature and the repo is still using 3.8 I think it's better to leave it as it is.

Giotino:
> There are a few more unresolved inline comments.

Sorry, I missed them, I was watching only the "Conversation" tab on Github.

sindresorhus:
This looks great now. Thanks for your hard work on this 🙌

Issue #1187, opened 2020-04-25
The `onCancel` handler was attached after the promise settled


myfreeer:
#### Describe the bug

 - Node.js version: 12.16.2 x64 (LTS)
 - OS & version: Windows 10 LTSC x64
 - got version: [v11.0.2](https://github.com/sindresorhus/got/releases/tag/v11.0.2)

Using `Promise API` with socket timeout and retry, this does not reproduce on Node.js 14.0.0.

#### Actual behavior
Promise rejected before retry, and stderr printed message like below:
```
(node:9792) UnhandledPromiseRejectionWarning: Error: The `onCancel` handler was attached after the promise settled.
    at onCancel (D:\TEMP\test-case\node_modules\p-cancelable\index.js:46:12)
    at makeRequest (D:\TEMP\test-case\node_modules\got\dist\source\as-promise\index.js:39:13)
    at Timeout.retry [as _onTimeout] (D:\TEMP\test-case\node_modules\got\dist\source\as-promise\index.js:182:25)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7)
(node:9792) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:9792) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Edited: some logs showing the timestamp.
```
1587995291087 response event fired
1587995291088 before await getStream.buffer(request)
1587995291117 error event fired
1587995291118 catch block after await getStream.buffer(request)
1587995291119 promise rejected in catch block
1587995292181 retry
```

#### Expected behavior
Retryed.

#### Code to reproduce

```js
const http = require('http');

const got = require('got').extend({
  timeout: {
    socket: 20
  },
  retry: {
    limit: 5
  }
});

let requested = false;

let server = http.createServer(((req, res) => {
  if (requested) {
    res.writeHead(200, {
      'Content-Length': 8
    });
    res.write('ok123456');
    return;
  }
  req.socket.pause();
  setTimeout(() => {
    res.writeHead(200, {
[...]

PR #1256, opened 2020-05-12
Use async iterators to get response body

szmarczak:
Just to utilize newest Node.js features :)

#### Checklist


source/core/utils/get-buffer.ts
```diff
@@ -0,0 +1,15 @@
...
+// TODO: Update `get-stream`
+
+const getBuffer = async (stream: Readable) => {
+	const chunks = [];
+
+	for await (const chunk of stream) {
+		chunks.push(Buffer.from(chunk));
+	}
+
+	return Buffer.concat(chunks);
```
sindresorhus, line 12: I would recommend providing the total length as it can slightly improve performance. https://github.com/sindresorhus/get-stdin/blob/master/index.js

sindresorhus:
Why not just do this directly in `get-stream`? We're planning to do the change there anyway.

szmarczak:
> Why not just do this directly in get-stream? We're planning to do the change there anyway.

Sure, I just want to test things before we make an actual release :)

szmarczak:
I'll merge this for now. I think this **may** fix #1187.

arcanis:
@szmarczak We upgraded `got` in Yarn, and I noticed it caused a breakage with all requests being returned with no content (it doesn't happen to other contributors, weirdly, even on the same Node versions). After investigating, the problem got introduced between the 11.3 and the 11.4, and more specifically by this PR.

I can patch it locally by replacing the following:

```ts
for await (const chunk of stream) {
    chunks.push(chunk);
    length += Buffer.byteLength(chunk);
}
```

By this:

```ts
const {PassThrough} = require(`stream`);

const copy = new PassThrough();
stream.pipe(copy);

for await (const chunk of copy) {
    chunks.push(chunk);
    length += Buffer.byteLength(chunk);
}
```

I suspect this is because the various Node stream interfaces aren't meant to be used together, and weird effects are possible should that happen. I don't know if that's the case for sure here 🙁

szmarczak:
@arcanis That seems weird. What Node.js version are you using? It'd be best to open a new issue so we don't lose the track here.

This looks like a Node.js bug.

Michelle99798 (approved):
Thanks 👍

PR #1264, opened 2020-05-14
Add `dnsLookupIpVersion` option

### LINKED ISSUE #1263: The types are missing `family` option
Hi, 
I want to use my local address v6 address but when i use "localAddress" property it show 
```typescript
import got from 'got'
got('https://api6.ipify.org', {
       localAddress: '2604:A880:0400:00D0:0000:0000:1832:C001'
})

// RequestError: bind EINVAL ... 

```
How to use my local ipv6 address
Thank you ?

Giotino:
#### Checklist

I've added the documentation relative to the `family` property. I've also added a stricter type definition.

I think the name "family" it's a little generic, it might be renamed to "IPfamily" or something like that.

Fixes #1263


readme.md
```diff
@@ -499,6 +499,17 @@ An instance of [`CacheableLookup`](https://github.com/szmarczak/cacheable-lookup
 **Note:** This should stay disabled when making requests to internal hostnames such as `localhost`, `database.local` etc.\
 `CacheableLookup` uses `dns.resolver4(..)` and `dns.resolver6(...)` under the hood and fall backs to `dns.lookup(...)` when the first two fail, which may lead to additional delay.
 
+###### family
```
sindresorhus, line 502:
```suggestion
###### ipVersion
```

?

readme.md
```diff
@@ -499,6 +499,17 @@ An instance of [`CacheableLookup`](https://github.com/szmarczak/cacheable-lookup
...
+
+Type: `0 | 4 | 6`\
```
sindresorhus, line 504: I don't like `0` having some special meaning.

Could we make this:

```suggestion
Type: `'auto' | 4 | 6`\
```

sindresorhus: Or maybe this to make the values the same type:

```suggestion
Type: `'auto' | 'ipv4' | 'ipv6'`\
```

szmarczak: Why not make this an array? `prefferedIpVersion: ['ipv4', 'ipv6']`? It would require us to pass custom `lookup` option internally.

Giotino: > I don't like `0` having some special meaning.
> 
> Could we make this:

I was thinking the same think, but that's going to break some implementation already using that undocumented feature.
If we rename it to `ipVersion` as you suggested I think
> Type: `'auto' | 'ipv4' | 'ipv6'`
Will be perfectly fine.

sindresorhus: > Why not make this an array? `prefferedIpVersion: ['ipv4', 'ipv6']`? It would require us to pass custom `lookup` option internally.

What would be the benefit of that? Does it solve a problem with the existing option?

Giotino: > Why not make this an array? `prefferedIpVersion: ['ipv4', 'ipv6']`? It would require us to pass custom `lookup` option internally.

The thing is: "There's really a need to have a preference on the IP version?"
When a user can explicit set the IP version or have "4 or 6" I can't find any usage for the "6 or 4" option.
I don't know if implementing it would be a good thing, as you said, we would have to implement a custom `lookup` function.

szmarczak: > Does it solve a problem with the existing option?

Is there a problem at all? No. It's just a nitpick. Some people prefer IPv6 over IPv4.

> The thing is: "There's really a need to have a preference on the IP version?"

That's the point. I don't think so. Then what does `auto` mean? It means: if there are no IPv4 addresses found, it will use IPv6. That needs to be documented.

source/core/index.ts
```diff
@@ -1330,6 +1333,17 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		delete options.request;
 		delete options.timeout;
 
+		const requestOptions = options as any; // Temporary workaround
```
szmarczak, line 1336:
```suggestion
		// Temporary workaround
		const requestOptions = options as any;
```

source/core/index.ts
```diff
@@ -1330,6 +1333,17 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
+
+		// If `ipVersion` is not present do not override `family`
```
szmarczak, line 1338: But... you know it defaults to `0`? So the last `else if` is unnecessary.

Giotino: I was thinking on a way to keep `family` is the user used that property.

But then I added it to the defaults...
If `ipVersion` has no default from Got `family` is keeped, if I remove the last `else if` it will be overwritten.

source/core/index.ts
Giotino: That's really a temp thing, before the merge (if any) it well be replaced with something better.

source/core/index.ts
szmarczak: > I was thinking on a way to keep family is the user used that property.

Then you need to change the last `else if`:

```
} else if (options.ipFamily === 'auto' && !('family' in options)) {
```

readme.md
Giotino: > It means: if there are no IPv4 addresses found, it will use IPv6. That needs to be documented.

I think you may have missed it, but i wrote it in the README.md

>  - `auto`: IPv4 (if present) or IPv6

source/core/index.ts
Giotino: The logic I am trying to have is:
`family = ipVersion !== undefined ? translate(ipVersion) : family`

Where `translate` is:
- `auto` => 0
- `ipv4` => 4
- `ipv6` => 6

Giotino:
I've added some IPv6 (and DNS) related tests

Giotino:
Since the `family` (now `ipVersion`) property is only related to the DNS aspect of the request maybe we should change its name to something that say explicitly that is referred to the DNS (e.g. `dnsResolutionIpVersion` or `dnsIpVersion`)

Giotino:
Since `family` (now `ipVersion`) is only related to DNS I think it should be renamed to something like `dnsIpVersion` or `resolvedIpVersion`.

readme.md
```diff
@@ -499,6 +499,19 @@ An instance of [`CacheableLookup`](https://github.com/szmarczak/cacheable-lookup
 **Note:** This should stay disabled when making requests to internal hostnames such as `localhost`, `database.local` etc.\
 `CacheableLookup` uses `dns.resolver4(..)` and `dns.resolver6(...)` under the hood and fall backs to `dns.lookup(...)` when the first two fail, which may lead to additional delay.
 
+###### dnsIpVersion
+
+Type: `'auto' | 'ipv4' | 'ipv6'`\
+Default: `'auto'`
+
+Indicates which DNS record family to use.\
```
sindresorhus, line 507: Can you maybe include a use-case of when the user would want to override this? It's not clear even to me when I would use this.

source/core/utils/dns-ip-version.ts
```diff
@@ -0,0 +1,24 @@
+export type DnsIpVersion = 'auto' | 'ipv4' | 'ipv6';
+export type DnsIpFamily = 0 | 4 | 6;
```
sindresorhus, line 2: Why are you exporting this? It's only used here.

sindresorhus:
Can you also add a `util.deprecate` call if `family` option is used?

source/core/utils/dns-ip-version.ts
Giotino: That was a mistake, I used to use it, but when I removed the reference I forgot the export.

readme.md
Giotino: For example, you can use it to retrieve a user public IPv6
```js
// api6.ipify.org is resolved as IPv4, the request is going to be over IPv4 (the website will respond with your IPv4)
await got('https://api6.ipify.org', {
  dnsIpVersion: 'ipv4'
});

// api6.ipify.org is resolved as IPv6, the request is going to be over IPv6 (the website will respond with your IPv6)
await got('https://api6.ipify.org', {
  dnsIpVersion: 'ipv6'
});
```

Is this example good for the documentation?

There could be also a lot of application specific usages.

sindresorhus: Yes, that’s a good example.

Giotino:
> Can you also add a `util.deprecate` call if `family` option is used?

According to the Node.JS documentation `util.deprecrate` is made for functions.
https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

sindresorhus:
There are benefits to `util.deprecate`. It only prints the warning once and it's also possible to silence it. Maybe we could make an utility that works the same way?

> [...] that will emit a DeprecationWarning using the 'warning' event. The warning will be emitted and printed to stderr the first time the returned function is called. After the warning is emitted, the wrapped function is called without emitting a warning.

Giotino:
I'm working on the "deprecation utilty" in this PR #1255
When it's ok I'll merge it here.

readme.md
```diff
@@ -499,6 +499,19 @@ An instance of [`CacheableLookup`](https://github.com/szmarczak/cacheable-lookup
 **Note:** This should stay disabled when making requests to internal hostnames such as `localhost`, `database.local` etc.\
 `CacheableLookup` uses `dns.resolver4(..)` and `dns.resolver6(...)` under the hood and fall backs to `dns.lookup(...)` when the first two fail, which may lead to additional delay.
 
+###### dnsIpVersion
```
szmarczak, line 502:
```suggestion
###### lookupIpVersion
```

@sindresorhus Maybe this?

readme.md
```diff
@@ -499,6 +499,19 @@ An instance of [`CacheableLookup`](https://github.com/szmarczak/cacheable-lookup
...
+Type: `'auto' | 'ipv4' | 'ipv6'`\
+Default: `'auto'`
+
+Indicates which DNS record family to use.\
+Values:
+ - `auto`: IPv4 (if present) or IPv6
+ - `ipv4`: only IPv4
+ - `ipv6`: only IPv6
+
+Note: if you are using the undocumented option `family`, `dnsIpVersion` will override it.
```
szmarczak, line 513: Is there a deprecation warning?

source/core/utils/dns-ip-version.ts
```diff
@@ -0,0 +1,24 @@
...
+export const isDnsIpVersion = (value: unknown): boolean => {
+	if (typeof value === 'string') {
+		return value in conversionTable;
+	}
+
+	return false;
+};
+
+export const dnsIpVersionToFamily = (dnsIpVersion: DnsIpVersion): DnsIpFamily => {
+	if (isDnsIpVersion(dnsIpVersion)) {
```
szmarczak, line 19:
```suggestion
	if (dnsIpVersion in conversionTable) {
```

readme.md
Giotino: > Is there a deprecation warning?

I'm going to add it when we're fine with `deprecation-warining.ts` from #1255

source/core/utils/dns-ip-version.ts
```diff
@@ -16,7 +16,7 @@ export const isDnsIpVersion = (value: unknown): boolean => {
 };
 
 export const dnsIpVersionToFamily = (dnsIpVersion: DnsIpVersion): DnsIpFamily => {
-	if (isDnsIpVersion(dnsIpVersion)) {
+	if (dnsIpVersion in conversionTable) {
```
szmarczak, line 19: The function above can be removed now, yup?

Giotino: No, because it's used in the assert inside the `normalizeArguments` function.

szmarczak: Ok. Then can you change this if to use `isDnsIpVersion` back again and to define

```
export const isDnsIpVersion = (value: unknown): boolean => {
	return value in conversionTable;
};
```

?

szmarczak: The `typeof` check is unnecessary I think.

readme.md
Giotino: Any news about this?

sindresorhus: I would go with `dnsLookupIpVersion`. It's verbose, but it's not a commonly-used option, so I think clarity is better.

sindresorhus:
This looks good except for the missing deprecation warning.

source/core/utils/dns-ip-version.ts
```diff
@@ -7,14 +7,14 @@ const conversionTable = {
...
-export const dnsIpVersionToFamily = (dnsIpVersion: DnsIpVersion): DnsIpFamily => {
-	if (isDnsIpVersion(dnsIpVersion)) {
-		return conversionTable[dnsIpVersion] as DnsIpFamily;
+export const dnsLookupIpVersionToFamily = (dnsLookupIpVersion: DnsLookupIpVersion): DnsIpFamily => {
+	if (isDnsLookupIpVersion(dnsLookupIpVersion)) {
+		return conversionTable[dnsLookupIpVersion] as DnsIpFamily;
 	}
 
-	throw new Error('Invalid DnsIpVersion');
+	throw new Error('Invalid DnsLookupIpVersion');
```
szmarczak, line 19: -> dnsLookupIpVersion option?

source/core/utils/dns-ip-version.ts
```diff
@@ -0,0 +1,20 @@
...
+export const isDnsLookupIpVersion = (value: any): boolean => {
+	return value in conversionTable;
+};
+
+export const dnsLookupIpVersionToFamily = (dnsLookupIpVersion: DnsLookupIpVersion): DnsIpFamily => {
+	if (isDnsLookupIpVersion(dnsLookupIpVersion)) {
+		return conversionTable[dnsLookupIpVersion] as DnsIpFamily;
+	}
+
+	throw new Error('Invalid DnsLookupIpVersion');
```
sindresorhus, line 19:
```suggestion
	throw new Error('Invalid `dnsLookupIpVersion` option value`');
```

test/http.ts
```diff
@@ -1,11 +1,27 @@
...
+	if (address === undefined) {
+		return response.end();
+	}
+
+	// IPv4 address mapped to IPv6
+	if (address === '::ffff:127.0.0.1') {
+		response.end('127.0.0.1');
+	} else {
+		response.end(address);
+	}
```
sindresorhus, line 22:
```suggestion
	// IPv4 address mapped to IPv6
	response.end(address === '::ffff:127.0.0.1' ? '127.0.0.1' : address);
```

source/core/utils/dns-ip-version.ts
Giotino: That function should be generic, not related to `dnsLookupIpVersion`.
I think it's better to catch the error and throw "Invalid `dnsLookupIpVersion` option value" where `dnsLookupIpVersionToFamily` is used:

```
requestOptions.family = dnsLookupIpVersionToFamily(options.dnsLookupIpVersion);
```
```
try {
  requestOptions.family = dnsLookupIpVersionToFamily(options.dnsLookupIpVersion);
} catch {
  throw new Error('Invalid `dnsLookupIpVersion` option value');
}
```

sindresorhus: 👍

sindresorhus, line 19: At least make it:

```suggestion
	throw new Error('Invalid DNS lookup IP version');
```

test/http.ts
```diff
@@ -255,3 +267,51 @@ test('does not destroy completed requests', withServer, async (t, server, got) =
...
+	server.get('/ok', echoIp);
+
+	try {
+		const response = await got('ok', {
+			dnsLookupIpVersion: 'ipv6'
+		});
+
+		t.true(isIPv6(response.body));
+	} catch {
+		// Assumes that IPv6 is not supported
```
sindresorhus, line 314: This is fragile as it might hide real error when IPv6 **is** supported. Is there any way we could detect if the system supports IPv6 and conditionalize the test itself instead?

Giotino: Something like this should work, I tested it on Windows 10.
```js
const os = require('os');
const ifaces = os.networkInterfaces();

let ipv6 = false;
for (const ifname in ifaces) {
  const iface = ifaces[ifname];
  for(const addr of iface) {
    if ('IPv6' === addr.family && addr.internal === true) {
      ipv6 = true;
      break;
    }
  }
  if (ipv6) break;
}
```

test/http.ts
```diff
@@ -269,16 +290,15 @@ test('does not destroy completed requests', withServer, async (t, server, got) =
 });
 
 test('IPv6 request', withServer, async (t, server) => {
+	if (!IPv6supported) {
```
szmarczak, line 293: The `if` should be a parent for these two tests.

test/http.ts
```diff
@@ -7,6 +7,27 @@ import getStream = require('get-stream');
 import pEvent = require('p-event');
 import got, {HTTPError, UnsupportedProtocolError} from '../source';
 import withServer from './helpers/with-server';
+import os = require('os');
+
+const ifaces = os.networkInterfaces();
+let IPv6supported = false;
```
szmarczak, line 13:
```suggestion
const IPv6supported = Object.values(os.networkInterfaces()).some(entry => !entry.internal && entry.family === 'IPv6');
```

?

Giotino: That's not the same, an equivalent code would be:
```
const IPv6supported = Object.values(os.networkInterfaces()).some(iface => iface?.some(addr => !addr.internal && addr.family === 'IPv6'));
```

szmarczak:
https://travis-ci.com/github/sindresorhus/got/jobs/341303159#L480

```
  Unhandled rejection in test/http.ts

  › ClientRequest.<anonymous> (dist/source/core/index.js:72:256)

  › ClientRequest.origin.emit (node_modules/@szmarczak/http-timer/dist/source/index.js:39:20)

  › emitErrorNT (internal/streams/destroy.js:100:8)

  › emitErrorCloseNT (internal/streams/destroy.js:68:3)
```

szmarczak:
the stack trace is mangled so you have to run the tests without `nyc`

Giotino:
There was a race condition on the two `await new Promise` (in `http` and `https`) used for the tests related to `DeprecationWarning`.
The promise could have been resolved before the request was resolved, resulting in the destruction of the web server while the request was still running.

sindresorhus:
Very nice work, @Giotino 👌

szmarczak:
@sindresorhus Maybe let's use numbers but `undefined` means `'auto'`?

sindresorhus:
Yeah, that would be nicer.

PR #1266, opened 2020-05-14
Make `calculateDelay` promisable

### LINKED ISSUE #1262: Support for async calculateDelay

Being able to use an async function to calculate the retry time.

#### Describe the feature
`calculateDelay` should be able to return a promise.

Ex:
```
async function calculateDelay ({ computedValue }) {
  const factor = await quantifyNetworkCongestion();
  return computedValue  * (1 + 9 * factor) ;
}
```

BTW, in theory with an async calculateDelay  we can also use: 
```
async function calculateDelay () {
  await waitUntilItIsTimeToRetry();
  return 1;
}
``` 

#### Checklist

Giotino:
#### Checklist

Fixes #1262


szmarczak:
Excellent! This just needs tests :)

Also Node.js 14.2.0 is failing somehow:

```
  Uncaught exception in test/retry.ts
  RequestError: Cannot call end after a stream was destroyed
  › PromisableRequest._destroy (dist/source/core/index.js:96:618)
  › PromisableRequest.destroy (internal/streams/destroy.js:36:8)
  › errorOrDestroy (internal/streams/destroy.js:146:12)
  › dist/source/core/index.js:94:579
  › endRequest (dist/source/core/index.js:94:221)
  › PromisableRequest._final (dist/source/core/index.js:94:706)
```

Giotino:
That issue is less trivial than it seems.
It's a problem with the `request` destruction.

https://github.com/sindresorhus/got/blob/master/source/as-promise/index.ts#L151,L224

```
request.once('error', async (error: RequestError) => {
  // Some slow logic
  request.destroy();
});

```

After the `error` event, the request fire the `response` event, but this is prevented by `request.destroy()`.
The problem in my PR is that being the `error` listener async the emitter sees it as immediately completed and goes on with the `response` event because `request.destroy()` has not been called yet.

Giotino:
The only thing that came to my mind was restoring the order of the two events (`error` and `response`) in the async environment. I used a sync barrier (https://github.com/sindresorhus/got/pull/1266/commits/ae24c2fa38524426b6cbba55cc591d3c7e2a3b87).

Note: the code needs some cleaup.

Giotino:
BTW, this commit is passing on my Travic CI (and also on my PC). Race conditions are really hard to debug...
https://travis-ci.org/github/Giotino/got/builds/687143936

source/as-promise/index.ts
```diff
@@ -20,6 +20,26 @@ const proxiedRequestEvents = [
 	'downloadProgress'
 ];
 
+// Thanks https://github.com/drpicox/async-barrier
+/* eslint-disable no-undef-init */
+function makeAsyncBarrier() {
```
szmarczak, line 25: There's no need to have this function...

Giotino: Look at the last commit, I tried to make something reusable (maybe it'll be needed in the future), also is cleaner than before.

source/as-promise/utils/event-listeners-reorderer.ts
```diff
@@ -0,0 +1,60 @@
+export default class EventListenerReorderer {
```
szmarczak, line 1: Is this really needed?

source/as-promise/utils/event-listeners-reorderer.ts
```diff
@@ -0,0 +1,60 @@
...
+	firstCalled = false;
+	waitPromise: Promise<any>;
+	waitResolve: any;
```
sindresorhus, line 4: Don't use `any`. Use a more specific type.

source/as-promise/utils/event-listeners-reorderer.ts
```diff
@@ -0,0 +1,60 @@
+export default class EventListenerReorderer {
+	firstCalled = false;
```
sindresorhus, line 2:
```suggestion
	isFirstCalled = false;
```

source/as-promise/utils/event-listeners-reorderer.ts
```diff
@@ -0,0 +1,60 @@
...
+	waitPromise: Promise<any>;
+	waitResolve: any;
+	waitCalled = false;
```
sindresorhus, line 5:
```suggestion
	isWaitCalled = false;
```

source/as-promise/utils/event-listeners-reorderer.ts
Giotino: I don't think it's really really needed, but I think is a lot better having that messy code (it can't be simplified too much) in a different file in order to not mess up `as-promise/index.ts`.

szmarczak: I'm talking if we need to use `EventListenerReorderer` at all.

szmarczak: Can't we just `request.socket.pause()` before the await and `request.socket.resume()` afterwards?

Giotino: > After the `error` event, the request fire the `response` event, but this is prevented by `request.destroy()`.
> The problem in my PR is that being the `error` listener async, the emitter sees it as immediately completed and goes on with the `response` event because `request.destroy()` has not been called yet.

That's the problem.

> The only thing that came to my mind was restoring the order of the two events (`error` and `response`) in the async environment.

An that's a solution I found.

If you've a better solution please let me know, I really do not have any idea other than restoring the order of the two listeners.

Giotino: > Can't we just `request.socket.pause()` before the await and `request.socket.resume()` afterwards?

I'll give it a try

Giotino: > > Can't we just `request.socket.pause()` before the await and `request.socket.resume()` afterwards?
> 
> I'll give it a try

```
  2 tests failed
  1 known failure
  1 test skipped
  2 unhandled rejections
  3 uncaught exceptions
```

According to the Node.JS documentation
> socket.pause() Pauses the reading of data

The thing is that we need to "pause the events", when `error` is emitted I think all the data has been already read, so it not "paused" the response event.

source/as-promise/index.ts
```diff
@@ -216,15 +217,16 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 				retryCount--;
 
 				if (error instanceof HTTPError) {
-					// It will be handled by the `response` event
+					// The error will be handled by the `response` event
```
szmarczak, line 220: Simple as that :)

source/as-promise/index.ts
```diff
@@ -172,7 +176,7 @@ export default function asPromise<T>(options: NormalizedOptions): CancelableRequ
 							retryOptions: options.retry,
 							error,
 							computedValue: 0
-						})
+						}) as number
```
sindresorhus, line 179: Why `as number`?

szmarczak: Because it can be also `Promise<number>`

sindresorhus: But then `computedValue` is typed incorrectly, as it has to accept `number | Promise<number>`?

szmarczak: I'll fix that.

source/as-promise/calculate-retry-delay.ts
```diff
@@ -6,13 +6,15 @@ import {
 	RetryFunction
 } from './types';
 
+type Returns<T extends (...args: any) => any, V> = (...args: Parameters<T>) => V;
```
sindresorhus, line 9:
```suggestion
type Returns<T extends (...args: any) => unknown, V> = (...args: Parameters<T>) => V;
```

Can this be `unknown`?

szmarczak:
@sindresorhus Is this looking good? I'd like to release a patch version before any API depracations.

szmarczak:
Nice one! 🙌

PR #1267, opened 2020-05-15
Upgrade to TypeScript 3.9

sindresorhus:


sindresorhus:
I think this is caused by https://github.com/microsoft/TypeScript/issues/38540

BenSjoberg:
The linked TypeScript bug also causes the current version of got to fail when using TS 3.9. The issue occurs when you mix the default import with type imports (i.e. `import got, { HTTPError } from 'got'`) or when using a dynamic import (i.e. `await import('got')`)

A workaround I found is to add `module.exports.__esModule = true` after [this line](https://github.com/sindresorhus/got/blob/master/source/index.ts#L125). TypeScript sets __esModule to true on the exported module by default, but setting `module.exports = got` overwrites it.

The right answer might be to just wait for TypeScript to fix the problem, but no one from the team has commented on the bug yet, so it might be a while. :(

DanielRosenwasser:
We should have a fix at https://github.com/microsoft/TypeScript/pull/38808. If you can pick up the build that will be produced at the bottom of the page and try it out, that'd really help give us some confidence that we can back-port it to 3.9.

sindresorhus:
I tried it in https://github.com/sindresorhus/got/pull/1267/commits/d6dae82c3b9514d8f7206f3bfad83801815c99a2, and seems like it doesn't fix our issue, but it doesn't cause any other problems either.

source/core/index.ts
```diff
@@ -1215,7 +1215,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 					}
 
 					if (options.username || options.password) {
+						// @ts-ignore
```
szmarczak, line 1218: What's the error here?

source/core/index.ts
```diff
@@ -1358,6 +1361,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			Object.assign(options, urlToOptions(url));
 
 			// `http-cache-semantics` checks this
+			// @ts-ignore
```
szmarczak, line 1364: And here?

source/core/index.ts
```diff
@@ -1467,6 +1471,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		// Prepare plain HTTP request options
 		options[kRequest] = realFn as HttpRequestFunction;
 		delete options.request;
+
+		// @ts-ignore
```
szmarczak, line 1475: And here too?

source/create.ts
```diff
@@ -120,13 +120,14 @@ const create = (defaults: InstanceDefaults): Got => {
...
 
-		// TODO: remove this in Got 12
+		// TODO: Remove this in Got 12.
 		if (is.plainObject(url)) {
 			const mergedOptions = {
 				...url as Options,
 				...options
 			};
 
+			// @ts-expect-error
```
szmarczak, line 130: Why is this needed?

source/create.ts
```diff
@@ -197,6 +198,7 @@ const create = (defaults: InstanceDefaults): Got => {
 
 	// Pagination
 	const paginateEach = (async function * <T, R>(url: string | URL, options?: OptionsWithPagination<T, R>) {
+		// @ts-expect-error
```
szmarczak, line 201: This too?

source/create.ts
sindresorhus: It didn't bother wasting time on it as that code is going away in the next major anyway.

Error:

```
source/create.ts:130:32 - error TS2322: Type '(string & Record<string, unknown>) | (URL & Record<string, unknown>)' is not assignable to type 'Options | Defaults | undefined'.
  Type 'string & Record<string, unknown>' is not assignable to type 'Options | Defaults | undefined'.
    Type 'string & Record<string, unknown>' is not assignable to type 'Options'.
      Types of property 'search' are incompatible.
        Type '{ (regexp: string | RegExp): number; (searcher: { [Symbol.search](string: string): number; }): number; }' is not assignable to type 'string'.

130  		setNonEnumerableProperties([url, options], mergedOptions);
```

source/create.ts
sindresorhus: I have no idea how to fix it. I just punted it as it's more important to get TS upgraded:

```
source/create.ts:201:51 - error TS2345: Argument of type 'Merge<Options, PaginationOptions<T, R>> | undefined' is not assignable to parameter of type 'Options | undefined'.
  Type 'Merge<Options, PaginationOptions<T, R>>' is not assignable to type 'Options | undefined'.
    Type 'Merge<Options, PaginationOptions<T, R>>' is not assignable to type 'Options'.
      Types of property 'pagination' are incompatible.
        Type '{ transform?: ((response: Response<R>) => T[] | Promise<T[]>) | undefined; filter?: ((item: T, allItems: T[], currentItems: T[]) => boolean) | undefined; ... 4 more ...; stackAllItems?: boolean | undefined; } | undefined' is not assignable to type '{ transform?: ((response: Response<unknown>) => unknown[] | Promise<unknown[]>) | undefined; filter?: ((item: unknown, allItems: unknown[], currentItems: unknown[]) => boolean) | undefined; paginate?: ((response: Response<unknown>, allItems: unknown[], currentItems: unknown[]) => false | Options) | undefined; should...'.
          Type '{ transform?: ((response: Response<R>) => T[] | Promise<T[]>) | undefined; filter?: ((item: T, allItems: T[], currentItems: T[]) => boolean) | undefined; ... 4 more ...; stackAllItems?: boolean | undefined; }' is not assignable to type '{ transform?: ((response: Response<unknown>) => unknown[] | Promise<unknown[]>) | undefined; filter?: ((item: unknown, allItems: unknown[], currentItems: unknown[]) => boolean) | undefined; paginate?: ((response: Response<unknown>, allItems: unknown[], currentItems: unknown[]) => false | Options) | undefined; should...'.
            Types of property 'transform' are incompatible.
              Type '((response: Response<R>) => T[] | Promise<T[]>) | undefined' is not assignable to type '((response: Response<unknown>) => unknown[] | Promise<unknown[]>) | undefined'.
                Type '(response: Response<R>) => T[] | Promise<T[]>' is not assignable to type '(response: Response<unknown>) => unknown[] | Promise<unknown[]>'.
                  Types of parameters 'response' and 'response' are incompatible.
                    Type 'Response<unknown>' is not assignable to type 'Response<R>'.
                      Type 'unknown' is not assignable to type 'R'.
                        'R' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.

201  	let normalizedOptions = normalizeArguments(url, options, defaults.options);
     	                                                ~~~~~~~


Found 1 error.
```

I'll add a TODO comment.

sindresorhus: TS 3.9 introduced a lot of issues, and it seems many will be fixed in TS 4, so I suggest ignoring it until then.

source/core/index.ts
sindresorhus: Some weird TS error, but seems it was fixed in later TS patch versions as I cannot reproduce it now.

source/create.ts
szmarczak: https://github.com/microsoft/TypeScript/issues/33436

szmarczak: Simplest repro:

```ts
interface Options<T = unknown> {
    fn: (value: T) => void
}

const normalizeOptions = (options: Options): void => {};

function someOtherFunction<T>(options: Options<T>) {
    normalizeOptions(options);
}
```

szmarczak: Replacing `unknown` with `any` seems to fix the issue.

szmarczak: Interestingly the example reproduces the error in all TS 3.x versions.

PR #1272, opened 2020-05-16
Integration test

### LINKED ISSUE #1056: Integration test
We should have a test that install Got with only production dependencies `npm i --production` and then try to require it and do a request, just to ensure nothing broke in production. For example, that we didn't forget a dependency (https://github.com/sindresorhus/got/issues/1055).

Giotino:
This is just a draft about @sindresorhus proposal (#1056).

Actually this is really simple, after all the tests it removes the dev dependencies and make a request to https://google.com, if it fails (or if something's wrong with the dependencies) it exit with an exit code != 0 resulting in a Travis CI failure.

Working: https://travis-ci.org/github/Giotino/got/builds/687760094
Non working: https://travis-ci.org/github/Giotino/got/builds/687760585

Fixes #1056


integration-test/example.js
```diff
@@ -0,0 +1,11 @@
+const got = require('../dist/source');
+
+(async () => {
+	try {
+		await got('https://google.com');
```
szmarczak, line 5: Can you create a local server instead? We should avoid querying 3rd party servers.

szmarczak: Actually for an integration test... I don't know.

Giotino: I think these tests should be as close as possible to real usages.

package.json
```diff
@@ -10,10 +10,12 @@
...
 	},
 	"scripts": {
+		"build": "del-cli dist && tsc",
 		"test": "xo && tsc --noEmit && nyc --reporter=html --reporter=text ava",
+		"coverage": "nyc report --reporter=text-lcov | coveralls",
+		"prepare": "npm run build",
 		"release": "np",
-		"build": "del-cli dist && tsc",
-		"prepare": "npm run build"
+		"clean": "rimraf dist coverage .nyc_output"
```
sindresorhus, line 18:
```suggestion
		"clean": "del-cli dist coverage .nyc_output"
```

package.json
```diff
@@ -10,10 +10,12 @@
 		"node": ">=10.19.0"
 	},
 	"scripts": {
+		"build": "del-cli dist && tsc",
 		"test": "xo && tsc --noEmit && nyc --reporter=html --reporter=text ava",
+		"coverage": "nyc report --reporter=text-lcov | coveralls",
```
sindresorhus, line 15: I would prefer to keep this in the Travis file.

integration-test/example.js
szmarczak: Can we use 3rd-party packages for this? They use Got more extensively I think.

szmarczak: Like https://github.com/yarnpkg/berry or https://github.com/renovatebot/renovate or https://github.com/MONEI/Shopify-api-node/ or (it goes on and on).

szmarczak:
IMO this is a good start and we can add more complicated integration tests later.

sindresorhus:
I think the integration test would be a good opportunity to use GitHub Actions.

sindresorhus:
@Giotino Anything blocking you to finish this or are you just busy?

Giotino:
> @Giotino Anything blocking you to finish this or are you just busy?

Yeah, I was busy and since this is not critical to Got I delayed it a little bit.

> I think the integration test would be a good opportunity to use GitHub Actions.

On the GitHub action side, I think we should use the same workflow as TravisCI (without coveralls).
I'm going to add the Github workflow file to this PR.

sindresorhus:
Ok. No worries. 👍🏻

szmarczak:
@Giotino Friendly ping :)

sindresorhus:
> I think the integration test would be a good opportunity to use GitHub Actions.

This was done in https://github.com/sindresorhus/got/commit/35f5822c2029b0075c381663d27aa39608dc0a4d

Issue #1122, opened 2020-03-16
Can’t use type with responseType option


sunknudsen:
#### Describe the bug

 - Node.js version: v12.14.1
 - OS & version: macOS Mojave 10.14.6

#### Actual behavior

I believe this issue is a TypeScript type declaration bug.

I am building on top of got and having issues with the `responseType` option.

#### Expected behavior

The following code block should not throw type error.

#### Code to reproduce

```typescript
import got from "got"

export type ResponseType = "text" | "json" | "buffer"

const send = async function(responseType: ResponseType) {
  console.log(await got({
    url: "https://api.example.net",
    method: "GET",
    responseType: responseType
  }))
}
```

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

sindresorhus:
What is the type error?

sunknudsen:
@sindresorhus 

```shell
The last overload gave the following error.
  Argument of type '{ method: Method; url: string; searchParams: string; json: { [key: string]: any; }; headers: Headers; responseType: ResponseType; hooks: { beforeRequest: (() => Promise<void>)[]; beforeRetry: ((options: NormalizedOptions, error: GeneralError, retryCount: number) => Promise<...>)[]; }; }' is not assignable to parameter of type 'string | Merge<Merge<RequestOptions, Merge<GotOptions, URLOptions>>, { isStream: true; }>'.
    Type '{ method: Method; url: string; searchParams: string; json: { [key: string]: any; }; headers: Headers; responseType: ResponseType; hooks: { beforeRequest: (() => Promise<void>)[]; beforeRetry: ((options: NormalizedOptions, error: GeneralError, retryCount: number) => Promise<...>)[]; }; }' is not assignable to type 'Merge<Merge<RequestOptions, Merge<GotOptions, URLOptions>>, { isStream: true; }>'.
      Property 'isStream' is missing in type '{ method: Method; url: string; searchParams: string; json: { [key: string]: any; }; headers: Headers; responseType: ResponseType; hooks: { beforeRequest: (() => Promise<void>)[]; beforeRetry: ((options: NormalizedOptions, error: GeneralError, retryCount: number) => Promise<...>)[]; }; }' but required in type '{ isStream: true; }'.
```

szmarczak:
Possibly a duplicate of #954

PR #1276, opened 2020-05-19
Provide an overload for unknown `responseType`

kschat:
#### Checklist

- [ ] I have included some tests (more on why this is unchecked below).

Wrapping `got` is difficult with the current function overloads. The changes made in #1051 made things much better, but given a `responseType` of `'text' | 'buffer' | 'json'` in the following call:

```ts
const responseType = 'text' as 'text' | 'buffer' | 'json';
const result = await got({
  url: 'github.com',
  responseType,
});
```

`got` will return `CancelableRequest<unknown> | Request` (or just `unknown` when `await`ed) instead of `CancelableRequest<Response<unknown>>`. This change adds an overload that catches the case where `responseType` isn't statically known, but we do know that a `Response` object will be returned.

I haven't added any tests for this PR since this is a type update and I didn't see any tests for types. Given the complexity of the types used by `got`, though, I'm happy to add some if desired. I've had good luck using [conditional-type-checks](https://github.com/dsherret/conditional-type-checks) for testing types in the past.

Related issues:
#954 
#1122

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -44,7 +44,7 @@ export interface ExtendOptions extends Options {
 export type OptionsOfTextResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType?: 'text'}>;
 export type OptionsOfJSONResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType?: 'json'}>;
 export type OptionsOfBufferResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType: 'buffer'}>;
-export type OptionsOfResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false}>;
+export type OptionsOfUnknownResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false}>;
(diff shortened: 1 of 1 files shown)


source/types.ts
```diff
@@ -44,6 +44,7 @@ export interface ExtendOptions extends Options {
 export type OptionsOfTextResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType?: 'text'}>;
 export type OptionsOfJSONResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType?: 'json'}>;
 export type OptionsOfBufferResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false; responseType: 'buffer'}>;
+export type OptionsOfResponseBody = Merge<Options, {isStream?: false; resolveBodyOnly?: false}>;
```
szmarczak, line 47: `OptionsOfResponseBody` -> `OptionsOfUnknownResponseBody`

szmarczak:
@kschat Can you also submit a test for this particular use case?

kschat:
@szmarczak sorry for the delayed response. I'm happy to open another PR with a test if that's still needed.

szmarczak:
Sure, that would be great.

PR #1289, opened 2020-05-26
Do not wait for `ready` event if the file descriptor is already opened

### LINKED ISSUE #1280: Got hangs if passed a stream which has already emitted 'open'
#### Describe the bug

 - Node.js version: 14.0.0
 - OS & version: Linux mint 19
 - got version: 11.1.4

Hi, I've got some code where a `fs.ReadStream` is created, and passed to Got some time later in an async callback. It seems Got will hang if the stream has already emitted the `open` event before being passed in. Any chance this can be fixed, or is this an unsupported scenario?

#### Code to reproduce

```js
const got = require('got');
const fs = require('fs');

const fstream = fs.createReadStream('test.txt');
fstream.once('open', async () => {
	// Prevent node from exiting early due to empty event loop.
	setTimeout(() => {}, 10000);

	console.log('Starting request...');
	const response = await got({
		url: 'https://www.google.com',
		method: 'POST',
		body: fstream,
	});
	// This is never executed.
	console.log(response.statusCode);
});
```
#### Checklist

Giotino:
got waits for the `open` event if the body is an `fs.ReadStream` even if the fd is already opened.

With this PR I added a check `ReadStream.pending` in order to immediately resolve `waitForOpenFile` if the fd is already opened.

#### Checklist

Fixes #1280

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -262,6 +262,8 @@ const waitForOpenFile = async (file: ReadStream): Promise<void> => new Promise((
 		reject(error);
 	};
 
+	if (!file.pending) resolve();
+
 	file.once('error', onError);
 	file.once('open', () => {
 		file.off('error', onError);
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -267,7 +267,7 @@ const waitForOpenFile = async (file: ReadStream): Promise<void> => new Promise((
 	}
 
 	file.once('error', onError);
-	file.once('open', () => {
+	file.once('ready', () => {
 		file.off('error', onError);
 		resolve();
 	});
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -262,7 +262,9 @@ const waitForOpenFile = async (file: ReadStream): Promise<void> => new Promise((
 		reject(error);
 	};
 
-	if (!file.pending) resolve();
+	if (!file.pending) {
+		resolve();
+	}
 
 	file.once('error', onError);
 	file.once('open', () => {
--- a/test/post.ts
+++ b/test/post.ts
@@ -4,6 +4,7 @@ import fs = require('fs');
 import path = require('path');
 import test from 'ava';
 import delay = require('delay');
+import pEvent = require('p-event');
 import {Handler} from 'express';
 import getStream = require('get-stream');
 import toReadableStream = require('to-readable-stream');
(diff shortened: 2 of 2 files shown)


source/core/index.ts
```diff
@@ -262,6 +262,10 @@ const waitForOpenFile = async (file: ReadStream): Promise<void> => new Promise((
 		reject(error);
 	};
 
+	if (!file.pending) {
+		resolve();
+	}
+
```
sindresorhus, line 268: Slightly unrelated, but shouldn't we listen to `ready` instead of `open`?

```
Event: 'ready'#

Added in: v9.11.0
Emitted when the fs.ReadStream is ready to be used.

Fires immediately after 'open'.
```

Probably will not change much, but maybe worth it just in case.

Giotino: I don't know why there's that code, but I think @szmarczak wanted to check if there was any error opening the file (eg. file does not exists), so `open` seems to be the best choice.

BTW: reading this "Fires immediately after 'open'." I think `open` and `ready` are completely equivalent, I think it's some guy at Node.JS that just anted to mess with the users.

sindresorhus: I’m asking as `.pending` docs mention the `ready` event, not `open`.

Giotino: The Node.JS documentation make no sense:
`open`: Emitted when the fs.ReadStream's file descriptor has been opened.
`ready`: Emitted when the fs.ReadStream is ready to be used.
`readStream.pending`: "This property is true if the underlying file has not been opened yet, i.e. before the 'ready' event is emitted."

`pending` should be "This property is true if the underlying file has not been opened yet, i.e. before the 'open' event is emitted."

Looking at the code:
`pending` is defined as `this.fd === null` (https://github.com/nodejs/node/blob/master/lib/internal/fs/streams.js#L265,L268)
The `open` event is fired right after the assignation of `fd` and the `ready` event is fired right after the `open` event.
(https://github.com/nodejs/node/blob/master/lib/internal/fs/streams.js#L151,L153)

sindresorhus: I would recommend opening a Node.js issue about it. The docs should be clear about which event to use for what. It's currently very unclear.

szmarczak: Let's for with `ready` to be consistent.

test/post.ts
```diff
@@ -318,6 +318,26 @@ test('body - file read stream', withServer, async (t, server, got) => {
...
 });
 
+test('body - file read stream, wait for `ready` event', withServer, async (t, server, got) => {
+	server.post('/', defaultEndpoint);
+
+	const fullPath = path.resolve('test/fixtures/ok');
+	const toSend = await getStream(fs.createReadStream(fullPath));
+	const ifStream = fs.createReadStream(fullPath);
+
+	await new Promise(resolve => ifStream.on('ready', () => {
```
sindresorhus, line 328: Use `p-event`.

PR #1296, opened 2020-05-29
Server side aborted request hang with node 14.3.0

### LINKED ISSUE #1295: Server side aborted request in stream mode hang with node 14.3.0
#### Describe the bug

 - Node.js version: 14.3.0
 - OS & version: Ubuntu 19

Server side aborted requests in stream mode conduce got to hang with node >= 14.3.0. In fact the `aborted` event from the `IncomingMessage` is not transmitted or transformed to an error in the stream got returns.

#### Code to reproduce

```js
const http = require('http');
const got = require('got');

// http server that aborts a request
const server = http.createServer((req, res) => {
	res.on('close', () => {
		console.log('server close')
	})

	res.writeHead(200, { 'Content-Type': 'text/plain' });
	res.write('chunk 1');

	setTimeout(() => res.write('chunk 2'), 1000);
	setTimeout(() => res.destroy(), 2000);
});
server.listen(8000);

// got request
(async () => {
	try {
		const gotOptions = {timeout: {socket: 2000, request: 3000}, isStream: true};
		const httpRequest = got('http://localhost:8000', gotOptions);
		for await (const chunk of httpRequest) {
			console.log(`client data: aborted=${httpRequest.aborted}, data=${chunk.toString()}`)
		}
		console.log(`client stream end`);
	} catch (e) {
		console.log(`client error: error=${e}`);
	}
	console.log(`end`);
})();
```

##### 14.3.0
```
client data: aborted=false, data=chunk 1
client data: aborted=false, data=chunk 2
server close
```
##### 14.2.0
```
client data: aborted=false, data=chunk 1
client data: aborted=false, data=chunk 2
client error: error=ReadError: The server aborted the pending request
end
server close
```

#### Comments
With node 14.2.0 we can see when the request is aborted that an error in thrown from the `for await` loop. We can see the `end` message too signaling the request is terminated. With node 14.3.0, no error is thrown and the stream is not terminated, we are stuck in the `for await` loop

#### Workaround
It's possible to use this workaround to restore the previous behavior (but we get a `got.RequestError` and not more a `got.ReadError`):
```js
const {once} = require('events');
const got = require('got');

(async () => {
	try {
		const gotOptions = {timeout: 3000, isStream: true};
		const httpRequest = got('http://localhost:8000', gotOptions);
		const [httpResponse] = await once(httpRequest, 'response');
		httpResponse.on('aborted', () => {
			httpRequest.destroy(new Error('The server aborted the pending request'));
		});
		for await (const chunk of httpRequest) {
			console.log(`client data: aborted=${httpRequest.aborted}, data=${chunk.toString()}`)
		}
		console.log(`client stream end`);
	} catch (e) {
		console.log(`client error: error=${e}`);
	}
	console.log(`end`);
})();
```

#### Checklist

Giotino:
Currently this is a solution that "works", but to me it doesn't seems right.

Fixes #1295

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1596,12 +1596,6 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 	_destroy(error: Error | null, callback: (error: Error | null) => void): void {
 		if (kRequest in this) {
 			this[kCancelTimeouts]!();
-
-			// TODO: Remove the next `if` when these get fixed:
-			// - https://github.com/nodejs/node/issues/32851
-			if (!this[kResponse]?.complete) {
-				this[kRequest]!.destroy();
-			}
 		}
 
 		if (error !== null && !is.undefined(error) && !(error instanceof RequestError)) {
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1057,10 +1057,6 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		});
 
 		response.once('aborted', () => {
-			if (this.aborted) {
-				return;
-			}
-
 			this._beforeError(new ReadError({
 				name: 'Error',
 				message: 'The server aborted the pending request'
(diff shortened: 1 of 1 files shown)


szmarczak:
You need to remove this if https://github.com/sindresorhus/got/blob/2c8fe196361e5fc3c063b56b816f916999d60c09/source/core/index.ts#L1492 (but the content should be present). And just put `if (this.destroyed) { return; }` at the beginning of `_beforeError`.

szmarczak:
^ just to prevent calling before error hooks if the error is not emitted.

Giotino:
Like this?https://github.com/sindresorhus/got/pull/1296/commits/601fe10f1705db21f185f91c70f366486d04fa96

PR #1298, opened 2020-05-31
Add options to customize parsing/stringifying JSON

### LINKED ISSUE #1274: parseJson option
As discussed there: https://github.com/sindresorhus/ky/issues/254#issuecomment-629613861

Option should allow to pass a user-function that will perform JSON parsing.

Use cases:

1. Use `bourne` to mitigate the hyped prototype pollution.
2. Parse JSON with `reviver`

Example API:

```typescript

type ParseJsonFunction = <OUTPUT = unknown>(text: string) => OUTPUT

const myGot = got.extend({
  parseJson: text => bourne(text),
})
```

Should be available both via `.extend()` and as an option to each request.

Giotino:
Custom function to parse JSON responses.

#### Checklist

Fixes #1274


source/core/index.ts
```diff
@@ -224,6 +227,7 @@ export interface Defaults {
 	allowGetBody: boolean;
 	https?: HTTPSOptions;
 	methodRewriting: boolean;
+	parseJson?: ParseJsonFn;
```
szmarczak, line 230:
```suggestion
	parseJson: ParseJsonFn;
```

szmarczak, line 230:
```suggestion
	parseJson: ParseJsonFn;
```

szmarczak:
@Giotino Can you also do `stringifyJson`?

szmarczak:
IMO this is good to merge.

szmarczak:
Actually I'm having second thoughts. This also adds more complexity. What's against having

```js
const {body} = await got(...);
const parsed = JSON.parse(body, parser);
```

Giotino:
I thought so, but with `parseJson` you can have an extended got that always behave the same without having to explicitly use your parser every time.

It's not a "must have"  feature, but since someone showed interest in that it means that it might be nice to have. Also I don't think it adds that much complexity.

Giotino:
Maybe we should ear more opinions before proceeding.

readme.md
```diff
@@ -317,6 +317,23 @@ const body = await got(url).json();
...
 ```
 
+###### parseJson
+
+Type: `(text: string) => unknown`\
+Default: `(text) => JSON.parse(text)`
+
+Function used to parse JSON responses.
+
+Example:
```
sindresorhus, line 327: It's not immediately clear to the user what this example does different.

See https://github.com/sindresorhus/ky/pull/257/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R349-R367 for better docs.

readme.md
```diff
@@ -317,6 +317,23 @@ const body = await got(url).json();
 const body = await got(url, {responseType: 'json', resolveBodyOnly: true});
 ```
 
+###### parseJson
+
+Type: `(text: string) => unknown`\
+Default: `(text) => JSON.parse(text)`
```
sindresorhus, line 323:
```suggestion
Default: `(text: string) => JSON.parse(text)`
```

readme.md
```diff
@@ -317,6 +317,23 @@ const body = await got(url).json();
...
+
+Function used to parse JSON responses.
+
+Example:
+
+```js
```
sindresorhus, line 329:
```suggestion
```js
const got = require('got');
```

source/core/index.ts
```diff
@@ -118,6 +118,7 @@ type CacheableRequestFn = (
 ) => CacheableRequest.Emitter;
 
 type CheckServerIdentityFn = (hostname: string, certificate: DetailedPeerCertificate) => Error | void;
+export type ParseJsonFn = (text: string) => unknown;
```
sindresorhus, line 121:
```suggestion
export type ParseJsonFunction = (text: string) => unknown;
```

I know we haven't been consistent with this, but never too late to start.

Giotino: Ok, I'm also going to change the other `*Fn` types, is it ok?

kirillgroshkov:
> Actually I'm having second thoughts. This also adds more complexity. What's against having
> 
> ```js
> const {body} = await got(...);
> const parsed = JSON.parse(body, parser);
> ```

In our case we have ~50 methods that call API via `got(...).json()`. To add a reviver to them right now means adding it to 50 places and risking that we forget to add it to newly added methods. Having it built in the extended got means the change in only 1 place and safety to add more methods.

Same thinking as why got.extend function and hooks exist in the first place. Without it - it'll need copy/pasting the hooks to every usage.

szmarczak:
Hmm... @kirillgroshkov is right. So `stringifyJson` is also welcome, but unnecessary. It's up to you :)

Giotino:
Examples needs to be fixed/improved.

@sindresorhus @szmarczak
While working on the documentation I noted that the examples are incosistent.
Examples of examples :)
```js
const got = require('got');

await got();
```

```js
const got = require('got');

(async () => {
  await got();
})();
```
```js
await got();
```

I think that `const got = require('got');` and the async wrapper are not needed, I can work on a PR normalizing all the examples.

szmarczak:
> I think that const got = require('got'); and the async wrapper are not needed

@sindresorhus has said a while ago that these examples should be 100% functional. Copy-paste and it should work.

source/core/index.ts
```diff
@@ -120,6 +120,7 @@ type CacheableRequestFunction = (
 
 type CheckServerIdentityFunction = (hostname: string, certificate: DetailedPeerCertificate) => Error | void;
 export type ParseJsonFunction = (text: string) => unknown;
+export type StringifyJsonFunction = (object: any) => string;
```
sindresorhus, line 123:
```suggestion
export type StringifyJsonFunction = (object: unknown) => string;
```

Makes no practical difference, but we prefer `unknown` whenever possible. Same with the other places `any` is used.

readme.md
```diff
@@ -1,4 +1,4 @@
-<div align="center">
+Œ<div align="center">
```
sindresorhus: I have no idea what happened in this commit. Looks like the GitHub web UI completely borked up the commit.

sindresorhus:
https://github.com/sindresorhus/got/pull/1298/files#r433074774 still needs to be resolved.

sindresorhus:
> I think that const got = require('got'); and the async wrapper are not needed, I can work on a PR normalizing all the examples.

That would be great as a separate PR. We should normalize to **always** require Got and always use the async IIFE.

Giotino:
> https://github.com/sindresorhus/got/pull/1298/files#r433074774 still needs to be resolved.

Yeah, I'm aware of it, I was kinda busy so I delayed it a bit.

readme.md
Giotino: I can't figure out any meaningful example for `stringifyJson`, currently I added a random one that transforms all numbers in strings.

readme.md
```diff
@@ -326,13 +326,18 @@ Function used to parse JSON responses.
 
 Example:
 
+Using Bourne to prevent prototype pollution
```
szmarczak, line 329:
```suggestion
- Using Bourne to prevent prototype pollution:

```

readme.md
```diff
@@ -344,15 +349,22 @@ Function used to stringify JSON requests body.
 
 Example:
 
+All numbers as strings
```
szmarczak, line 352:
```suggestion
- All numbers as strings:

```

readme.md
szmarczak: A better example would be something that skips keys that begin with `_`.

sindresorhus: > A better example would be something that skips keys that begin with _.

Yes, in addition ot the current one with numbers.

sindresorhus:
Excellent 👌

PR #1305, opened 2020-06-05
Fix `response` event not being emitted on cache verify request

### LINKED ISSUE #1287: Request hangs up with GOT v11
#### Describe the bug

 - Node.js version: 12.16.3
 - OS & version: Debian Buster

#### Actual behavior
Request gets stuck while downloading.
...

#### Expected behavior
Request should either timeout or give correct response
...

#### Code to reproduce

```js

const agent = {
  http: new http.Agent({ keepAlive: true }),
  https: new https.Agent({ keepAlive: true })
};

var requestOptions = {
    timeout: 45000, // milliseconds
    responseType: 'buffer',
    retry: 0,
    maxRedirects: 5,
    dnsCache: false,
    headers: {
      'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:57.0) Gecko/20100101 Firefox/57.0',
      'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',
      'Accept-Language': 'en-US,en;q=0.5'
    },
    shared: false,
    agent,
    cache: cacheStore
  };

got.get("https://www.empiresuppliesonline.co.uk/ekmps/shops/empiresupplies/images/Waring-X-Prep-Kitchen-Blender-2523-p.jpg", requestOptions);

I am using https://www.npmjs.com/package/improved-cacheable-request as cache store.

Please note I download around 500 such URLs per minute but the issue majorly happens with this given domain name.

``` 

#### Checklist

Giotino:
~~That's only a workaround, a real solution is still WIP (this PR will be updated).~~

#### Checklist

Fixes #1287

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1296,16 +1296,17 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			// `http-cache-semantics` checks this
 			delete (options as unknown as NormalizedOptions).url;
 
+			let request: ClientRequest;
+
 			// This is ugly
 			const cacheRequest = cacheableStore.get((options as any).cache)!(options, response => {
 				const typedResponse = response as unknown as IncomingMessageWithTimings & {req: ClientRequest};
-				const {req} = typedResponse;
 
 				// TODO: Fix `cacheable-response`
 				(typedResponse as any)._readableState.autoDestroy = false;
 
-				if (req) {
-					req.emit('cacheableResponse', typedResponse);
+				if (request) {
+					request.emit('cacheableResponse', typedResponse);
 				}
 
 				resolve(typedResponse as unknown as ResponseLike);
@@ -1315,7 +1316,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			(options as unknown as NormalizedOptions).url = url;
 
 			cacheRequest.once('error', reject);
-			cacheRequest.once('request', resolve);
+			cacheRequest.once('request', r => {
+				request = r;
+				resolve();
+			});
 		});
 	}
 
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1318,7 +1318,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			cacheRequest.once('error', reject);
 			cacheRequest.once('request', r => {
 				request = r;
-				resolve();
+				resolve(r);
 			});
 		});
 	}
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1318,7 +1318,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			cacheRequest.once('error', reject);
 			cacheRequest.once('request', r => {
 				request = r;
-				resolve(r);
+				r.end();
 			});
 		});
 	}
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1318,7 +1318,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			cacheRequest.once('error', reject);
 			cacheRequest.once('request', r => {
 				request = r;
-				r.end();
+				resolve(r);
 			});
 		});
 	}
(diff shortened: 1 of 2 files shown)


test/cache.ts
```diff
@@ -310,3 +310,31 @@ test('does not hang on huge response', withServer, async (t, server, got) => {
...
+		}
+	});
+
+	const cache = new Map();
+
+	const originalResponse = await got({cache});
+
+	t.false(originalResponse.isFromCache);
+	t.is(originalResponse.body, body);
+	t.is(cache.size, 1);
```
szmarczak, line 334: This test fails with a value of `0`, not `1`.

test/cache.ts
```diff
@@ -310,3 +310,31 @@ test('does not hang on huge response', withServer, async (t, server, got) => {
...
 	t.is(body.length, bufferSize * times);
 });
+
+test('cached response ETag', withServer, async (t, server, got) => {
+	const etag = 'foobar';
+	const body = 'responseBody';
+
+	server.get('/', (request, response) => {
+		if (request.headers['if-none-match'] === etag) {
+			response.writeHead(304, {ETag: 'asdf'});
```
szmarczak, line 320: I don't think you should change the etag header if you reply with 304...

test/cache.ts
Giotino: I know, I wrote it in the commit message.
I think it's some kind of bug of got and not of the test.

test/cache.ts
Giotino: Yes, it was a copy and paste from the line of the 200 answer. I am going to remove it

test/cache.ts
Giotino: When the first request `await got({cache})` is resolved, the cache is not immediately populated with the response.

Giotino: From `cacheable-request` documentation:
> .on('request', request)
> request event to get the request object of the request.
>
> This event will only fire if an HTTP request is actually made, not when a response is retrieved from cache. However, you should always handle the request event to end the request and handle any potential request errors.

"you should always handle the request event to end the request"
So i ended the request as suggested and it worked. Also the the promise is now resolved with the actual response and not the request.

source/core/index.ts
```diff
@@ -1318,7 +1318,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			cacheRequest.once('error', reject);
 			cacheRequest.once('request', r => {
 				request = r;
-				resolve(r);
+				r.end();
```
szmarczak, line 1321: Why this change? This seems to be incorrect.

Giotino: I haven't read a lot of got code, I'm still learning how it works internally.
Can you explain me what's the difference between resolving that promise with a request rather than a response?

Giotino: > If yes the correct solution should be putting `resolve(r)` after the `r.end()`.

I read the `_onRequest` code and that solution is wrong, tomorrow I'll look into it.

Giotino: In `cacheable-request` the `request` and `response` events could be emitted before the cache write.

Request event: https://github.com/lukechilds/cacheable-request/blob/master/src/index.js#L156
Response event: https://github.com/lukechilds/cacheable-request/blob/master/src/index.js#L146
Cache write: https://github.com/lukechilds/cacheable-request/blob/master/src/index.js#L131

I can't find any solution to adapt this behavior to got.
I see only 2 options:
The first is to leave it as it is, close (in terms of time) requests might not have the cache.
The second would be working on adding a "cache write" event to `cacheable-request`, to ensure that the got request is resolved only when the cache is successfully written.

I think the best would be the first one, the cache is not a "must", I don't see any problem by having to do a request 2 times (if close in time). I would prefer the speed of the got request, by not waiting a local or __remote__ cache storage.

adityapatadia:
Any thoughts on merging this one? #1287 is pending due to this I think?

szmarczak:
Good job @Giotino! :raised_hands:

adityapatadia:
I think this is still not available in `11.4.0`. Is this planned for future release?

szmarczak:
> Is this planned for future release?

Yes. I'll try to make another one ASAP.

PR #1310, opened 2020-06-09
documentation: update proxy example

Kikobeats:
Feeling `http` is the most common use case there for proxying – normally you don't have access to proxy certificates (also the example is not reflecting that)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1720,8 +1720,9 @@ const tunnel = require('tunnel');
 
 got('https://sindresorhus.com', {
 	agent: {
-		http: tunnel.httpOverHttp({
+		https: tunnel.httpOverHttp({
 			proxy: {
+				rejectUnauthorized: false,
 				host: 'localhost'
 			}
 		})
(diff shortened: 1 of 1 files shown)


sindresorhus:
The URL in the example is using HTTPS though.

Kikobeats:
@sindresorhus true! the thing I was missing is to pass `rejectUnauthorized: false`, so I updated it 🙂

readme.md
```diff
@@ -1722,6 +1722,7 @@ got('https://sindresorhus.com', {
 	agent: {
 		https: tunnel.httpOverHttp({
 			proxy: {
+				rejectUnauthorized: false,
```
sindresorhus, line 1725: Not sure it's a good idea to recommend this. If we are to do it, this needs a well-written comment that explain why it's there, why it's ok to do it, and any potential downsides.

Kikobeats: You're right. Now I understand better the interface; don't feel it's a necessary thing!

PR #1329, opened 2020-06-25
Add tips on stream behavior for new users

awendland:
First off, thank you so much for creating this library! It's been a huge time saver while working with HTTP requests in Node.js.

I was recently working with the stream variant of the API and experienced some confusion around when events were triggered and how to access the response body. I was under the impression that the `response` event returned a response object just like the promise methods and therefore `rawBody` would be populated (which is not the case).

I thought that some slight tweaks to the readme might clarify this for new users. Hopefully these changes will ease onboarding.

#### Checklist

- [x] ~~I have included some tests.~~ N/A 
- [x] ~~If it's a new feature, I have included documentation updates.~~ N/A

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1183,13 +1183,13 @@ The number of times the request was retried.
 
 Sets `options.isStream` to `true`.
 
-Returns a [duplex stream](https://nodejs.org/api/stream.html#stream_class_stream_duplex) with additional events:
+Returns a [duplex stream](https://nodejs.org/api/stream.html#stream_class_stream_duplex) with additional events. This stream can be read from (e.g. listening to the `data` and `end` events) for retrieving the response body, and can be written to (e.g. calling `write(data)` and `end`). If the stream is not read from then the response body will not be downloaded. The additional events expose request and response metadata:
 
 ##### .on('request', request)
 
 `request` event to get the request object of the request.
 
-**Tip:** You can use `request` event to abort request:
+**Tip:** You can use `request` event to abort request (you can also call `destroy` on the stream to abort):
 
 ```js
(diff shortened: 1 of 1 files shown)


szmarczak:
Thanks for the tips, but this is not a Node.js tutorial. If you want to use streams then you have to know how they work. Streams are the same across all Node.js projects. They got docs for it: https://nodejs.org/api/stream.html

awendland:
That's fair. I'm new to Streams and was confoudning that with my usage of this project, and I wouldn't want got's docs to be overrun with what should be common knowledge.

Would you be open to adding types to the stream events in the docs? Such as the `on('response')` event emitting a `PlainResponse`, instead of a `Response<T = unknown>` type which a new user might expect? Event emitters aren't typed in the codebase (though maybe it would be nice to type them such as [1](https://basarat.gitbook.io/typescript/main-1/typed-event) or [2](https://basarat.gitbook.io/typescript/main-1/typed-event)) so having that information in the docs might be helpful for new users of the library since it varies from the Promise API and is outside the scope of the standard Node.js DuplexStream API.

awendland:
My bad, I see that the `Request` object extends `RequestEvents` which does type `on` appropriately with the appropriate callback types.

For some reason my IDE is ignoring this information and only returning the `Duplex` types. Hopefully it's working out better for other people! Sorry for bugging you about this non-issue, I hope your day is otherwise going well!

Screen Shot 2020-06-25 at 11 45 36 AM

szmarczak:
Thanks for your feedback. Don't hesitate to reach out if you have any other questions. Have a good day as well!

PR #1344, opened 2020-07-05
Fix TypeScript types for Promise API

### LINKED ISSUE #1312: NormalizedOptions incorrect types in Hooks
#### Describe the bug

NormalizedOptions passed to the beforeRequest hook seems to be invalid. It *appears* to be the [`NormalizedOptions` defined here](https://github.com/sindresorhus/got/blob/c98f0d7c39ebe71bd0b4875e03c1a7a7b5207347/source/core/index.ts#L186) which is missing some properties.

This bug could be avoided in future by not having different interfaces with the same name.

#### Actual behavior

Typescript error:

```
Type '(options: NormalizedOptions) => Promise<void>' is not assignable to type 'BeforeRequestHook'.
  Types of parameters 'options' and 'options' are incompatible.
    Type 'NormalizedOptions' is missing the following properties from type 'NormalizedOptions': responseType, resolveBodyOnly, retry, isStream
```

#### Expected behavior

According to the docs this should work

#### Code to reproduce

```ts
import { Got, NormalizedOptions } from 'got';

export default function extend(got: Got): Got {
  return got.extend({
    hooks: {
      beforeRequest: [
        (options: NormalizedOptions): Promise<void> => {
          // confirmed that options.responseType is present in `options`, so the type is incorrect
          options.responseType = 'buffer';
        },
      ],
    },
  });
}
```

#### Checklist

szmarczak:
Fixes #1312


sindresorhus:
Thanks for fixing this. Maybe we should add some type tests that ensure we don't regress this? Just some fixtures from https://github.com/sindresorhus/got/issues/1312 and what you used to manually test your changes.

szmarczak:
I just tested it from #1312. Will add it now.

PR #1345, opened 2020-07-05
Upgrade dependencies

sindresorhus:


benchmark/index.ts
```diff
@@ -67,44 +67,44 @@ const suite = new Benchmark.Suite();
 // Benchmarking
 suite.add('got - promise', {
 	defer: true,
-	fn: async (deferred: {resolve(): void}) => {
+	fn: async (deferred: {resolve: () => void}) => {
```
sindresorhus, line 70: TS team recommends this style as it's slightly stricter.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/method-signature-style.md

package.json
```diff
@@ -116,13 +116,11 @@
 		],
 		"rules": {
 			"@typescript-eslint/no-empty-function": "off",
-			"@typescript-eslint/no-base-to-string": "off",
 			"node/prefer-global/url": "off",
 			"node/prefer-global/url-search-params": "off",
-			"@typescript-eslint/prefer-readonly-parameter-types": "off",
+			"import/no-anonymous-default-export": "off",
 			"@typescript-eslint/no-floating-promises": "off",
```
sindresorhus, line 122: @szmarczak We need to enable this rule. It will help us catch unhandled exceptions errors. I think it's better if you do it though, as you can better answer whether something is safe to leave as floating or not. See https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md

source/core/index.ts
```diff
@@ -836,6 +832,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 				options.cookieJar = {
 					setCookie,
+					// TODO: Fix this when upgrading to TypeScript 4.
+					// @ts-expect-error
```
sindresorhus, line 836: This is probably solvable, but I'm not sure and I need sleep.

source/utils/deep-freeze.ts
```diff
@@ -1,6 +1,6 @@
 import is from '@sindresorhus/is';
 
-export default function deepFreeze<T extends object>(object: T): Readonly<T> {
+export default function deepFreeze<T extends Record<string, any>>(object: T): Readonly<T> {
```
sindresorhus, line 3: This is also slightly stricter than `object`, which is why we should prefer it.

source/core/index.ts
```diff
@@ -60,15 +60,13 @@ export interface Agents {
 export const withoutBody: ReadonlySet<string> = new Set(['GET', 'HEAD']);
 
 export interface ToughCookieJar {
-	getCookieString(currentUrl: string, options: {[key: string]: unknown}, cb: (err: Error | null, cookies: string) => void): void;
-	getCookieString(url: string, callback: (error: Error | null, cookieHeader: string) => void): void;
-	setCookie(cookieOrString: unknown, currentUrl: string, options: {[key: string]: unknown}, cb: (err: Error | null, cookie: unknown) => void): void;
-	setCookie(rawCookie: string, url: string, callback: (error: Error | null, result: unknown) => void): void;
+	getCookieString: ((currentUrl: string, options: {[key: string]: unknown}, cb: (err: Error | null, cookies: string) => void) => void) & ((url: string, callback: (error: Error | null, cookieHeader: string) => void) => void);
```
szmarczak, line 63: Why this change?

source/core/index.ts
```diff
@@ -279,10 +277,7 @@ export interface Response<T = unknown> extends PlainResponse {
 }
 
 export interface RequestEvents<T> {
-	on(name: 'request', listener: (request: http.ClientRequest) => void): T;
-	on<R extends Response>(name: 'response', listener: (response: R) => void): T;
-	on<R extends Response, N extends NormalizedOptions>(name: 'redirect', listener: (response: R, nextOptions: N) => void): T;
-	on(name: 'uploadProgress' | 'downloadProgress', listener: (progress: Progress) => void): T;
+	on: ((name: 'request', listener: (request: http.ClientRequest) => void) => T) & (<R extends Response>(name: 'response', listener: (response: R) => void) => T) & (<R extends Response, N extends NormalizedOptions>(name: 'redirect', listener: (response: R, nextOptions: N) => void) => T) & ((name: 'uploadProgress' | 'downloadProgress', listener: (progress: Progress) => void) => T);
```
szmarczak, line 280: This looks ugly.

source/core/index.ts
sindresorhus: It's the linter => https://github.com/sindresorhus/got/pull/1345#discussion_r449921708

source/utils/deep-freeze.ts
szmarczak:
```suggestion
export default function deepFreeze<T extends Record<string, unknown>>(object: T): Readonly<T> {
```

?

source/core/index.ts
sindresorhus: I think we can move the `&` to a separate line.

source/utils/deep-freeze.ts
sindresorhus: It has to be `any` because it's spreading. It's a dumb TS thing. 

https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#higher-order-type-inference-from-generic-constructors See the `{ [k: string]: unknown } is no longer a wildcard assignment target` heading.

source/core/index.ts
sindresorhus: Done: https://github.com/sindresorhus/got/pull/1345/commits/4dfe3e808558251ada2b0be10b2767c2098ad310

source/utils/deep-freeze.ts
szmarczak: I don't know what this exactly means, but this works:

```ts
import is from '@sindresorhus/is';

const isArray = is.array as (arg: unknown) => arg is unknown[];

export default function deepFreeze<T extends Record<string, unknown> | unknown[]>(object: T): Readonly<T> {
	for (const value of Object.values(object)) {
		if (is.plainObject(value) || isArray(value)) {
			deepFreeze(value);
		}
	}

	return Object.freeze(object);
}
```

and this does too:

```ts
import is from '@sindresorhus/is';

export default function deepFreeze<T extends Record<string, unknown>>(object: T): Readonly<T> {
	for (const value of Object.values(object)) {
		if (is.plainObject(value) || is.array(value)) {
			deepFreeze(value as Record<string, unknown>);
		}
	}

	return Object.freeze(object);
}
```

package.json
szmarczak: Well I've fixed it but it says that [this](https://github.com/sindresorhus/got/blob/b51d836f27b3dec16871ec3367d363607af42bb2/test/progress.ts#L15-L42) returns a Promise even though we clearly see that it doesn't.

I think we should revert this PR to 6b7e0ec and go with this option off. Maybe make an issue about this.

szmarczak: It's really a pity you cannot mark a promise as `safe` (non-throwing). If it was, it should skip the error if there's a `return;` after the promise since it's an async task. Like some sort of events.

source/utils/deep-freeze.ts
sindresorhus: 1. Ah right. I missed that it also accepts arrays. Still fails though:

```
source/create.ts:292:59 - error TS2345: Argument of type 'InstanceDefaults' is not assignable to parameter of type 'unknown[] | Record<string, unknown>'.
  Type 'InstanceDefaults' is not assignable to type 'Record<string, unknown>'.
    Index signature is missing in type 'InstanceDefaults'.

292  	value: defaults.mutableDefaults ? defaults : deepFreeze(defaults),
```

2. Obviously that passes as you're using `as` which force-coerces it to whatever you want. That's not a proper solution though.

package.json
sindresorhus: > It's really a pity you cannot mark a promise as safe (non-throwing).

Such an annotation would be useless as anything can throw in JS. Even accessing a defined property can throw (proxies/getter).

The good thing about this `void` annotation is that it forces you to think about the problem and decide whether it's reasonably safe to keep the promise floating or not. It's usually not, per my comment that anything can potentially throw.

sindresorhus: > Well I've fixed it but it says that this returns a Promise even though we clearly see that it doesn't.

Just add an ESLint ignore comment on it. Something there are false-positives if a type is some kind of `any` as it could potentially be a Promise then, TS doesn't know.

sindresorhus: Travis is still failing.

source/core/index.ts
```diff
@@ -862,7 +862,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 									result.catch(handler);
 								} else if (event === 'abort') {
 									// eslint-disable-next-line promise/prefer-await-to-then
-									result.then((request: unknown): void => {
+									void result.then((request: unknown): void => {
```
sindresorhus, line 865: Are you 100% sure `result` will never reject at this point? Maybe do an async-IIFE with a try/catch just to be sure?

source/core/index.ts
```diff
@@ -1302,7 +1305,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		const responseEventName = options.cache ? 'cacheableResponse' : 'response';
 
 		request.once(responseEventName, (response: IncomingMessageWithTimings) => {
-			this._onResponse(response);
+			void this._onResponse(response);
```
sindresorhus, line 1308: Are you sure `_onResponse` could never reject? The code in it is quite complicated so I don't think we should assume so. Better to be safe.

source/utils/deep-freeze.ts
szmarczak: Yeah, you're right. Let's go with any.

source/core/index.ts
szmarczak: The only possibility is when the `response` object is not an `IncomingMessage`. In that case it will be unhandled rejection. The solution would be to

```js
try {
	await this._onResponse(response);
} catch (error) {
	this._beforeError(error);
}
```

or just add the `try/catch` global clause inside `_onResponse`.

So, what should I do?

source/core/index.ts
szmarczak: If `result` throws then the error is caught here:

https://github.com/sindresorhus/got/blob/545e6a239ffcc4f10599565c2bfbd512d3bbcd43/source/core/index.ts#L1516-L1517

Since in that case `fn` would be `this._createCacheableRequest` which returns `result` (the `request` event).

But JS may not see that it's caught **later** and may throw unhandled rejection. The fix is just to add `.catch(() => {})`.

package.json
szmarczak: Actually it's the `checkEvents` function was improperly typed.

source/core/index.ts
sindresorhus: Wrapping the whole function body would make the code hard to read, instead I would make a wrapper function that try/catches `_beforeError` and use that one instead everywhere.

- `_beforeErrorBase` - no try/catch
- `_beforeError` - try catches `_beforeErrorBase` and is the one we actually use.

source/core/index.ts
```diff
@@ -861,10 +861,12 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
 									result.catch(handler);
 								} else if (event === 'abort') {
+									// The empty catch is needed here in case when
+									// it rejects before it's `await`ed in `_makeRequest`.
 									// eslint-disable-next-line promise/prefer-await-to-then
-									void result.then((request: unknown): void => {
+									result.then((request: unknown): void => {
 										(request as ClientRequest).once('abort', handler);
-									});
+									}).catch(() => {});
```
sindresorhus:
```js
(async () => {
	try {
		let request = await result;
		(request as ClientRequest).once('abort', handler);
	} catch {}
})();
```

sindresorhus: I haven't tried, but if this works, use this:

```js
(async () => {
	try {
		let request: ClientRequest = await result;
		request.once('abort', handler);
	} catch {}
})();
```

szmarczak: Should definitely work. Will fix.

source/core/index.ts
szmarczak: `_beforeError` never throws :P

Issue #1358, opened 2020-07-14
Document whether RequestError is the ancestor of all error classes


redneb:
I was looking at `got`'s code and it appears that all errors thrown by it are of a class that is a subclass of `RequestError`. Am I correct about this? If so, I think it would be nice to make that fact part of the module's public api and add it to the documentation. This would allow users to determine if an exception they caught was thrown by `got` by testing whether the exception is an instance of `RequestError`.

szmarczak:
Yes, you are correct. You can easily guess so from reading the first paragraph of https://github.com/sindresorhus/got#errors

A PR is welcome :)

redneb:
Upon more careful inspection, I noticed that one of the possible exception class that can be thrown by `got` is `CancelError` which is a defined in `p-cancelable` and is thus *not* a subclass of `RequestError`.

redneb:
Done #1359.

PR #1359, opened 2020-07-14
Document the hierarchy of error classes

redneb:
This PR addresses #1358. It also adds to the docs information about `UploadError`, which was previously missing. Note that while this PR contains only changes in the documentation, it effectively makes part of the public api of `got` the contract that all errors are instances of `RequestError`, except `CancelError`.

#### Checklist


redneb:
I don't like that `CancelError` is not a subclass of `RequestError` (unlike all other error types), but I guess it is fine as it can be argued that `CancelError`, unlike all other error types, is explicitly caused by the user of the library and not the library itself.

szmarczak:
@redneb Very good point! Can you make an issue about this?

PR #1374, opened 2020-07-23
Fix and add tests for duplicate hook calls on paginated requests

### LINKED ISSUE #1367: BeforeRequest/AfterResponse hooks called multiple times on paginated requests
#### Describe the bug

 - Node.js version: v12.18.1
 - OS & version: Windows 10 1909

When using pagination features, it seems that the `beforeRequest` and `afterResponse` hooks are being called twice for each actual paged request. That is to say, the first request calls each hook twice (before, before, after, after), then the second request calls each hook twice, etc.

I have verified via Fiddler that only one actual request is being made.

#### Actual behavior
Each hook is called twice per physical request

#### Expected behavior
Each hook is only called once per physical request

#### Code to reproduce

```js
const got = require("got");

let first = true;

const client = got.extend({
    prefixUrl: "https://postman-echo.com",
    responseType: "json",
    hooks: {
        beforeRequest: [
            (request) => {
                console.log(`Before request ${request.url}`);
            }
        ],
        afterResponse: [
            (response) => {
                console.log("After response");
                return response;
            }
        ]
    },
    pagination: {
        paginate: () => {
            console.log("paginate", first);
            if(first) {
                first = false;
                return {};
            }
            return false;
        },
        countLimit: 500,
        transform: (response) => Object.keys(response.body),
    }
});

client.paginate.all("get").then((data) => {
    console.log(data);
}).catch((err) => {
    throw err;
});
```

This produces the console ouptut:
```
Before request https://postman-echo.com/get
Before request https://postman-echo.com/get
After response
After response
paginate true
Before request https://postman-echo.com/get
Before request https://postman-echo.com/get
After response
After response
paginate false
[ 'args', 'headers', 'url', 'args', 'headers', 'url' ]
```

#### Checklist

hellodavie:
Resolves #1367 ✨

Modified the options normaliser to check for duplicate hooks before concatenating arrays.
Following up from my comment on the issue, I decided to just filter the array instead of setting some sort of flag to indicate duplicated hooks.

I also added two tests for this specific scenario 😇 Wasn't really sure where to put it (hooks or pagination?) so I made a new file.

#### Checklist
- [x] If it's a new feature (not applicable), I have included documentation updates.


test/hooks-pagination.ts
```diff
@@ -0,0 +1,106 @@
+import test from 'ava';
```
szmarczak, line 1: Move the content of this file either to `test/hooks.ts` or `test/pagination.ts`

source/core/index.ts
```diff
@@ -930,12 +930,16 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		if (defaults && !areHooksDefault) {
 			for (const event of knownHookEvents) {
 				const defaultHooks = defaults.hooks[event];
-
 				if (defaultHooks.length !== 0) {
+					let optionsHooks = options.hooks[event] as any[];
+					if (optionsHooks && optionsHooks.length !== 0) {
+						optionsHooks = optionsHooks.filter(hook => !(defaultHooks as any[]).includes(hook));
```
szmarczak, line 936: This prevents intentional duplicates.

hellodavie: Oh oopsies! I'll add some sort of flag instead of filtering through the hooks, unsure whether I should use `kIsNormalizedAlready` since it might have unintended side effects, we'll see after the tests run.

test/hooks-pagination.ts
hellodavie: Done, moved it into `test/hooks.ts` 🔥
Everything looks good now, just adding one more test to make sure intentional duplicate hooks still work.

hellodavie:
All fixed now. Your thoughts?

szmarczak:
I'll review this on Sunday. I'm away from home now.

hellodavie:
All good! Enjoy your time out 😊

source/core/index.ts
```diff
@@ -927,7 +929,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			}
 		}
 
-		if (defaults && !areHooksDefault) {
+		const hooksNormalizedAlready = options && (options as NormalizedOptions)[kHooksNormalizedAlready];
```
sindresorhus, line 932: Can you use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining here?

hellodavie: Ooh neat, I didn't know about that. Done! Also slightly tweaked the tests that I added.

szmarczak:
I figured out why this happens. It's because you call `.paginate()` on the instance which has already defined hooks. The function normalizes the options here

https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/source/create.ts#L206

And calls `instance(normalizedOptions)`, which results in duplicate properties because `normalizedOptions` can be the same as defaults.

https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/source/create.ts#L227

Currently I'm on my way back, I'll be home in 7h. This fix is not correct, it should be in `create.ts`. I'll try to fix this by tomorrow evening, but feel free to propose a solution :)

hellodavie:
Thanks for your help. From what I see, the main purpose of this line is to make sure the pagination options are normalised for use in `paginateEach`, and not to merge hooks or any other properties (even though it does).
https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/source/create.ts#L206

So we can just use `instance(normalizedOptions)` to handle all the merging of options, by passing in `options` instead of `normalizedOptions`. However, for all the pages afterwards, we'll need to access the normalized options, merge them, and then pass them back in. To prevent duplicates, we need to reset the hooks because the options have been normalized when we access them here: https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/source/create.ts#L259-L263

Thats my proposed solution. What do you think?

Also, I modified the tests here to match the function signature `paginate.all(url, options)`. Some of the tests look like this https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/test/pagination.ts#L553 But some look like this https://github.com/sindresorhus/got/blob/a3e171c4d43c8bbb236ef3a04feb25ba6fdf6c49/test/pagination.ts#L358 

I don't know why the ci is failing, cause it passed on my own repo.

szmarczak:
Thanks for the tests :heart: Fixed in e02845f1aa737d55dce23381d0f4f2a61b1eb5e1

PR #1380, opened 2020-07-26
Add got4aws to AWS section

SamVerschueren:
As discussed, I replaced the code in the AWS section with [`got4aws`](https://github.com/SamVerschueren/got4aws) because the code wasn't working either.

Feel free to let me know if you want to rephrase some other things as well.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -699,8 +699,6 @@ Default: `[]`
 
 Called with [normalized](source/core/index.ts) [request options](#options). Got will make no further changes to the request before it is sent. This is especially useful in conjunction with [`got.extend()`](#instances) when you want to create an API client that, for example, uses HMAC-signing.
 
-See the [AWS section](#aws) for an example.
-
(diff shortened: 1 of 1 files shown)


szmarczak:
An example on how to make a request w/ credentials would be great :)

SamVerschueren:
You mean something like I do in the `got4aws` example itself?

```ts
import got4aws from 'got4aws';

// Load credentials from `~/.aws/credentials`
const got = got4aws({
	providers: new AWS.SharedIniFileCredentials({profile: 'myProfile'})
});

(async () => {
	const {body} = await got.get('https://12abc34.execute-api.us-east-1.amazonaws.com/v0');

	console.log(body);
	//=> {status: 'ok'}
})();
```

SamVerschueren:
But then without TS and the IIFE.

readme.md
```diff
@@ -1860,29 +1858,14 @@ got('unix:/var/run/docker.sock:/containers/json');
...
-				const credentials = await chain.resolvePromise();
-				aws4.sign(options, credentials);
-			}
-		]
-	}
-});
+const awsClient = got4aws();
 
-const response = await awsClient('endpoint/path', {
+const response = await awsClient('https://<api-id>.execute-api.<api-region>.amazonaws.com/<stage>/endpoint/path', {
```
sindresorhus, line 1868: Is it clear enough to the user that they need to fill in `<api-id>` with their own value?

szmarczak: I guess so. If anyone finds this hard then he may create an issue and we'll fix that :)

PR #1384, opened 2020-08-01
Add `retry` stream event

### LINKED ISSUE #1381: Add retrying for the stream API
Is this just a technical limitation of streams itself, or is it just that it hasn't been implemented?

szmarczak:
Fixes #1381

```js
const got = require('got');
const fs = require('fs');

let writeStream;

const fn = (retryCount = 0) => {
	const stream = got.stream('https://example.com');
	stream.retryCount = retryCount;

	if (writeStream) {
		writeStream.destroy();
	}

	writeStream = fs.createWriteStream('example.com');

	stream.pipe(writeStream);

	// If you don't attach the listener, it will NOT make a retry.
	// It automatically checks the listener count so it knows whether to retry or not :)
	stream.once('retry', fn);
};

fn();
```

#### Checklist


szmarczak:
@epicfaace Will you be happy with the API above?

epicfaace:
Yes, that looks good! Perhaps retryCount could be specified in the `got.stream` method options, but otherwise it's great.

szmarczak:
> Perhaps retryCount could be specified in the got.stream method options

No, because when you have defined the options as a constant, then you would need to do `{...options, retryCount}` which would lead to a significant performance drop.

PR #1403, opened 2020-08-11
Don't allocate buffer twice

xamgore:
* `getBuffer` returns a buffer of concatenated response chunks.
* Its result is stored in `rawBody` and passed to `parseBody`.
* Then it meets `Buffer.from(rawBody)` and triggers a new buffer allocation:
   https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_buffer

Now a new Buffer created without copying the underlying memory.
https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -319,6 +319,8 @@ const body = await got(url).json();
 const body = await got(url, {responseType: 'json', resolveBodyOnly: true});
 ```
 
+**Note:** `buffer` will return the raw body buffer. Modifying it will also alter the result of `promise.text()` and `promise.json()`. Before overwritting the buffer, please copy it first via `Buffer.from(buffer)`. See https://github.com/nodejs/node/issues/27080
+
 ###### parseJson
 
 Type: `(text: string) => unknown`\
--- a/source/as-promise/parse-body.ts
+++ b/source/as-promise/parse-body.ts
@@ -18,7 +18,7 @@ const parseBody = (response: Response, responseType: ResponseType, parseJson: Pa
 		}
 
 		if (responseType === 'buffer') {
-			return Buffer.from(rawBody.buffer);
+			return rawBody;
 		}
 
 		throw new ParseError({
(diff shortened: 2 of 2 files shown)


szmarczak:
`parseBody` copies the body so even if you modify the buffer then `promise.text()` will give the original body.

szmarczak:
@sindresorhus Although we could add an environment variable to disable this.

```ts
		if (responseType === 'buffer') {
			return process.env.GOT_MUTABLE_BODY ? rawBody : Buffer.from(rawBody);
		}
```

or should we just return the original buffer? If so, then this could happen:


```ts
const promise = got('https://somesitehere.com'); //=> 123
const buffer = await promise.buffer();
buffer.write(49, 0);

const text = await promise.text(); //=> 223, not 123!
```

or we could warn users in the docs... If we decide to return the original buffer then it will save 2x RAM for people who do `got().buffer()`.

sindresorhus:
> or should we just return the original buffer? If so, then this could happen:

That seems extremely unlikely though. Why would they write to a buffer they *got* from a method call. But maybe we could just override the `.write()` method and throw an error if used?

---

Also, upvote and comment on this issue: https://github.com/nodejs/node/issues/27080

sindresorhus:
> But maybe we could just override the .write() method and throw an error if used?

@szmarczak Thoughts on this?

szmarczak:
> Thoughts on this?

That wouldn't work as you'd still be able to do `buffer[0] = 123;` since it's also `Uint8Array`...

sindresorhus:
We could `Object.freeze` it?

szmarczak:
```
TypeError: Cannot freeze array buffer views with elements
```

sindresorhus:
@szmarczak Maybe we could at least wrap the TS type in `Readonly` so at least TS users are protected?

PR #1411, opened 2020-08-19
Improve type inference

xamgore:
`PCancelable<T>` extends `Promise<T>`, but due to the default import `p-cancelable` this fact is erased.
As a result `await got.get(...).json<E>()` returns `any`. After the fix Webstorm IDE returns `E`.

#### Checklist

### DIFF
--- a/source/as-promise/types.ts
+++ b/source/as-promise/types.ts
@@ -286,7 +286,7 @@ export class CancelError extends RequestError {
 	}
 }
 
-export interface CancelableRequest<T extends Response | Response['body'] = Response['body']> extends PCancelable<T>, RequestEvents<CancelableRequest<T>> {
+export interface CancelableRequest<T extends Response | Response['body'] = Response['body']> extends PCancelable<T>, Promise<T>, RequestEvents<CancelableRequest<T>> {
 	json: <ReturnType>() => CancelableRequest<ReturnType>;
 	buffer: () => CancelableRequest<Buffer>;
 	text: () => CancelableRequest<string>;


szmarczak:
This needs a test.

xamgore:
What kind of tests? [dtslint](https://github.com/microsoft/dtslint)? Can't think of one.

szmarczak:
> What kind of tests?

linting =/= testing

> What should be tested?

The change you introduce.

> Can't think of one.

There are already 500+ unit tests. Feel free to look at them.

sindresorhus:
> PCancelable<T> extends Promise<T>, but due to the default import p-cancelable this fact is erased.

I don't see why that would cause the type to be erased.

> After the fix Webstorm IDE returns E.

Can you check with VS Code to ensure it's not a Webstorm bug?

source/as-promise/types.ts
```diff
@@ -286,7 +286,7 @@ export class CancelError extends RequestError {
 	}
 }
 
-export interface CancelableRequest<T extends Response | Response['body'] = Response['body']> extends PCancelable<T>, RequestEvents<CancelableRequest<T>> {
+export interface CancelableRequest<T extends Response | Response['body'] = Response['body']> extends PCancelable<T>, Promise<T>, RequestEvents<CancelableRequest<T>> {
```
szmarczak, line 289: This is actually incorrect. `PCancelable` already extends the `Promise`.

https://github.com/sindresorhus/p-cancelable/blob/4b7d014c5be44889e40014dea098ae44819b95d6/index.d.ts#L22

xamgore: Due to the interface [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) there is no new information for Typescript. But Webstorm's type checker can't extract information from `p-cancelable` and states `type PCancelable = any`, thus `CancelableRequest.then` returns `any`.

`CancelableRequest extends PCancelable<T>, Promise<T>` forces Webstorm to merge `type PCancelable = any` with `Promise<T>`, so `CancelableRequest.then` returns `T` now.

The reason Webstorm can't extract information from `p-cancelable` is awkward: it's not listed in the `package.json`. Compare, not listed:



listed:

szmarczak: > Due to the interface declaration merging there is no new information for Typescript

How does this relate to the actual issue? We don't use this.

> But Webstorm's type checker

See @sindresorhus's response above.

> The reason Webstorm can't extract information from p-cancelable is awkward: it's not listed in the package.json.

You're wrong:

https://github.com/sindresorhus/got/blob/eaf1e02b21e4b8c08197c6400844d4098828ecf4/package.json#L55

xamgore:
> @szmarczak:
> linting =/= testing
> There are already 500+ unit tests. Feel free to look at them.

Thank you for mentioning, though **dtslint** tests a TypeScript declaration file both for style and [**correctness**](https://github.com/microsoft/dtslint#write-tests). My suggestion was to write a test that asserts that an expression `got.get(...).json<E>()` is an instance of `Promise<E>` type with `then`, `catch`, `finally` methods.

Such a test should detect two possible future changes: either `PCancelable` or `CancelableRequest` is not an inheritor of `Promise`, in other words `then`, `catch`, and `finally` methods are not defined, what would mean already existing 500 unit tests should fail. Seems no additional tests should be written.

> @sindresorhus: I don't see why that would cause the type to be erased. Can you check with VS Code to ensure it's not a Webstorm bug?

They shouldn't, though Webstorm implemented their own type checker (I can find a thread in twitter from one of their developers stating that, if you're interested) with bugs included. The [description](https://github.com/sindresorhus/got/pull/1411#discussion_r473317936) is in the comment.

Checked with the same `tsconfig` in vscode, couldn't reproduce.

szmarczak:
> My suggestion was to write a test that asserts that an expression got.get(...).json<E>() is an instance of Promise<E> type with then, catch, finally methods.

A unit test is fine. No need to introduce `dtslint` or `tsd` at the moment, since the unit tests are also written in TypeScript.

> Seems no additional tests should be written.

Then your change is useless if it requires no tests.

> Checked with the same tsconfig in vscode, couldn't reproduce.

Closing since it is not a Got issue.

sindresorhus:
> Checked with the same tsconfig in vscode, couldn't reproduce.

Then you should open an issue on Webstorm.

xamgore:
That issue would be fixed in 1-3 years if ever :smirk:

source/as-promise/types.ts
xamgore: * `PCancelable` extends `Promise`, so can do `CancelableRequest`
  * As typescript provides [interface declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces), there wouldn't be any new information for the compiler
  * As you write `await got(...)` in examples, you expect `CancelableRequest` to extend `Promise`
  * If `PCancelable` stops inheriting `Promise` someday, there will be a type error `CancelableRequest should implement Promise interface methods: then, catch`
  * A clear parents list improves Webstorm type inheritance

With all the above I don't really understand what may be incorrect here. A small change of code makes everyone happy and doesn't make the code worse.

> @xamgore: The reason Webstorm can't extract information from p-cancelable is awkward: it's not listed in the package.json.
> @szmarczak: You're wrong: "link to got/package.json"

If `p-cancelable` wasn't listed in `got/package.json` the project wouldn't be compilable. As you could guess, I meant `%username%/project/package.json`.

Issue #1421, opened 2020-08-26
Can not remove accept-encoding from headers


EsterDanielYtterbrink:
#### Describe the bug

 - Node.js version: v14.8.0
 - OS & version: macOS 10.15.6 (19G2021)

I am migrating some code to got and it has existing tests that, among others, checks the sent headers. I want to be able to use those tests to migrate the code safely. But when setting the header 'accept-encoding' to undefined to match the tests it is set to 'gzip, deflate, br'

#### Actual behavior
Setting `options['accept-encoding'] = undefined` results in `options['accept-encoding'] =  'gzip, deflate, br'`
...

#### Expected behavior
Setting `options['accept-encoding'] = undefined` results in options not to have the property accept-encoding.

...

#### Code to reproduce

I am quite sure the problem is in index.ts on line 2269 and would love to write a test for it, but am unsure where to place the test. 

```js
for (const key in headers) {
			if (is.undefined(headers[key])) {
				// eslint-disable-next-line @typescript-eslint/no-dynamic-delete
				delete headers[key];
			} else if (is.null_(headers[key])) {
				throw new TypeError(`Use \`undefined\` instead of \`null\` to delete the \`${key}\` header`);
			}
		}

		if (options.decompress && is.undefined(headers['accept-encoding'])) {
			headers['accept-encoding'] = supportsBrotli ? 'gzip, deflate, br' : 'gzip, deflate';
		}
...
```

#### Checklist

- [x ] I have read the documentation.
- [ x] I have tried my code with the latest version of Node.js and Got.

szmarczak:
Read the docs. Also the code you've posted exactly describes the scenario in which the header is set.

EsterDanielYtterbrink:
So why should it not be possible to remove accept-encoding from your header?

szmarczak:
It is already possible. Read the docs.

EsterDanielYtterbrink:
I have read the docs, and tried to set it to undefined, but then it is removed just to be set again by 		
```
if (options.decompress && is.undefined(headers['accept-encoding'])) {
			headers['accept-encoding'] = supportsBrotli ? 'gzip, deflate, br' : 'gzip, deflate';
		}
}
```
Or is it something else I have missed? 
- I have done a project wide search on accept-encoding, and find nothing
- I have a test in my codebase that confirms the bug and that using my branch solves it

Though if I allow retries the accept-encoding will show up again for the retries. So it is not a universal fix.

[...]

PR #1422, opened 2020-08-26
Makes it possible to remove accept-endcoding 

EsterDanielYtterbrink:
Changes the order when options are set and checks if the property exists rather than if it is undefined.
 Fixes issue #1421
#### Checklist


szmarczak:
https://github.com/sindresorhus/got/issues/1421#issuecomment-680837314

PR #1364, opened 2020-07-17
Add `pfx` HTTPS option

markdboyd:
Fixes #1306

Adds support for the `pfx` option from TLS `SecureContextOptions` to the `HTTPSOptions` accepted by got.  

I tried to follow the existing convention of `passphrase` where `options.pfx` and `options.https.pfx` are both allowed, but using `options.pfx` prints a deprecation warning. 

~I also added a basic test of the deprecation warning~. I removed this test because it was failing since I wasn't providing a real PFX fixture. It also wasn't a very meaningful test.

I wasn't sure what other tests to add but I'm happy to add any that are requested.

#### Checklist

- [x] I have read the documentation.
- [x] I have included a pull request description of my changes.
- [x] I have included some tests.
- [x] If it's a new feature, I have included documentation updates.

PR #1430, opened 2020-08-31
Add more HTTPS tests

Giotino:
Some tests for the HTTPS options.

These tests simulate a real PKI application.
E.g. https://github.com/Giotino/got/blob/ca8d81314ac93bacd5e0c1da207210adf4e85a4a/test/https.ts#L232-L248

Types are messed up, I'm going to fix them soon, I'm also going to add support for `pfx`, needed for #1364.

This is a draft, I'm going to add more tests covering more functionalities (e.g. key passphrase or invalid certificates).

`create-cert-test-server.ts` is like https://github.com/lukechilds/create-test-server, but more specific for our needs.

#### TODO List
- [x] Add PFX test environment
- [x] Fix types

#### Checklist


Giotino:
In order to do more deeper HTTPS tests it has been needed to replace https://github.com/lukechilds/create-test-server for those tests.

The new `create-https-test-server` is an helper in the got test suite. It's extensible and can easily be adapted to create an HTTPS server with the need of every single tests, this is going to be useful when new HTTPS options are going to be added (e.g. the `pfx` PR).

This has also enabled me to remove external dependencies for testing the expired certificate and the wrong host (https://github.com/sindresorhus/got/pull/1430/files#diff-ef513147d9bbe98f3dc4e9dc04783109L123-R132).

All that said I think this PR is ready for some review.
@szmarczak @sindresorhus

szmarczak:
Does the tests pass on Windows?

test/helpers/create-https-test-server.ts
```diff
@@ -0,0 +1,72 @@
...
+	https: https.Server;
+	caKey: Buffer;
+	caCert: Buffer;
+	url: string;
+	port: number;
+	close: () => Promise<any>;
+}
+
+const createHttpsTestServer = async (options: HttpsServerOptions = {}): Promise<ExtendedHttpsTestServer> => {
+	const createCSR = pify(pem.createCSR);
```
szmarczak, line 22:
```suggestion
	const createCertificateSingingRequest = pify(pem.createCSR);
```

test/helpers/create-https-test-server.ts
```diff
@@ -0,0 +1,72 @@
...
+		clientKey: serverCSRResult.clientKey,
+		serviceKey: caKey,
+		serviceCertificate: caCert,
+		days: options.days ?? 365
+	});
+	const serverKey = serverResult.clientKey;
+	const serverCert = serverResult.certificate;
+
+	const server = express() as ExtendedHttpsTestServer;
+	server.https = https.createServer(
```
szmarczak, line 46: Do we need to set the `https` property?

szmarczak: Ah, yes, you're right.

szmarczak: Maybe let's rename `server` to `app` and `server.https` to `app.server`? Does Express allow us to access the underlying server some other way?

Giotino:
> Does the tests pass on Windows?

Yes, I'm running it on Windows

test/https.ts
```diff
@@ -1,37 +1,40 @@
...
+	t.truthy((await got({
 		https: {
+			certificateAuthority: [],
 			rejectUnauthorized: false
 		}
 	})).body);
 });
 
-test('https request with ca', withServer, async (t, server, got) => {
+test('https request with ca', withHttpsServer(), async (t, server, got) => {
```
szmarczak, line 27: Can you define `withHttpsServer` so `()` is not necessary to call every time?

Giotino: I could, but there are some tests that give parameters to it, so I thought that leaving it always as a function call was going to be cleaner.

szmarczak: Cool. I'm fine with that :)

test/helpers/create-https-test-server.ts
Giotino: HTTPS is not 100% needed, but it could be useful to have it there.
I don't think there's a way to access the server from Express because the HTTPS Server calls Express as a function, so the connection is like `Server => Express` and not `Express => Server`.

The rename is ok for me, I had just copied the lukechilds notation.

sindresorhus (approved):

test/helpers/create-https-test-server.ts
Giotino: I prefer to leave it as it is because is a mirror of the name of the original function.

BTW: I see that this conversation is marked as resolved, so is it fine?

test/helpers/create-https-test-server.ts
Giotino: Since this conversation is marked as resolved I assume the rename is not needed, am I correct?

Issue #1441, opened 2020-09-09
11.6.1 hangs when running on node v14.10.0


benjamincburns:
#### Describe the bug

 - Node.js version: v14.10.0
 - OS & version: all

I use `got` in a node CLI tool. I accidentally published it with a `package-lock.json` instead of `npm-shrinkwrap.json`. Unfortunately this meant that my tool was tracking the latest version of `got`. When `got@11.6.1` was released, my tool started hanging when making HTTP requests when running under node v14.10.0.

I also see that [the CI build for `got`](https://github.com/sindresorhus/got/runs/1088472472) is broken for the v11.6.1 release commit. The failure occurred on the node v14 build, and apparently [the build failed due to tests timing out](https://travis-ci.com/github/sindresorhus/got/jobs/382443656#L869), which suggests that this issue is already reproduced extensively by the existing test suite.

#### Actual behavior

`got` hangs indefinitely when executing [this statement](https://github.com/PegaSysEng/quorum-dev-quickstart/blob/2e443b18c4a90c6f617712c1b6cce5c1ffdf62b5/src/service/orchestrate/index.ts#L91-L93)

#### Expected behavior

The line lined above would execute as normal.

#### Code to reproduce

See link above, or, per your CI (also linked above), just run your test suite under node v14.10.0

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

benjamincburns:
I did a bit of testing, and it appears that this is only an issue on node v14.10.0, which was also released earlier today. I've done no digging to identify the actual bug, but given the version-specific nature of the beast, I would imagine that this is a bug in NodeJS, as a minor release has changed behaviour in existing code.

targos:
Reproduction:

```js
require('got')('https://www.google.com').then(console.log, console.log)
```

This logs the response in Node.js 14.9.0 and exits without any logs in Node.js 14.10.0

Giotino:
Following the discussion on https://github.com/nodejs/node/issues/35116

`this[kStopReading]` should be an indicator of "stop reading the stream" (as pointed out by @targos Node.JS v14.10.0 destroys the stream even on success, which is correct), but in the `_isAboutToError` getter is used as an indicator of an error state.

ronag:
Could somebody apply the following [PR](https://github.com/nodejs/node/pull/35120) on v14.x-staging, build it and try to reproduce this issue? To see if it resolves it.

[...]

PR #1446, opened 2020-09-09
Node.JS v14.10.0

Giotino:
Current fixes:

- Possible fix for the `kStopReading` wrong usage:
> `this[kStopReading]` should be an indicator of "stop reading the stream" and it's set to true in `_destroy`.
But in the `_isAboutToError` getter `this[kStopReading]` is used as an indicator of an error state.
As pointed out by @targos: Node.JS v14.10.0 destroys the stream even on success, which is correct.

#### Checklist

Fixes: #1441


Giotino:
Tests passed without "Possible fix for the `kStopReading` wrong usage": 166
Tests passed with "Possible fix for the `kStopReading` wrong usage": 444

There are still a lot of tests in timeout.

Issue #1444, opened 2020-09-09
some defaults are not being respected in normalizeArguments


ghermeto:
#### Describe the bug

 - Node.js version: 10, 12, 14
 - OS & version: MacOS

There is a problem with the function `normalizeArguments` related to `prefixUrl` where defaults are not being applied. 

#### Actual behavior

The following test fails with `TypeError: No URL protocol specified`:

```ts
test('extending prefixUrl', withServer, async (t, server) => {
	server.get('/', (req, res) => {
		res.json(req.headers);
	});

	const instance1 = got.extend({
		prefixUrl: server.url
	});

	const instance2 = got.extend({
		headers: { 'x-test': 'test' },
		responseType: 'json'
	});

	const merged = instance1.extend(instance2);

	const {body} = await merged.get('');
	t.is((body as unknown as Record<string, unknown>)['x-test'], 'test');
});
```
However, if `merged` is defined as:

```ts
const merged = instance2.extend(instance1);
```
the test will pass. It seems like some defaults are not being properly applied because when some properties are not defined, it is set them to an empty string. So, when the next instance performs a shallow merge with the object spread, it will prefer the empty string.  

#### Expected behavior

I should be able to define merge both as `instance1.extend(instance2)` and `instance2.extend(instance1)`

#### Quick Patch
Adding the following code to [L1610 in /source/core/index.ts](https://github.com/sindresorhus/got/blob/master/source/core/index.ts#L1610) will solve the issue, but other properties might also be affected by the same issue.
 
```ts
// !options.prefixUrl checks if value is falsy (includes empty string and 0)
if (defaults?.prefixUrl && !options.prefixUrl) {
    options.prefixUrl = defaults.prefixUrl;
}
```
I will check the other properties and I can submit a PR later this week.

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

Giotino:
> I will check the other properties and I can submit a PR later this week.

`prefixUrl` is a special case because its default is an empty string and this is enforced during normalization.

In your example `instance1.prefixUrl` is `server.url` and `instance2.prefixUrl` is `''`, but `''` is not treated as "to overwrite" from the `normalizeArguments` function.
This PR will fix it https://github.com/sindresorhus/got/pull/1448

szmarczak:
@ghermeto If you expect to ignore Got defaults when merging, let's continue the discussion at #1450

ghermeto:
[...]

PR #1448, opened 2020-09-09
Inherit the `prefixUrl` option from parent if it's `undefined`

Giotino:
#### Checklist

Fixes: #1444

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1608,14 +1608,14 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		options.password = options.password ?? '';
 
 		// `options.prefixUrl` & `options.url`
-		if (options.prefixUrl) {
+		if (options.prefixUrl && options.prefixUrl !== '') {
 			options.prefixUrl = options.prefixUrl.toString();
 
 			if (options.prefixUrl !== '' && !options.prefixUrl.endsWith('/')) {
 				options.prefixUrl += '/';
 			}
 		} else {
-			options.prefixUrl = '';
+			options.prefixUrl = defaults?.prefixUrl ?? '';
 		}
 
 		if (is.string(options.url)) {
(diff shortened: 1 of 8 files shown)


Giotino:
`prefixUrl: ''` is the default, but it was overwriting an explicit specified `prefixUrl` on `extend`.

Now (in this PR), `prefixUrl: ''` gets overwritten, but it's impossible to remove it from the options (I think this shouldn't be considered a problem).

The failing `cache` and `stream` tests are test that use `withServer` (which use `prefixUrl`) but not really need it, I'm going to remove it.
The last failing test is a `pagination` test, I'm working on it.

Giotino:
The only breaking change (might not be so) that I could find is that with the pagination (if you're using `prefixUrl`) you can't change the URL returning something like this `{ url: 'http://something/?page=1234&token=abcd' }` in `pagination.paginate` unless you split the `url` into `prefixUrl` and `url` (in order to overwrite the `prefixUrl`):
```js
{ 
  prefixUrl: 'http://something/',
  url: '?page=1234&token=abcd'
}
```

source/core/index.ts
```diff
@@ -1608,14 +1608,14 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		options.password = options.password ?? '';
 
 		// `options.prefixUrl` & `options.url`
-		if (options.prefixUrl) {
+		if (options.prefixUrl && options.prefixUrl !== '') {
```
szmarczak, line 1611: This is unnecessary as an empty string is `falsy` anyway.

szmarczak:
> Now (in this PR), prefixUrl: '' gets overwritten, but it's impossible to remove it from the options (I think this shouldn't be considered a problem).

Actually, if we explicitly pass `prefixUrl: ''` then it should be empty. If it is `undefined` then it should take the previous value.

PR #1454, opened 2020-09-10
Downgrade Travis CI Node.js version to 14.9.0

Giotino:
In order to keep using Travis CI correctly v14.9.0 has been specified instead of v14 (latest available).

This is a temporary measure, a rollback of the commit "stream: simpler and faster Readable async iterator" which broke nearly everything in Got has been planned for the next Node.JS release (v14.10.1) https://github.com/nodejs/node/pull/35137

I don't know if downgrading Node.JS for CI is the right thing to do, but I think it won't do any harm while waiting for some fix for v14.10.0.

@sindresorhus @szmarczak


Giotino:
Now that the commit "stream: simpler and faster Readable async iterator" has been reverted this PR should also be reverted.

@sindresorhus

sindresorhus:
@Giotino We have to wait until it's released and available on Travis.

Giotino:
> @Giotino We have to wait until it's released and available on Travis.

As you can see here https://travis-ci.com/github/sindresorhus/got/jobs/383613035
NVM on Travis CI is installing `v14.10.1` for the `14` target.

sindresorhus:
https://github.com/sindresorhus/got/commit/0e0eefdd8538e910cba6d89df5a3268743aa7ec8

PR #1457, opened 2020-09-10
Disable a test as the built-in `openssl` on macOS does not support negative days

Giotino:
Title self-explanatory.

Reference: https://ci.nodejs.org/job/citgm-smoker-nobuild/nodes=osx1015/935/testReport/junit/(root)/citgm/got_v11_6_1/

The test has been disabled, I'm looking for some kind of replacement. We cannot play with timers because the TLS logic related to the certificate expiration is done by openssl.

### DIFF
--- a/test/https.ts
+++ b/test/https.ts
@@ -115,14 +115,16 @@ test('https request with `checkServerIdentity` NOT OK', withHttpsServer(), async
 	);
 });
 
-test('https request with expired certificate', withHttpsServer({days: -1}), async (t, _server, got) => {
-	await t.throwsAsync(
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):
I think it's fine to just disable this test on macOS. Not worth wasting time on finding a workaround.

Giotino:
I think we can leave this test disabled for MacOS. The CI infrastructure is based on Linux, we can catch this problem. If it doesn't work on MacOS (but works on Linux) the problem is probably going to be the openssl on MacOS.

PR #1461, opened 2020-09-11
General improvements to the tests and CI Environment

Giotino:
As can be seen here: https://ci.nodejs.org/job/citgm-smoker-nobuild/935/ the fact that the Got tests pass on "an ubuntu release" doesn't mean that it works everywhere.

I added some OS: Ubuntu Trusty (14.04), Ubuntu Xenial (16.04), Ubuntu Bionic (18.04), Ubuntu Focal (20.04), Windows 1809 (newer versions are not supported), macOS 10.14, macOS 10.15.

Whenever we decide to switch to Github Actions these are the available OS.

Virtual environment | YAML workflow label
-- | --
Windows Server 2019 | windows-latest or windows-2019
Ubuntu 20.04 | ubuntu-20.04
Ubuntu 18.04 | ubuntu-latest or ubuntu-18.04
Ubuntu 16.04 | ubuntu-16.04
macOS Catalina 10.15 | macos-latest or macos-10.15


Giotino:
With the latest commit (https://github.com/sindresorhus/got/pull/1461/commits/9fa1d1cbc70666f67fc2ff9220885f9b60124731) I copy-and-pasted (and ported to Typescript) https://github.com/lukechilds/create-test-server .
With my modifications it doesn't create a new HTTPS certificate each time (which takes time), in fact it doesn't even support HTTPS.

For tests regarding HTTPS `createHttpsTestServer` (already present in the repo) should be used, and it's already used by all the `https.ts` tests.
I had to disable 5 tests (4 marked as failing, 1 commented) because they need to be ported to the new HTTPS Server, I'm working on it.

Tests are now much faster and shouldn't timeout on slower system.

#### How is this related to this PR?

Looking at https://ci.nodejs.org/job/citgm-smoker-nobuild/nodes=osx1014/935/testReport/junit/(root)/citgm/got_v11_6_1/ (macOS 10.14)
the tests timed out, as you can see there are some tests which took up to a minute and a half to run.

It could be due to the "different" openssl present on macOS (might be a slower release), or to the fact that the hardware beneath the OS is a Mac mini (a common setup for a macOS server).

Giotino:
It seems that Tavis CI Ubuntu Focal VM does not resolve IPv6 hostnames. I disabled that test for Focal on Travis CI.
```js
process.env.TRAVIS_DIST !== 'focal'
```

test/agent.ts
```diff
@@ -80,53 +54,79 @@ test('non-object agent option works with https', withServer, async (t, server, g
 	agent.destroy();
 });
 
-test('redirects from http to https work with an agent object', withServer, async (t, server, got) => {
-	prepareServer(server);
+test('redirects from http to https work with an agent object', withServer, async (t, serverHttp) => {
+	await withHttpsServer()(t, async (t, serverHttps, got) => {
```
szmarczak, line 58: Maybe let's do `withHttpAndHttpsServer`?

test/error.ts
```diff
@@ -36,7 +36,7 @@ test('properties', withServer, async (t, server, got) => {
 test('catches dns errors', async t => {
 	const error = await t.throwsAsync<RequestError>(got('http://doesntexist', {retry: 0}));
 	t.truthy(error);
-	t.regex(error.message, /ENOTFOUND/);
+	t.regex(error.message, /(ENOTFOUND|EAI_AGAIN)/);
```
szmarczak, line 39: Why this change?

test/http.ts
```diff
@@ -283,9 +283,14 @@ if (IPv6supported) {
...
 
-		const response = await got(`http://[::1]:${server.port}/ok`);
-
-		t.is(response.body, '::1');
+		try {
+			const response = await got(`http://[::1]:${server.port}/ok`);
+
+			t.is(response.body, '::1');
+		} catch (error) {
+			// If network unreachable (happens on TravisCI with Ubuntu Bionic and Focal)
```
szmarczak, line 291: Do we know why it does not work?

szmarczak: Ah. Just saw the comment below. Then maybe let's do conditional `test.skip`?

szmarczak:
```ts
{
	const testFn = process.env.TRAVIS_DIST !== 'focal' ? test : test.skip;

	testFn(...);
}
```

test/stream.ts
```diff
@@ -178,7 +182,7 @@ test('redirect response contains old url', withServer, async (t, server, got) =>
 	server.get('/redirect', redirectHandler);
 
 	const {requestUrl} = await pEvent(got.stream('redirect'), 'response');
-	t.is(requestUrl, `${server.url}/redirect`);
+	t.is(requestUrl as string, `${server.url}/redirect`);
```
szmarczak, line 185: Is this necessary?

test/timeout.ts
```diff
@@ -273,9 +273,9 @@ test.serial('connect timeout (ip address)', withServerAndFakeTimers, async (t, _
 	);
 });
 
-test.serial('secureConnect timeout', withServerAndFakeTimers, async (t, _server, got, clock) => {
+test.serial('secureConnect timeout', withHttpsServer({}, true), async (t, _server, got, clock) => {
```
szmarczak, line 276: Avoid this. It has ambiguous meaning. Does the `true` mean that we're using the HTTPS server? What if it's false? Does it mean that the HTTPS is turned off?

It would be best to include it as an option inside the object (first argument).

test/timeout.ts
```diff
@@ -293,7 +293,7 @@ test.serial('secureConnect timeout', withServerAndFakeTimers, async (t, _server,
 			retry: 0
 		}).on('request', (request: http.ClientRequest) => {
 			request.on('socket', () => {
-				clock.runAll();
+				clock?.runAll();
```
szmarczak, line 296: Why this change?

test/unix-socket.ts
```diff
@@ -31,10 +31,11 @@ if (process.platform !== 'win32') {
...
 
 	test('throws on invalid URL', async t => {
-		await t.throwsAsync(got('unix:', {retry: 0}), {
-			instanceOf: got.RequestError,
-			code: 'ENOTFOUND'
-		});
+		try {
+			await got('unix:', {retry: 0});
+		} catch (error) {
+			t.regex(error.code, /(ENOTFOUND|EAI_AGAIN)/);
```
szmarczak, line 37: Why this change?

szmarczak: Maybe the VM needs to be privileged? (run as root)?

szmarczak:
1. Can we use GitHub Actions instead of Travis?
2. It'd be best if this PR was split up into two: 1) CI improvements, 2) test improvements.

test/stream.ts
Giotino: Sorry, I forgot to remove it

test/error.ts
Giotino: On Ubuntu 18.04 and 20.04 it was throwing that error, meaning a DNS Lookup Timeout.

I think this should be ok for this test.

test/unix-socket.ts
Giotino: I think this test should be rewrote, `ENOTFOUND` is thrown because `unix:` is viewed as `http://unix/`

Giotino:
> 1. Can we use GitHub Actions instead of Travis?

Yes, we are going have less platforms to run the tests, but that shouldn't be a problem.

> 2. It'd be best if this PR was split up into two: 1) CI improvements, 2) test improvements.

I opened this PR to make the `master` branch able to run on CITGM.
The fixed tests and the speed-up due to the new HTTP Test Server were aiming to a successful run on CIRGM.
The increased CI targets were needed for us to look at issues on other platforms.

---

I would like to have this PR on `master` as soon as possible, then I'm going to open two new PRs with the aim of improving the Test Environment (as you can see the new modifications are a little messy, I would like to clean it up in the new PR) and the CI Environment (aka. migration/trial on Gitlab Action).

Issue pointed out by you, such as:
- The meaning of `true` in `withHttpsServer({}, true)`
- "Maybe let's do `withHttpAndHttpsServer`?"
- Why `clock?.runAll();`
- GitHub Actions

Are going to be addressed in the new PRs.

test/timeout.ts
Giotino: In the HTTPS Test Server the clock is optional.
I don't see any reason to create a clock for each test, especially if there are no tests that require a "local" clock (not installed by `FakeTimers.install()`).

test/http.ts
Giotino: [`7625a7d` (#1461)](https://github.com/sindresorhus/got/pull/1461/commits/7625a7d691e532456fe080780de5af80987f5ffb)

test/stream.ts
Giotino: [`e7be039` (#1461)](https://github.com/sindresorhus/got/pull/1461/commits/e7be0398c6641f60e57f98f75b2812f97ad17611)

sindresorhus:
I don't really see the point of testing on so many Linux combinations. It's unlikely to make a difference but will make builds slower. This is especially true for macOS. I think we should limit it to latest macOS and Windows, and also Ubuntu Bionic (18.04) and Ubuntu Focal (20.04).

.travis.yml
```diff
@@ -1,8 +1,71 @@
...
+  
 node_js:
-  - '14'
-  - '13'
-  - '12'
-  - '10'
+  - 14
+  - 13
+  - 12
+  - 10
```
sindresorhus, line 17: These should be quotes. Numbers are funky in YAML and it's better to be safe and use strings, as they're actually strings.

.travis.yml
```diff
@@ -1,8 +1,71 @@
...
+  # macOS 10.14.6
+  - xcode11.3
+  
 node_js:
-  - '14'
-  - '13'
-  - '12'
-  - '10'
+  - 14
+  - 13
```
sindresorhus, line 15: I suggest we remove this. No one is using 13 and it's enough to test on 14.

test/unix-socket.ts
sindresorhus: Also no need to wrap it in parens

test/timeout.ts
Giotino: https://github.com/sindresorhus/got/pull/1461#issuecomment-691153647

test/timeout.ts
Giotino: https://github.com/sindresorhus/got/pull/1461#issuecomment-691153647

test/unix-socket.ts
szmarczak: > I think this test should be rewritten, ENOTFOUND is thrown because unix: is viewed as http://unix/

It shouldn't throw. If the hostname is `unix`, then Got does this:

https://github.com/sindresorhus/got/blob/0e0eefdd8538e910cba6d89df5a3268743aa7ec8/source/core/index.ts#L2311-L2324

test/helpers/with-server.ts
```diff
@@ -4,25 +4,24 @@ import is from '@sindresorhus/is';
...
 
 	// Re-enable body parsing to investigate https://github.com/sindresorhus/got/issues/1186
-	const server = await createTestServer(is.plainObject(testServerOptions) ? testServerOptions : {
+	const server = await createHttpTestServer(is.plainObject(testServerOptions) ? testServerOptions : {
 		bodyParser: {
 			type: () => false
-		},
-		certificate: 'example.com'
-	}) as ExtendedTestServer;
+		} as any
```
szmarczak, line 23: Why `any`?

test/http.ts
```diff
@@ -309,19 +309,23 @@ test('DNS IPv4', withServer, async (t, server, got) => {
...
 			dnsLookupIpVersion: 'ipv6'
 		});
 
 		t.true(isIPv6(response.body));
-	});
-}
+	} catch (error) {
+		// If network unreachable (happens on TravisCI with Ubuntu Bionic and Focal)
+		t.is(error.code, 'ENETUNREACH');
+	}
```
szmarczak, line 325: I guess this is moot now?

test/helpers/with-server.ts
Giotino: It was an easy fix, when I was working on it I had a problem with the typing of `bodyParser` and had no time to fix it.

I left it as `any` because that part is going to be changed in the "General improvements of the Test Environment" PR that I'm going to open after the merging of this.

test/unix-socket.ts
Giotino: I think this needs more investigation

test/http.ts
Giotino: This is still valid because we are testing on Focal (20.04) on Travis CI.
When we switch to Gitlab CI we will need to understand how they treat the IPv6 addresses.

test/agent.ts
Giotino: https://github.com/sindresorhus/got/pull/1461#issuecomment-691153647

test/https.ts
```diff
@@ -421,7 +421,7 @@ test('invalid key passphrase', withHttpsServer(), async (t, server, got) => {
 			await request;
 			t.fail();
 		} catch (error) {
-			t.true((error.message as string).includes('bad decrypt'));
+			t.regex(error.message, /bad decrypt/);
```
sindresorhus, line 424: The `.includes` was was better. No point in using regex if it's not necessary.

Giotino: Using the `regex(a, b)` you can see the message if regex doesn't match, with `true(a.includes(b))` you only see something like "expected 'true' found 'false'".

That's the reason why I changed it (take a look at this: https://travis-ci.com/github/sindresorhus/got/jobs/384356165)

sindresorhus: You can use the second argument to see the actual value.

```js
t.true((error.message as string).includes('bad decrypt'), error.message);
```

test/http.ts
szmarczak: > When we switch to Gitlab CI

Do you mean GitHub?

szmarczak, line 325: Use conditional `t.failing` instead. That way we'll know what tests should be fixed.

szmarczak (changes requested):

test/http.ts
Giotino: > > When we switch to Gitlab CI
> 
> Do you mean GitHub?

Yeah, sorry, I meant Github Actions

test/timeout.ts
szmarczak: Then you need to use `clock!.runAll();` since we expect `clock` to be defined.

test/http.ts
Giotino: I fail to understand how do you want to use `t.failing` can you make an example, thanks.

sindresorhus: @szmarczak meant `test.failing()`.

https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests

szmarczak: Yes. Sorry for the confusion 😅

Giotino: Test test was already using a conditional `test`: `testIPv6`. I forgot to remove the `try, catch`. The thing is that it isn't failing...

szmarczak: What about `test/stream.ts` and `test/unix-socket.ts`?

sindresorhus:
Merging this as there was a request to get it merged fast. But please address the feedback given here in the follow-up PR.

Giotino:
> Merging this as there was a request to get it merged fast. But please address the feedback given here in the follow-up PR.

Sure, I'm already preparing it, a draft is coming soon.

PR #1464, opened 2020-09-15
General improvements to the test environment

Giotino:
Following the previous PR here's the promised "General improvements to the Test Environment".

This PR is going to address:
- Clean up the test server helper
- Find a clean way to have 2 servers for a single test
- Conditional testing
- Tests review

If you have more suggestions feel free to write it in the comments, this PR is aiming to improve tests, everything regarding that matter is welcome.

### Clean up the test server helper

Achieving a separation between the HTTP server (which receive requests and send the responses) and the actual framework (Express, which use the requests and craft the responses) we can replace the 2 component independently and we only need to edit a single thing to extend the test server. (this code is already present in the PR)

### Find a clean way to have 2 servers for a single test

As pointed out by @szmarczak, this is ugly:
```js
test('', withHttpServer(), async (t, serverHttp) => {
	await withHttpsServer()(t, async (t, serverHttps, got) => {
	}
});
```

My idea whould be having something like this:
```js
test('', withDoubleServer(withHttpServer(), withHttpsServer()), async (t, server1, server2, got1, got2) => {});
```

Please give me some feedback: @szmarczak @sindresorhus 

### Conditional testing

Instead of "making the tests disappear" with if statements the test should have a conditional `test.failing` or `test.skip` instead of `test`, when needed. (this code is already present in the PR)

`test.failing` where the test is failing (and it shouldn't), `test.skip` where the test has no reason to run (e.g. unsupported platform).

### Tests review

A general review of the tests.

Problems found until now:
- unix-socket.ts `throws on invalid URL`: The test is passing, but it treats `unix:` as `http://unix/` (like a hostname and not a unix path)


szmarczak:
> Find a clean way to have 2 servers for a single test

```ts
​test​(​''​,​ ​withDoubleServer​(​withHttpServer​(​)​,​ ​withHttpsServer​(​)​)​,​ ​async​ ​(​t​,​ ​server1​,​ ​server2​,​ ​got1​,​ ​got2​)​ ​=>​ ​{​}​)​;
```

This is still ugly. Why not:

```ts
​test​(​''​,​ ​​withHttpServer​(​)​,​ ​withHttpsServer​(​)​,​ ​async​ ​(​t​,​ ​server1​,​ ​got1​,​ server2, got2 ​=>​ ​{​}​)​;
```

Or this would've been even better:

```ts

test​(​''​,​ ​​withServer(​{http: true, https: true})​,​ ​async​ ​(​t​,​ ​server, got) ​=>​ ​{​}​)​;
```

szmarczak:
> unix-socket.ts throws on invalid URL: The test is passing, but it treats unix: as http://unix/ (like a hostname and not a unix path)/

This not a problem nor a bug. See https://github.com/sindresorhus/got#unix-domain-sockets

Giotino:
> > unix-socket.ts throws on invalid URL: The test is passing, but it treats unix: as [http://unix](http://unix/) (like a hostname and not a unix path)/
> 
> This not a problem nor a bug. See [sindresorhus/got#unix-domain-sockets](https://github.com/sindresorhus/got#unix-domain-sockets)

Yeah, but in my opinion the test name "throws on invalid URL" does not reflect the test content.

`unix:` is treated as an HTTP-over-TCP URL instead of an HTTP-over-UNIXSOCKET URL. 

The name should be something like `"unix" is a valid hostname`

Giotino:
I also thought about something like this:
> This is still ugly. Why not:
> 
> ```ts
> ​test​(​''​,​ ​​withHttpServer​(​)​,​ ​withHttpsServer​(​)​,​ ​async​ ​(​t​,​ ​server1​,​ ​got1​,​ server2, got2 ​=>​ ​{​}​)​;
> ```
> 

I don't know how I'm going to pull it off with Typescript, but I will try.

szmarczak:
> I don't know how I'm going to pull it off with Typescript, but I will try.

No worries, for now you can just leave it as it is. I'll try to do this.

szmarczak:
> The name should be something like "unix" is a valid hostname

I understand now. It should throw on `http://unix/` AND on `unix:` because the socket path is missing.

sindresorhus:
How common is it to need to servers for a single test? If it's only a few cases, I would prefer the boilerplate over a clever abstraction.

test/helpers/http-server.ts
```diff
@@ -0,0 +1,157 @@
...
+import https = require('https');
+import net = require('net');
+import express = require('express');
+import pify = require('pify');
+import pem = require('pem');
+import bodyParser = require('body-parser');
+import tempy = require('tempy');
+
+export type BaseServerOptions = {
+	protocol: 'HTTP' | 'HTTPS' | 'socket';
```
sindresorhus, line 11: I think these should be lowercase. They are used as identifiers, so it's ok that they don't match the actual acronym.

Giotino:
> How common is it to need to servers for a single test? If it's only a few cases, I would prefer the boilerplate over a clever abstraction.

Currently we have only 5 tests.
4 are about redirects from HTTP to HTTPS and vice versa
1 is about the redirect from a port to another on the same host and protocol

test/helpers/http-server.ts
```diff
@@ -0,0 +1,157 @@
...
+	commonName?: string;
+	days?: number;
+};
+export type HttpSocketServerOptions = BaseServerOptions & {
+	protocol: 'socket';
+};
+export type ServerOptions = HttpServerOptions | HttpsServerOptions | HttpSocketServerOptions;
+
+export interface BaseExtendedServer extends express.Express {
+	close: () => Promise<any>;
```
sindresorhus, line 28: This should be better typed than `any`.

sindresorhus (changes requested):

Giotino:
Little note: currently types are a little messy, I'm trying to make Typescript infer as much as possible, before this PR is marked "Ready for review" all the types should be ok.

test/helpers/http-server.ts
Giotino: Actually, I put `any` because I don't care of the result, maybe `unknown` is better?

szmarczak:
Did you do `git pull --rebase`?

package.json
```diff
@@ -107,6 +109,7 @@
 		}
 	},
 	"nyc": {
+		"extends": "@istanbuljs/nyc-config-typescript",
```
szmarczak, line 112: Why this change?

test/helpers/http-server.ts
szmarczak: `Promise<void>`

Giotino:
> Did you do `git pull --rebase`?



Cherry picked that commit...

szmarczak:
Always do `git pull --rebase` when master changes. That way you'll prevent duplicating commits like this one:

image

szmarczak:
Can you clone this branch to e.g. `test-env-2`? Then squash all the commits in **this** branch and force push to clean up. Tip: never do force push unless you want to revert a commit or rebase failed or GitHub messes up (this sometimes happens!).

package.json
Giotino: This was to support Typescript for the code coverage, but it works even without, just by enabling the source map generation.

szmarczak:
Here's a nice way to support multiple servers:

```ts
import test, {ExecutionContext} from 'ava';
import {AddressInfo} from 'net';
import http = require('http');

type Implementation<Context = unknown,> = (t: ExecutionContext<Context>, ...rest: any[]) => Promise<void> | void;
type ChainElement<Context = unknown> = (t: ExecutionContext<Context>, run: Implementation<Context>, ...rest: unknown[]) => Promise<void>;
type Chain<Context = unknown> = (t: ExecutionContext<Context>, elements: ChainElement<Context>[], implementation: Implementation<Context>) => Promise<void>;

const withServer: (module?: {createServer: typeof http.createServer}) => ChainElement = (module = http) => async (t, run, ...rest) => {
	const server = module.createServer();

	await new Promise(resolve => server.listen(resolve));

	try {
		await run(t, ...rest, server);
	} finally {
		await new Promise(resolve => server.close(resolve));
	}
};

const chain: Chain = async (t, elements, implementation) => {
	const {length} = elements;
	let index = 0;

	const iterate = async (t: any, ...args: any[]) => {
		return elements[index++](t, index === length ? implementation : iterate, ...args);
	};

	return iterate(t);
};

test('asdf', chain, [withServer(), withServer()], async (t, http1: http.Server, http2: http.Server) => {
	http1.on('request', (request, response) => {
		request.resume();
		response.end('http 1');
	});

	http2.on('request', (request, response) => {
		request.resume();
		response.end('http 2');
	});

	console.log(`http://localhost:${(http1.address() as AddressInfo).port}`);
	console.log(`http://localhost:${(http2.address() as AddressInfo).port}`);

	await new Promise(resolve => setTimeout(resolve, 60000));

	t.pass();
});
```

Giotino:
What about

```js
test('asdf', chain([withServer(), withServer()]), async (t, http1: http.Server, http2: http.Server) => {});
```

instead of

```js
test('asdf', chain, [withServer(), withServer()], async (t, http1: http.Server, http2: http.Server) => {});
```
?

To me it seems more correct, if we consider each option of `test` as a macro, as documented, I think `[withServer(), withServer()]` makes no sense alone.

sindresorhus:
> test('asdf', chain([withServer(), withServer()]),

@Giotino That looks better to me.

szmarczak:
> as documented, I think [withServer(), withServer()] makes no sense alone.

It makes sense to me. Please take a look at: https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#reusing-test-logic-through-macros

As you can see, it's a free choice to set up the arguments, so both is right.

> That looks better to me.

I'm good with both. I'd just get rid of the `[]` so it would look like the original proposal.

sindresorhus:
@Giotino Still interested in finishing this? It touches a lot of files, so the longer it's open, the more conflicts it will get.

Giotino:
Sorry, lately I've been really busy. I want to finish this PR, but currently I'm unable to do so.
I would like to leave this PR open (as a draft) and, when I am going to finally find some time, decide if it's better to continue on this PR or open a new one.

sindresorhus:
Ok, no worries.

sindresorhus:
I'm going to close this for now as it's not moving forward. We can reopen it again if it receives some activity.

Issue #1466, opened 2020-09-16
Got 12 planning


sindresorhus:
There are a few issues: https://github.com/sindresorhus/got/milestone/9

Anything else we want to change that is risky or breaking? It doesn't have to be a big breaking release though.

- [x] Rename `lookup` option to `dnsLookup`
- [ ] Throw an error here: https://github.com/sindresorhus/got/blob/2b40da2247016f2f4f2b2049a40aa51265161259/source/create.ts#L138-L149

// @szmarczak @Giotino

sindresorhus:
@szmarczak I believe there was a breaking change we wanted to do to improve Node.js 15 compatibility?

szmarczak:
Yes. It would introduce a new error event: `retryError`.

szmarczak:
Should there be a separate branch for Got 12 development?

sindresorhus:
I think https://github.com/sindresorhus/got/issues/1353 would also be breaking.

sindresorhus:
> Should there be a separate branch for Got 12 development?

No, we can get out one more minor release and then switch the focus of master to v12.

Giotino:
What about this? https://github.com/sindresorhus/got/issues/1306#issuecomment-639687180

By not sending Got `options` directly to the request module we can have more control on the request, and we can swap the underlying request module without breaking Got (e.g. testing Undici and use it if it suits our needs)

sindresorhus:
> What about this? #1306 (comment)

👍🏻 Yes, we can do that one too.

szmarczak:
Unless there's anything else I think we can release `11.7.0`.

sindresorhus:
> Unless there's anything else I think we can release 11.7.0.

Done: https://github.com/sindresorhus/got/releases/tag/v11.7.0

We can start preparing for Got 12 now.

geminigeek:
hi,
kindly have a provision to send custom error status code. i have a use case steps below  

1. got client calls an express endpoint 
2. the express endpoint calls web api and process the request  and send reply ( got client gets this reply)
3. i want to send custom error status code  related to post processing errors in express app to got client

Giotino:
> Should there be a separate branch for Got 12 development?

I think we should use a new branch before the release, in order to continue patching the current release, if needed.

sindresorhus:
I'd rather do a `v11` branch for that, if we need to backport.

kanongil:
[...]

PR #1483, opened 2020-10-03
Add `lookup` option documentation

Giotino:
Added the `lookup` option documentation. I've not specified the function signature because it's already well documented on the Node.JS documentation page and I thought that it would be redundant to copy and paste it on the Got documentation.

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -658,7 +658,7 @@ await got('https://api6.ipify.org', {
 ###### lookup
 
 Type: `Function`
-Default `dns.lookup`
+Default: [`dns.lookup`](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback)
 
 Custom lookup function.
 
--- a/readme.md
+++ b/readme.md
@@ -660,7 +660,7 @@ await got('https://api6.ipify.org', {
 Type: `Function`\
 Default: [`dns.lookup`](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback)
 
-Custom lookup function.
+Custom DNS resolution logic.
 
 The function signature is the same as [`dns.lookup`](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback).
 
--- a/readme.md
+++ b/readme.md
@@ -662,8 +662,7 @@ Default `dns.lookup`
 
 Custom lookup function.
 
-The function signature is the same of `dns.lookup`, please refer to the Node.JS documentation page for more informations https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
-
+The function signature is the same of `dns.lookup`, please refer to the Node.JS documentation page for more information https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
 
 ###### request
 
--- a/readme.md
+++ b/readme.md
@@ -657,7 +657,7 @@ await got('https://api6.ipify.org', {
 
 ###### lookup
 
-Type: `Function`
+Type: `Function`\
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -662,8 +662,7 @@ Default `dns.lookup`
 
 Custom lookup function.
 
-The function signature is the same of `dns.lookup`, please refer to the Node.JS documentation page for more informations https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
-
+The function signature is the same of `dns.lookup`, please refer to the Node.JS documentation page for more information https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
```
szmarczak, line 665: Node.JS -> Node.js

readme.md
```diff
@@ -655,6 +655,15 @@ await got('https://api6.ipify.org', {
 });
 ```
 
+###### lookup
+
+Type: `Function`
+Default `dns.lookup`
```
szmarczak, line 661:
```suggestion
Default: `dns.lookup`
```

also it may be worth to link dns.lookup to the Node.js doc page

readme.md
```diff
@@ -655,6 +655,15 @@ await got('https://api6.ipify.org', {
...
+
+Custom lookup function.
+
+The function signature is the same of `dns.lookup`, please refer to the Node.js documentation page for more information https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
```
sindresorhus, line 665:
```suggestion
The function signature is the same as [`dns.lookup`](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback).
```

readme.md
```diff
@@ -655,6 +655,15 @@ await got('https://api6.ipify.org', {
 });
 ```
 
+###### lookup
+
+Type: `Function`
+Default `dns.lookup`
+
+Custom lookup function.
```
sindresorhus, line 663: Can we maybe succinctly explain what "lookup" means?

readme.md
```diff
@@ -655,6 +655,15 @@ await got('https://api6.ipify.org', {
 });
 ```
 
+###### lookup
+
+Type: `Function`
```
sindresorhus, line 660:
```suggestion
Type: `Function`\
```

readme.md
```diff
@@ -655,6 +655,15 @@ await got('https://api6.ipify.org', {
 });
 ```
 
+###### lookup
```
sindresorhus, line 658:
```suggestion
###### dnsLookup
```

?

sindresorhus (changes requested):

readme.md
Giotino: Actually is called `lookup` should we change the name?

Since we are near Got 12 I think the name should be changed in it.

readme.md
Giotino: Maybe something like: "Custom lookup function. Use it to implement a custom DNS resolution logic"

readme.md
sindresorhus: Yeah, I was asking if you should rename it. I think we should.

readme.md
sindresorhus: How about:

> Custom DNS resolution logic.

readme.md
Giotino: Let's leave it as `lookup` and then change it in Got 12. If there are going to be a reasonable number of dns related options we could create a `dns` option that includes all of them.

sindresorhus: Added it to https://github.com/sindresorhus/got/issues/1466

sindresorhus (approved):

PR #1484, opened 2020-10-03
Parse JSON without conversion to string

xamgore:
Node.js and V8 allows to parse JSON from buffer, so an intermediate conversion to string is not required.

Though typescript doesn't have correct typing due to historic issues. I would like do extend JSON interface with a method that allows Buffer as its first argument. Will it be ok for you?

### DIFF
--- a/source/as-promise/parse-body.ts
+++ b/source/as-promise/parse-body.ts
@@ -14,7 +14,7 @@ const parseBody = (response: Response, responseType: ResponseType, parseJson: Pa
 		}
 
 		if (responseType === 'json') {
-			return rawBody.length === 0 ? '' : parseJson(rawBody.toString());
+			return rawBody.length === 0 ? '' : parseJson(rawBody);
 		}
 
 		if (responseType === 'buffer') {


Giotino:
Can you please provide some links to the Node.JS documentation about `JSON.parse` with a buffer argument.

The only things I can't find on the Node.JS documentation are links to developer.mozilla.org which only talks about strings (since there is no `Buffer` in Javascript). If the only references are source code snippets I don't think is a good idea to trust them since they can be changed at any time.

Giotino:
Also the current `parseJson` function is already documented as `(text: string) => unknown` and we cannot change its signature in Got 11, since it would be a breaking change (but we are working on Got 12).

https://github.com/sindresorhus/got#parsejson

xamgore:
> Can you please provide some links to the Node.JS documentation about `JSON.parse` with a buffer argument.

NodeJS doesn't have any notes in the documentation, as it's a concern of the language itself. http://www.ecma-international.org/ecma-262/5.1/#sec-15.12.2 states:

```
parse ( text [ , reviver ] )

1. Let JText be ToString(text).
2. Parse JText using the grammars in 15.12.1. (...)
```

`ToString(text)` is `text.toString` in case of objects, thus [`buffer.toString()`](https://github.com/nodejs/node/blob/e7bb8b6dc08b06edcad6749d01bd5ee72ae15123/lib/buffer.js#L759) is called. But now I see, it makes no difference.

---

> Also the current parseJson function is already documented as `(text: string) => unknown` and we cannot change its signature in Got 11, since it would be a breaking change.

Indeed. Though, one could possibly use a streaming parser (like `xml2js` or [`node-xml`](https://github.com/dylang/node-xml) but with json) that extracts only a small set of required fields.

Giotino:
In Got 12 we can change the signature of `parseJson` from `(text: string) => unknown` to `(text: Buffer) => unknown`.

But is there a real advantage by doing so?

Can you make some real world examples where this improvement matter.

You wrote about a JSON stream parser, but with the Buffer we already have all the data, so I don't think there would be a real difference.

xamgore:
If Got had support for XML (via `parseXml: (input: string | Buffer) => unknown`) I could say [`node-xml2json`](https://github.com/buglabs/node-xml2json) could be a good case. It is a wrapper around [`node-expat`](https://github.com/astro/node-expat), which in turn is a C++ addon, implementing the fastest streaming XML parser. The parser [accepts `Buffer`](https://github.com/astro/node-expat/blob/master/node-expat.cc#L136). An extra `toString` conversion is not required.

The same is true for JSON, there is a really fast JSON streaming parser, [`simdjson`](https://github.com/simdjson/simdjson), with [nodejs bindings](https://github.com/luizperes/simdjson_nodejs) which can be used to serve the same purpose: I want to pick _some data_ from a buffer in a fast manner. I could still use `passToStreamingParser(got.get().buffer())` though.

As far as I know, there are no performant streaming parsers implemented in C++ that can be used with node.js streams. String parsers written in javascript are too slow. My personal use case is "I'd like to fetch & parse big amounts of JSON/XML and extract some fields from it without creating the full object".

szmarczak:
The `parseJson` function expects a string. If you really need to use a buffer, go for `.buffer()` instead. If you need streams, go for `got.stream()`.

We won't probably support XML as there's no native solution.

PR #1487, opened 2020-10-05
Fix `url` not being reused on retry in rare case

qlba:
#### Checklist

Currently, when request URL is given in parts, `options.port` is removed from `options` after the initial request. Then `retryWithMergedOptions` attempts to connect to the default port.

This test fails after https://github.com/sindresorhus/got/commit/52de13bbdcd94db58ffcf39f87293af9249594c1

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1645,6 +1645,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				delete options.port;
 			}
 
+			if ('protocol' in options) {
+				delete options.protocol;
+			}
+
 			// Make it possible to change `options.prefixUrl`
 			let {prefixUrl} = options;
 			Object.defineProperty(options, 'prefixUrl', {
--- a/test/hooks.ts
+++ b/test/hooks.ts
@@ -940,6 +940,7 @@ test('async afterResponse allows to retry with allowGetBody and json payload', w
 		retry: 0,
 		throwHttpErrors: false
 	});
+
 	t.is(statusCode, 200);
 });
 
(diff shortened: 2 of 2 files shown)


Giotino:
@szmarczak these behaviors are due to the fact that we work on the same `options` object and it's becoming a nearly impossible to track all the Got-made modifications and roll them back when needed. As I previously suggested, I think the best option is to have the `options` object given by the user and another one crafted specifically for the underlying HTTP client.
In this way the client `options` object can be discarded after the request and re-crafted before retry.

szmarczak:
> @szmarczak these behaviors are due to the fact that we work on the same options object and it's becoming a nearly impossible to track all the Got-made modifications and roll them back when needed.

Rolling back is unnecessary. Node.js should accept them without any modifications after they're normalized (except the `timeout` and `agent` options). The destination options are normalized into `options.url`. I had an idea to drop **all** legacy Url options in favour of WHATWG URL. I still stand by this and I even left a TODO in the code to do so. That way we would remove lots of undiscovered bugs related to legacy Url options. I wanted to do this in Got 10 but that would be too many breaking changes in a release.

> As I previously suggested, I think the best option is to have the options object given by the user and another one crafted specifically for the underlying HTTP client.
In this way the client options object can be discarded after the request and re-crafted before retry.

I'm :-1: on recreation. The performance impact would be huuuuge. I'm not quite sure how we could avoid it but using a Proxy object to normalize options is worth trying I think.

szmarczak:
> When normalizing and preparing the options for the request Got pollute the user options and then, trying to clean it up, Got mess up.

The options object is meant to be handled and modified by Got, and we should minimize the amount of optional options (including the legacy Url ones) in order to avoid confusion.

szmarczak:
In regards to the OP's issue I'll see where's the problem located. `options.port` is normalized into `options.url.port` and it should use that.

PR #1495, opened 2020-10-09
Set default value for an options object

mrmlnc:
Hello,

I ran into a problem that after the client extension by the `.extend` method, the `init` hook gets an `undefined` instead of an object as options.

The following example illustrates the problem.

```js
const got = require('got').default;

const client = got.extend({
	hooks: {
		init: [
			(options) => {
				console.dir({ options }, { colors: true });
			}
		]
	}
});

(async () => {
	await client('https://google.com'); // { options: undefined }
	await client('https://google.com', {}); // { options: {} }
	await client({ url: 'https://google.com' }); // { options: { url: <value> }}
})();
```

Looks like a bug, because I see that the `init` hook should get an `Options` object according to the signature in the types.

https://github.com/sindresorhus/got/blob/87dadd53bdef4184ecba1b263a726bba4674c23c/source/core/index.ts#L103

So, this PR fixes the problem. If you accept this change, I can make a backport to the `v11` branch.

#### Checklist

### DIFF
--- a/source/create.ts
+++ b/source/create.ts
@@ -125,7 +125,7 @@ const create = (defaults: InstanceDefaults): Got => {
 	}));
 
 	// Got interface
-	const got: Got = ((url: string | URL, options?: Options, _defaults?: Defaults): GotReturn => {
+	const got: Got = ((url: string | URL, options: Options = {}, _defaults?: Defaults): GotReturn => {
 		let iteration = 0;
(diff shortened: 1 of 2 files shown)


mrmlnc:
Looks like CI fails due to last update of node.js. Works fine under `12.18.4` and breaks after `12.19.0`.

szmarczak:
@mrmlnc Can you please open another issue for this?

PR #1518, opened 2020-11-02
Add read timeout

### LINKED ISSUE #1512: Response read timeout
#### Describe the bug

 - Node.js version: 14.14.0
 - OS & version: Ubuntu 20.04

We've noticed that the timeout isn't applied very consistently when using streams and piping the stream to something slow. It's not particularly obvious if the timeout is meant to account for this.

#### Actual behavior

Depending on the backpressure of the stream being piped to, the stream returned by `got.stream()` may end, error, or hang.

#### Expected behavior

I expected the timeout to trigger consistently rather than depending on the backpressure.

#### Code to reproduce

```js
const http = require('http');
const { finished, Readable, Writable } = require('stream');
const got = require('got');

function createReadStream(size) {
  return new Readable({
    read(n) {
      const chunkSize = Math.min(size, n);
      size -= chunkSize;
      if (chunkSize) {
        setImmediate(() => {
          this.push(Buffer.alloc(chunkSize, 'a'));
        });
      } else {
        this.push(null);
      }
    }
  });
}

function createBlackHole() {
  return new Writable({
    write() {}
  });
}

const server = http.createServer();

server.on('request', (request, response) => {
  const size = parseInt(request.url.slice(1)) || Infinity;

  response.statusCode = 200;
  response.setHeader('Content-Type', 'application/octet-stream');
  createReadStream(size).pipe(response);
});

async function makeRequest(size) {
  const timeout = 2000;
  const stream = got.stream(`http://localhost:3000/${size}`, { timeout });

  console.log(`\nMaking request for stream of ${size} bytes...`);

  return new Promise((resolve, reject) => {
    const timer = setTimeout(() => {
      console.warn('Stream never ended or errored!');
      resolve();
    }, timeout + 1000);

    const cleanup = finished(stream, err => {
      cleanup();
      clearTimeout(timer);

      if (err) {
        console.log(`Stream errored: ${err.message}`);
      } else {
        console.log('Stream ended');
      }

      resolve();
    });

    stream.pipe(createBlackHole());
  });
}

async function runTests() {
  // At a small number of bytes, the response emits end and the request emits
  // close, and the stream will be finished. No timeouts.
  await makeRequest(1024);

  // At a number of bytes slightly bigger than 16kb, the request will emit
  // close which will cause got to clear the request timeout. The response
  // doesn't emit end and the stream will never be finished. Our fallback timer
  // will expire.
  await makeRequest(1024 * 20);

  // At a number of bytes larger than 32kb, the request will not emit close and
  // the response will not emit end. Got's request timeout will fire and destroy
  // the request.
  await makeRequest(1024 * 40);
}

server.listen(3000, err => {
  if (err) {
    return console.error(err);
  }

  runTests().then(() => {
    process.exit();
  });
});
```

**Output:**

```
Making request for stream of 1024 bytes...
Stream ended

Making request for stream of 20480 bytes...
Stream never ended or errored!

Making request for stream of 40960 bytes...
Stream errored: Timeout awaiting 'request' for 2000ms
```

#### Checklist

---

I'm willing to try to open a PR to address this, but I want to be sure that this isn't considered expected behavior since it has as much to do with the stream being piped to as it does got itself. While that's true, I think the expectation when passing `timeout` is that the request can't hang indefinitely.

szmarczak:
Fixes #1512

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -518,7 +518,10 @@ This also accepts an `object` with the following fields to constrain the duratio
 - `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
 - `response` starts when the request has been written to the socket and ends when the response headers are received.
 - `send` starts when the socket is connected and ends with the request has been written to the socket.
-- `request` starts when the request is initiated and ends when the response's end event fires.
+- `request` starts when the request is initiated and ends when the response's `end` event fires.
+- `read` starts when the `response` event is emitted and ends when the response's `end` event fires.
+
+**Note:** The `read` timeout is blocked by https://github.com/nodejs/node/issues/35923
 
 ###### retry
 
--- a/readme.md
+++ b/readme.md
@@ -519,7 +519,7 @@ This also accepts an `object` with the following fields to constrain the duratio
 - `response` starts when the request has been written to the socket and ends when the response headers are received.
 - `send` starts when the socket is connected and ends with the request has been written to the socket.
 - `request` starts when the request is initiated and ends when the response's `end` event fires.
-- `read` starts when the `response` event is emitted and ends when the response's `end` event fires.
+- ~~`read` starts when the `response` event is emitted and ends when the response's `end` event fires.~~
 
 **Note:** The `read` timeout is blocked by https://github.com/nodejs/node/issues/35923
 
--- a/source/core/utils/timed-out.ts
+++ b/source/core/utils/timed-out.ts
@@ -13,11 +13,13 @@ interface TimedOutOptions {
 
 export interface Delays {
 	lookup?: number;
+	socket?: number;
 	connect?: number;
 	secureConnect?: number;
-	socket?: number;
-	response?: number;
 	send?: number;
+	response?: number;
+	read?: number;
+
 	request?: number;
 }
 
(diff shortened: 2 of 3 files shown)


szmarczak:
Blocked by https://github.com/nodejs/node/issues/35923

sindresorhus:
Doesn't look like the Node.js issue will be fixed anytime soon. Could you do a new PR with just the tests as `test.failing()`? Then we can close this PR for now and link to it from https://github.com/sindresorhus/got/issues/1512.

readme.md
```diff
@@ -518,7 +518,10 @@ This also accepts an `object` with the following fields to constrain the duratio
 - `socket` starts when the socket is connected. See [request.setTimeout](https://nodejs.org/api/http.html#http_request_settimeout_timeout_callback).
 - `response` starts when the request has been written to the socket and ends when the response headers are received.
 - `send` starts when the socket is connected and ends with the request has been written to the socket.
-- `request` starts when the request is initiated and ends when the response's end event fires.
+- `request` starts when the request is initiated and ends when the response's `end` event fires.
+- `read` starts when the `response` event is emitted and ends when the response's `end` event fires.
```
sindresorhus, line 522:
```suggestion
- ~~`read` starts when the `response` event is emitted and ends when the response's `end` event fires.~~
```

?

source/core/utils/timed-out.ts
```diff
@@ -13,11 +13,13 @@ interface TimedOutOptions {
...
 	lookup?: number;
+	socket?: number;
 	connect?: number;
 	secureConnect?: number;
-	socket?: number;
-	response?: number;
 	send?: number;
+	response?: number;
+	read?: number;
+
```
sindresorhus, line 22:
```suggestion
```

PR #1521, opened 2020-11-07
Improve option normalization

### LINKED ISSUE #1307: Invert the `methodRewriting` option logic
#### Describe the bug

 - Node.js version: 12.18.0
 - OS & version: Linux 5.6

According to the HTTP spec in [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3), HTTP clients should NOT redirect on 3XX response codes UNLESS the original request was a GET or a HEAD.

Got's default behavior causes it to repeat the original request (even if it's something like a POST) to the redirect URL, breaking very common webserver upload patterns.

#### Actual behavior

Common webserver behavior is to accept a POST to create some sort of resource, then reply with a 302 to point the user to the newly creating resource.

Got's default redirection behavior is to simply perform the same exact request again on some redirect code, regardless of the initial request method. So a POST request with the same body was then performed to the new URL from the location header of the redirect, which almost always will cause an error, as the redirected location generally will not accept a POST request, and certainly not expecting the same body.

#### Expected behavior

HTTP spec indicates that you SHOULD NOT automatically redirect any request that is not originally a GET or a HEAD request.

As for this specific scenario, it's worth noting that every single web browser, and a vast majority of HTTP clients treat a 302 after a POST as a redirection to then perform a GET on the 302, even though the original request was a POST.

At the _very least_ by default got should not repeat requests to a redirected location on 3XX if the original request was not a GET or a HEAD, to stay compliant with the http spec by default.

#### Code to reproduce

```js
const got = require('got');
async function main() {
  await got('https://httpbin.org/status/302', {
    method: 'POST',
    body: 'some data',
    retry: 0,
  });
}

main().catch(console.error)
```

#### Checklist

szmarczak:
This allows us to normalize options on-the-fly. This will replace `got.normalizeOptions` / `Request.normalizeOptions`. The following should give a significant performance boost when called

```ts
import {Options} from 'got';
const options = new Options(); // or new Options({url: 'https://localhost'})

options.url = 'https://localhost';
options.url; //=> URL

options.url = 'invalid'; //=> throws

(async () => {
	for (let i = 0; i < 1000; i++) {
		await got(options);
	}
})();
```

Also it will be a lot easier to test the options.

#### TODO

- [x] Make some properties non-enumerable
- [x] Normalize Promise API arguments
- [ ] Implement option merging
- [ ] Implement defaults
- [ ] Fix Promise API
- [ ] Fix Got interface
- [ ] Fix benchmarks
- [x] Invert `option.methodRewriting` mechanism (fixes #1307)
- [ ] Fix `Got options -> HTTP(S) options` `Proxy` handler. Missing `has()` and `ownKeys()`.
- [ ] ...

#### Checklist

- [ ] I have read the documentation.


szmarczak:
Of course it still will be possible to call `got(url, {headers: {...}, ...})`.

szmarczak:
First drawback: there is no clean way to make the properties set by `set ...() / get ...()` enumerable.

<details>
<summary>So logging `normalizedOptions` will give this (click me plz)</summary>

```
{
  _request: undefined,
  _agent: {},
  _decompress: true,
  _timeout: {},
  _prefixUrl: '',
  _body: undefined,
  _form: undefined,
  _json: undefined,
  _cookieJar: undefined,
  _ignoreInvalidCookies: false,
  _searchParameters: undefined,
  _dnsCache: undefined,
  _context: {},
  _hooks: {
    init: [],
    beforeRequest: [],
    beforeError: [],
    beforeRedirect: [],
    beforeRetry: [],
    afterResponse: []
  },
  _followRedirect: true,
  _maxRedirects: 10,
  _cache: undefined,
  _throwHttpErrors: true,
  _username: '',
  _http2: false,
  _allowGetBody: false,
  _lookup: undefined,
  _headers: {},
  _methodRewriting: false,
  _dnsLookupIpVersion: 'auto',
  _parseJson: undefined,
  _stringifyJson: undefined,
  _retry: {
    limit: 2,
    methods: [ 'GET', 'PUT', 'HEAD', 'DELETE', 'OPTIONS', 'TRACE' ],
    statusCodes: [
      408, 413, 429, 500,
      502, 503, 504, 521,
      522, 524
    ],
    errorCodes: [
      'ETIMEDOUT',
      'ECONNRESET',
      'EADDRINUSE',
      'ECONNREFUSED',
      'EPIPE',
      'ENOTFOUND',
      'ENETUNREACH',
      'EAI_AGAIN'
    ],
    maxRetryAfter: undefined,
    calculateDelay: [Function: calculateDelay]
  },
  _localAddress: undefined,
  _socketPath: undefined,
  _method: 'GET',
  _createConnection: undefined,
  _cacheOptions: {},
  _httpsOptions: {},
  _url: URL {
    href: 'http://example.com/',
    origin: 'http://example.com',
    protocol: 'http:',
    username: '',
    password: '',
    host: 'example.com',
    hostname: 'example.com',
    port: '',
    pathname: '/',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  },
  username: undefined
}
```

</details>

szmarczak:
Also it will throw when passing a non-existing property on the Options. Migrate:

```diff
-got({token: 'asdf'});
+got({context: {token: 'asdf'}});
```

szmarczak:
It gives a significant performance boost in terms of normalization:

```
got - normalize options x 95,800 ops/sec ±2.19% (91 runs sampled)
got - new options x 202,692 ops/sec ±0.33% (96 runs sampled)
```

It's a `2.1x` as fast as `got.normalizeOptions`!

szmarczak:
Ok, so `console.log(new Options())` also logged internals, which would be OK as the getters aren't logged by default. So I wanted to make `this[INTERNALS].body`, `this[INTERNALS].json` and `this[INTERNALS].form` not enumerable and the performance dropped back to 95k.

So I did make the entire `INTERNALS` not enumerable and it "kind of" solved the problem because the benchmarks now look like this:

```
got - normalize options x 95,929 ops/sec ±1.81% (94 runs sampled)
got - new options x 182,247 ops/sec ±0.72% (98 runs sampled)
```

`1.9x` is also great. But I wanted all the getters to get logged by default, but encountered [this](https://github.com/nodejs/node/issues/36045).

szmarczak:
Also the Options template now requires to pass an `ErrorType`. This is to extract the types to a different file. In the end Got will export `Options = Options<RequestError>` anyway, so the types remain the same.

szmarczak:
Ok, I've got the streams working, which is pretty nice!

https://github.com/szmarczak/got/blob/8798afc0c5f3ac4bbecde9e49c83cb2cb33722d6/demo.js

sindresorhus:
Is this a breaking change?

szmarczak:
probably yes but the final say should be when I finish this PR. I don't expect any breaking changes than option normalization (but it should still be 99% compatible for the end user). I'm short on time, preparing for uni exams hence I'm not so active since last week :(

szmarczak:
Closing in favor of #1667

PR #1529, opened 2020-11-19
Instantiate CacheableLookup only when needed

fungiboletus:
The `dnsCache` option is disabled by default but a `CacheableLookup` instance is always created. The proposed change is to instantiate CacheableLookup only when needed, which may make `got` a bit faster to load by default and when you do not use the DNS cache.

Moreover, the CacheableLookup class instantiates the DNS [`Resolver` class in NodeJS the standard library](https://nodejs.org/api/dns.html#dns_class_dnspromises_resolver), and the test framework [Jest](https://github.com/facebook/jest/) has a problem with it. It wrongly detects that something keeps running after the tests are finished, while it's just an instance of the DNS Resolver class. See https://github.com/facebook/jest/issues/6423

Jest should be fixed but meanwhile this small improvement in got may also hide the Jest issue for many developers.

#### Checklist

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -31,7 +31,7 @@ import normalizePromiseArguments from '../as-promise/normalize-arguments';
 import {PromiseOnly} from '../as-promise/types';
 import calculateRetryDelay from './calculate-retry-delay';
 
-const globalDnsCache = new CacheableLookup();
+let globalDnsCache;
 
 type HttpRequestFunction = typeof httpRequest;
 type Error = NodeJS.ErrnoException;
@@ -1758,6 +1758,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 		// `options.dnsCache`
 		if (options.dnsCache === true) {
+			if (!globalDnsCache) {
+				globalDnsCache = new CacheableLookup();
+			}
 			options.dnsCache = globalDnsCache;
 		} else if (!is.undefined(options.dnsCache) && !options.dnsCache.lookup) {
 			throw new TypeError(`Parameter \`dnsCache\` must be a CacheableLookup instance or a boolean, got ${is(options.dnsCache)}`);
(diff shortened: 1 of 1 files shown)


fungiboletus:
It looks like the tests succeed excepts with Node 12. But I'm not sure it's related to the changes in this pull request.  The last commit on the master branch was also failing on Node 12 : https://github.com/sindresorhus/got/runs/1342216844

szmarczak:
This will cause the `CacheableLookup` instance to be genereated on every `got` call, which we don't want. `dnsCache: true` needs to point to a global instance, which is the current code. Unfortunately the bug needs to be fixed by Jest themselves.

fungiboletus:
Are you sure ? I think it will be generated only the first time.

szmarczak:
> It's still a global instance, just not initialized before it's used.

043c9501b85172e09819d44ac8eb49c574b27bda

szmarczak:
I'm sure.

fungiboletus:
I understand you don't want to create one instance for each call, and this test is still successful with my suggested modification.

https://travis-ci.com/github/sindresorhus/got/jobs/443599127

image

szmarczak:
Ah, I see now. I didn't notice the `if`.

Giotino (approved):

fungiboletus:
No problems, I understand why you were suspicious. Many thanks for the merge!

szmarczak:
Thanks to you for the PR! Never underestimate your work :)

yinzara:
Could we get a release cut of `got` (I assume 10.8.2) to get this and other issues fixed since the last release?

szmarczak:
No, there is v11 already and v12 is pending.

sindresorhus:
We decided to back port it: https://github.com/sindresorhus/got/releases/tag/v11.8.2

PR #1536, opened 2020-11-20
Move to GitHub Actions and Codecov

sindresorhus:
Travis is no longer free.

Moving to Codecov as it's easier to set up.


.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
+name: CI
+
+on:
+  - push
+  - pull_request
```
sindresorhus, line 5: Do we really need `pull_request` when we have `push`?

szmarczak: I think so.

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#push

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

szmarczak (approved):

PR #1543, opened 2020-11-30
Initialize globalResponse in case of ignored HTTPError

### LINKED ISSUE #1540: JSON shortcut does not work when using throwHttpErrors: false
#### Describe the bug

 - Node.js version: v14.14.0
 - OS & version: Fedora 33

When the server returns an error code with JSON body, and I try to parse it using `.json()` and `{throwHttpErrors: false}`, I get the following error:

```
TypeError: Cannot read property 'request' of undefined
    at /path/to/my/project/node_modules/got/dist/source/as-promise/index.js:157:48
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

#### Actual behavior
TypeError is thrown

#### Expected behavior
Returning the body of response, parsed as JSON (matching the behavior of `{responseType: 'json', resolveOnlyBody: true}`, as documented)

#### Code to reproduce
https://runkit.com/embed/3tt06wxjbio1
```js
const got = require("got");
await got("https://api.github.com/user", {throwHttpErrors: false}).json();
```

#### Checklist

deadNightTiger:
This should fix #1540 

#### Checklist

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -124,12 +124,12 @@ export default function asPromise<T>(normalizedOptions: NormalizedOptions): Canc
 					return;
 				}
 
+				globalResponse = response;
+
 				if (!isResponseOk(response)) {
 					request._beforeError(new HTTPError(response));
 					return;
 				}
-
-				globalResponse = response;
 
 				resolve(request.options.resolveBodyOnly ? response.body as T : response as unknown as T);
 			});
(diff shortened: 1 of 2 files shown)


szmarczak:
Can we move https://github.com/sindresorhus/got/blob/6c9c61224b4115a755e92ced793baab740e42540/source/as-promise/index.ts#L132 before the if above instead?

PR #1561, opened 2020-12-17
Return PCancelable from shortcuts

### LINKED ISSUE #1417: Shortcut methods don't implement PCancelable
#### Describe the bug

 - Node.js version: v14.8
 - OS & version: macOS Catalina v10.15.2

`got.get(...).json()` is not an instance of `CancelableRequest` and `cancel()` method is not implemented:

```ts
export interface CancelableRequest<T extends Response | Response['body'] = Response['body']> extends PCancelable<T>, RequestEvents<CancelableRequest<T>> {
	json: <ReturnType>() => CancelableRequest<ReturnType>;
	buffer: () => CancelableRequest<Buffer>;
	text: () => CancelableRequest<string>;
}
```

#### Actual behavior

`got.get(...).json().cancel()` throws `TypeError { message: 'p.cancel is not a function' }`

#### Expected behavior

Cancels the `got.get(...)` promise without an error.

#### Code to reproduce

```ts
test('recover from cancellation of shortcut using error instance', async t => {
	// Canceled before connection started
	const p = got('http://example.com').buffer();
	const recover = p.catch((error: Error) => {
		if (error instanceof got.CancelError) {
			return;
		}

		throw error;
	});

	p.cancel();

	await t.notThrowsAsync(recover);
});
```

#### Checklist

xamgore:
Fixes #1417. Briefly, `got.get(...).buffer().cancel()` throws `TypeError { message: 'p.cancel is not a function' }`.

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -175,14 +175,19 @@ export default function asPromise<T>(normalizedOptions: NormalizedOptions): Canc
 	};
 
 	const shortcut = <T>(responseType: NormalizedOptions['responseType']): CancelableRequest<T> => {
-		const newPromise = (async () => {
-			// Wait until downloading has ended
-			await promise;
+		const newPromise = new PCancelable(async (resolve, reject, onCancel) => {
+			onCancel(() => promise.cancel());
+			try {
+				// Wait until downloading has ended
+				await promise;
+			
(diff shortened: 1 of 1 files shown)


xamgore:
okay, I'm helpless here

PR #1575, opened 2021-01-05
Require Node.js 12

sindresorhus:
Node.js 10 is soon out of LTS (in April), so I think it makes sense to do it for this major release.

This PR also prepares us for ESM. The ESM move is blocked by: https://github.com/avajs/typescript/issues/5#issuecomment-754678634


sindresorhus:
@szmarczak I tried to remove https://github.com/sindresorhus/got/blob/23695808da78b689b79febb9d5ff3a7061bf578d/source/core/index.ts#L1391-L1392 and it's relevant code, but tests started failing. How do you intend to remove it?

benchmark/index.ts
```diff
@@ -1,12 +1,11 @@
-'use strict';
 import {URL} from 'url';
-import https = require('https');
+import * as https from 'https';
```
szmarczak, line 2: Why this change?

sindresorhus: To prepare for ESM. `= require(…)` is not allowed when compiling to native ESM.

szmarczak:
> it's relevant code, but tests started failing. How do you intend to remove it?

Let me see...

szmarczak:
`pem` crashes on `macOS`

szmarczak:
I'm not really sure why but this line fails on Node.js 12 only: https://github.com/nodejs/node/blob/v12.20.1/lib/_http_client.js#L472 only when running timeout tests. If a timeout gets breached then Node.js throws that assertion error.

sindresorhus:
> pem crashes on macOS

That's an issue in `master` too.

> I'm not really sure why but this line fails on Node.js 12 only: https://github.com/nodejs/node/blob/v12.20.1/lib/_http_client.js#L472 only when running timeout tests. If a timeout gets breached then Node.js throws that assertion error.

This is also an existing issue in `master`. Should definitely be fixed, but not really relevant to this PR.

szmarczak (approved):

PR #1640, opened 2021-02-25
Update to `http2-wrapper@2.0.0`

szmarczak:
I've released `http2-wrapper@1.0.0` and I believe it's quite stable. Unfortunately older versions of Node.js are buggy, so it requires Node.js `15.10.0` or above. We can't put this package in the deps, so it needs to be installed manually.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
 		"cacheable-lookup": "^6.0.0",
 		"cacheable-request": "^7.0.1",
 		"decompress-response": "^6.0.0",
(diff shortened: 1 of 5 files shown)


szmarczak:
I had to re-release the beta version as `1.0.3` because I genuinely didn't expect that `1.0.0` would fall under `^1.0.0-beta`. So the breaking version has been released as `2.0.0`.

sindresorhus:
🙌🏻

PR #1644, opened 2021-02-26
Improve the pagination API

### LINKED ISSUE #1215: Changes to the `.pagination` API

Improve `.pagination` API (and possibly performance and stability)

#### Describe the feature

I would humbly suggest the following 2 breaking changes:

- Switch order of `allItems` and `currentItems` parameters of `pagination.paginate`, `pagination.filter` and `pagination.shouldContinue`

My reasoning about parameters order is that probably most use cases don't need the `allItems` parameters, with `currentItems` typically being more used. To be confirmed by other users, but if indeed `currentItems` is used more than `allItems` then it should reflect in the order of parameters.

It doesn't change how the Got code works but could save custom `pagination.paginate`, `pagination.filter` and `pagination.shouldContinue` from having an unused and unaesthetic `_` as the second parameter... ;)

As per the default of `allItems`, it implies a waste of CPU and RAM resources when unused. This is especially noticeable when iterating large datasets (and can results in a node.js crash in worst case scenarios)

#### Checklist

szmarczak:
Fixes #1215
Closes #1646

Let's make the `stackAllItems` option default to `true` (leave it as it is).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -980,18 +980,18 @@ const got = require('got');
 ###### pagination.filter
 
 Type: `Function`\
-Default: `(item, allItems, currentItems) => true`
+Default: `({item, allItems, currentItems}) => true`
 
 Checks whether the item should be emitted or not.
 
 ###### pagination.shouldContinue
 
 Type: `Function`\
-Default: `(item, allItems, currentItems) => true`
+Default: `({item, allItems, currentItems}) => true`
 
 Checks whether the pagination should continue.
 
-For example, if you need to stop **before** emitting an entry with some flag, you should use `(item, allItems, currentItems) => !item.flag`. If you want to stop **after** emitting the entry, you should use `(item, allItems, currentItems) => allItems.some(entry => entry.flag)` instead.
+For example, if you need to stop **before** emitting an entry with some flag, you should use `({item}) => !item.flag`. If you want to stop **after** emitting the entry, you should use `({item, allItems}) => allItems.some(item => item.flag)` instead.
 
 ###### pagination.countLimit
 
--- a/readme.md
+++ b/readme.md
@@ -934,7 +934,7 @@ A function that transform [`Response`](#response) into an array of items. This i
 Type: `Function`\
 Default: [`Link` header logic](source/index.ts)
 
-The function takes three arguments:
+The function takes an object with the following properties:
 - `response` - The current response object.
 - `allItems` - An array of the emitted items.
 - `currentItems` - Items from the current response.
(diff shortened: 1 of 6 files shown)


szmarczak:
@sindresorhus I need your opinion on:

> Switch order of allItems and currentItems parameters of pagination.paginate, pagination.filter and pagination.shouldContinue

or maybe we should pass an object too? I'm good with anything tbh. It just feel a bit weird if it's an object though, because the `array.filter()` JS API takes an entry object as the first argument.

source/as-promise/types.ts
```diff
@@ -11,6 +11,12 @@ All parsing methods supported by Got.
 */
 export type ResponseType = 'json' | 'buffer' | 'text';
 
+export interface PaginateData<R, T> {
```
sindresorhus, line 14:
```suggestion
export interface PaginateData<ResponseType, ElementType> {
```

To make it a little bit more readable?

sindresorhus:
> or maybe we should pass an object too? I'm good with anything tbh. It just feel a bit weird if it's an object though, because the array.filter() JS API takes an entry object as the first argument.

Yeah, I think that would be nicer as you might not need all the arguments there either, and it would be more consistent.

sindresorhus (changes requested):

readme.md
```diff
@@ -934,7 +934,7 @@ A function that transform [`Response`](#response) into an array of items. This i
 Type: `Function`\
 Default: [`Link` header logic](source/index.ts)
 
-The function takes three arguments:
+The function takes an object with the following properties:
 - `response` - The current response object.
 - `allItems` - An array of the emitted items.
```
PopGoesTheWza, line 939: Maybe add a reference to `pagination.stackAllItems` behaviour?

PopGoesTheWza (commented):
Typescript documentation needs to be updated in a few places IMHO:

https://github.com/sindresorhus/got/blob/a1b259cb747c449f6f6daf94b5ed24490fccf010/source/as-promise/types.ts#L41
https://github.com/sindresorhus/got/blob/a1b259cb747c449f6f6daf94b5ed24490fccf010/source/as-promise/types.ts#L63

readme.md
szmarczak: It's just a scroll away: https://github.com/sindresorhus/got#paginationstackallitems

PopGoesTheWza:
In PR #1646 I generalized parameters as object to `pagination.filter` and `pagination.shouldContinue` as well. It makes the API more consistent IMHO.

PR #1646, opened 2021-02-26
feat: `pagination` parameters as object

### LINKED ISSUE #1215: Changes to the `.pagination` API

PopGoesTheWza:
BREAKING CHANGE

Fixes #1215 

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -934,7 +934,7 @@ A function that transform [`Response`](#response) into an array of items. This i
 Type: `Function`\
 Default: [`Link` header logic](source/index.ts)
 
-The function takes three arguments:
+The function takes an object with three properties as single argument:
 - `response` - The current response object.
 - `allItems` - An array of the emitted items.
 - `currentItems` - Items from the current response.
@@ -955,7 +955,7 @@ const got = require('got');
 			offset: 0
 		},
 		pagination: {
-			paginate: (response, allItems, currentItems) => {
+			paginate: ({response, allItems, currentItems}) => {
 				const previousSearchParams = response.request.options.searchParams;
 				const previousOffset = previousSearchParams.get('offset');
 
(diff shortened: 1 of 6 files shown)


PopGoesTheWza:
@szmarczak should I close this PR?

szmarczak:
Well I don't mind if yours gets merged instead :) @sindresorhus

sindresorhus:
I went with #1644 as it was technically first (that's usually how I resolve duplicate PRs that are the same quality).

Sorry for the wasted work @PopGoesTheWza. It was unfortunate that the timing was so so close.

If you see any improvements to https://github.com/sindresorhus/got/pull/1644 or the pagination code in general, we're more than happy to other PRs from you 🙌🏻

PopGoesTheWza:
> Sorry for the wasted work @PopGoesTheWza. It was unfortunate that the timing was so so close.

No worries. It offered me the possibility to criticize @szmarczak work! ;)

sindresorhus:
> It offered me the possibility to criticize @szmarczak work! ;)

PR reviews are the best kind of contribution, by the way. 👍🏻

PR #1647, opened 2021-02-26
Always declare `retryCount`and `error`  in `beforeRetry` hooks

### LINKED ISSUE #1420: The type on BeforeRetryHook shouldn't contain any optional parameters
#### Describe the bug

In the typescript definition of `BeforeRetryHook`, `error` and `retryCount` are declared as optional, meaning they're potentially `undefined`. However, it's not true.

https://github.com/sindresorhus/got/blob/0d43f5fea0460977089236b58f2354307fbf27fc/source/core/index.ts#L107

#### Actual behavior

`error` and `retryCount` contains the `undefined` type

#### Expected behavior

`error` and `retryCount` should be just simply `RequestError` and `number`

szmarczak:
~~Partially~~ Fully fixes #1420

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -832,7 +832,7 @@ const got = require('got');
 got.post('https://example.com', {
 	hooks: {
 		beforeRetry: [
-			(options, error, retryCount) => {
+			(options, retryCount, error) => {
 				if (error.response.statusCode === 413) { // Payload too large
 					options.body = getNewBody();
 				}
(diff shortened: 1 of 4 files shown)


szmarczak:
@sindresorhus What do you think of replacing `(options, error, retryCount)` with `(error, request)`? `options` can be accessed via `request.options` and `retryCount` via `request.retryCount`.

sindresorhus:
> What do you think of replacing (options, error, retryCount) with (error, request)? options can be accessed via request.options and retryCount via request.retryCount.

Sounds good as long as we clearly document where the user can get the `options` and `retryCount`. Since it's a breaking change, I would take the opportunity to make it an object `({error, request})` which would make it easier to add more properties in the future.

szmarczak:
Closing in favor of #1667, which does a much better job as it introduces a new error `RetryError` which always triggers a new retry when thrown.

PR #1648, opened 2021-02-26
Add `ownContext` to `RequestError`

### LINKED ISSUE #1353: Custom error in BeforeError hook
#### Describe the bug

I was creating custom errors in `beforeError` hook. After upgrading to newer got version, it's not allowed by types (but still working if to cast types).
#### Actual behavior

I have to cast type.

#### Expected behavior

Return type for `BeforeErrorHook` is any `Error`.

#### Code to reproduce

Example of casting:
```js
hooks: { beforeError: [error => new MyCustomError(error) as any]; }
```

#### Checklist

szmarczak:
Fixes #1353

#### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -897,6 +897,8 @@ Default: `[]`
 
 Called with an `Error` instance. The error is passed to the hook right before it's thrown. This is especially useful when you want to have more detailed errors.
 
+**Note:** `RequestError` instances have an `ownContext` property. By default, it is `undefined`. You can freely modify it to gather any data you need to catch.
+
 **Note:** Errors thrown while normalizing input options are thrown directly and not part of this hook.
 
 ```js
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -1186,6 +1186,7 @@ export class RequestError extends Error {
 	readonly response?: Response;
 	readonly request?: Request;
 	readonly timings?: Timings;
+	ownContext?: Record<string, unknown>;
 
 	constructor(message: string, error: Partial<Error & {code?: string}>, self: Request | NormalizedOptions) {
 		super(message);
--- a/test/hooks.ts
+++ b/test/hooks.ts
@@ -265,6 +265,27 @@ test('catches beforeError errors', async t => {
 	}), {message: errorString});
 });
 
+test('errors have their own context', async t => {
+	const error = await t.throwsAsync<RequestError>(got('https://example.com', {
+		request: () => {
+			throw new Error('No way');
+		},
+		hooks: {
+			beforeError: [
+				error => {
+					error.ownContext = {
+						foo: 'bar'
+					};
+
+					return error;
+				}
+			]
+		}
+	}), {message: errorString});
+
+	t.is(error.ownContext?.foo, 'bar');
+});
+
 test('init is called with options', withServer, async (t, server, got) => {
 	server.get('/', echoHeaders);
 


sindresorhus:
Doesn't look like this solves the use-case mentioned in https://github.com/sindresorhus/got/issues/1353#issuecomment-709592464 Do we want that?

readme.md
```diff
@@ -897,6 +897,8 @@ Default: `[]`
 
 Called with an `Error` instance. The error is passed to the hook right before it's thrown. This is especially useful when you want to have more detailed errors.
 
+**Note:** `RequestError` instances have an `ownContext` property. By default, it is `undefined`. You can freely modify it to gather any data you need to catch.
```
sindresorhus, line 900: Might be nice to make it an empty object by default `Object.create(null)`, so people can just attach properties if they want. Also makes the type nicer.

Should we make the property enumerable?

szmarczak:
> Doesn't look like this solves the use-case mentioned in #1353 (comment)

It does. Users will need to set `error.ownContext` and return that error. Retuning an object is too magical IMO.

readme.md
szmarczak: > Should we make the property enumerable?

It is enumerable already.

sindresorhus: Yeah, I'm asking whether or not it should be enumerable?

sindresorhus:
@coleabbeduto-NM Does this solve your use-case?

readme.md
szmarczak: I would leave it enumerable by default. If people don't want this then they can use `Object.defineProperty`.

sindresorhus: Ok, I agree. But we should at least document that it's enumerable by default.

szmarczak: Followed by an example on how to make it not enumerable :)

aalexgabi:
@callumlocke @sindresorhus It looks like this implementation requires users to add the properties to errors themselves. I don't think this is a good behavior out of the box. Errors thrown by got should have at least the url and the method in the error object as enumerable properties.

aalexgabi:
@szmarczak What about something like this:
```
		Object.defineProperty(this, 'options', {
			// This fails because of TS 3.7.2 useDefineForClassFields
			// Ref: https://github.com/microsoft/TypeScript/issues/34972
			enumerable: true,
			value: _.omit(options, ['options.headers.Authorization'])
		});
```
In https://github.com/sindresorhus/got/blob/1d012a513a2263f0f51d327e7b509d9d3bd8ffc5/source/errors.ts#L20

szmarczak:
> It looks like this implementation requires users to add the properties to errors themselves.

This would do the same as https://github.com/sindresorhus/got/issues/1353#issuecomment-709592464

> This requires making sure that no developer forgets to add this to every request.

The `error.options` property is enumerable, but the keys inside are not due to a JS limitation.
You need to use `error.options.toJSON()` or `util.inspect`.

This works great when using just the default `console.log(error.options)`. I haven't checked external loggers yet.

> I don't think this is a good behavior out of the box.

On the other side, leaking secrets is not good either.

> What about something like this:

We don't use `lodash` here. I don't think it's necessary.

szmarczak:
@sindresorhus Would you prefer `ownContext` or `errorContext`? I think `errorContext` would be more descriptive.

sindresorhus:
Let's go with `.errorContext`.

szmarczak:
Maybe let's postpone this until after Got 12 gets released? We can release this as a minor patch. This should not be breaking.

I'm still open to discussions. Meanwhile let's focus on getting Got 12 ready. I just need to do the docs, migration guide and we're good to go.

sindresorhus:
Agreed

Vladisvob (approved):
#2154

sindresorhus:
Closing for lack of activity.

PR #1662, opened 2021-03-10
Minor tweak

Richienb:
`yoctodelay` can be used instead of `delay` to retain space while being modular.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
 		"http2-wrapper": "^2.0.0",
 		"lowercase-keys": "^2.0.0",
 		"p-cancelable": "^2.0.0",
-		"responselike": "^2.0.0"
+		"responselike": "^2.0.0",
+		"yoctodelay": "^1.2.0"
 	},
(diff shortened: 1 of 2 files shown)


szmarczak:
> yoctodelay can be used instead of delay to retain space while being modular.

But we haven't used `delay`, so you've just added an additional dependency. `yoctodelay` weighs [4.5KB](https://packagephobia.com/result?p=yoctodelay) while the small piece of code is just 90 bytes and is used at only one place.

Richienb:
The size of the [actual code that runs](https://github.com/sindresorhus/yoctodelay/blob/main/index.js) in 49 bytes, which when bundled is actually smaller.

But that's beside the point: it's the modularity that counts, even if it's just a little bit.

Node.js 16 will support `import {setTimeout: delay} = 'timers/promises'` so we will be able to do this with no dependency and with no extra code then.

szmarczak:
I agree :)

Issue #1654, opened 2021-03-05
If agent fails, try again without


joergd:
If I start up a stream request with an agent (proxy), and that request fails (if say the proxy is unavailable), is there an easy way to fallback to making the call without the agent?

szmarczak:
It should be possible to remove the `agent` option in the `beforeRetry` hooks.

szmarczak:
Let's keep this issue. This needs a test.

joergd:
I can't use that hook, as I am streaming ...

szmarczak:
In that case - yes, it's possible: https://github.com/sindresorhus/got#retry

joergd:
Yeah I looked at that - but I couldn't get that to work. Just to confirm with that sample code, the writeStream is the stream that represents the actual response that I can stream further along in my pipeline?

szmarczak:
This is still possible with Got v11.8.2:

```js
const https = require('https');
const got = require('got');

class MyAgent extends https.Agent {
    createConnection(port, options, callback) {
        console.log(`Connecting with MyAgent`);
        return https.Agent.prototype.createConnection.call(this, port, options, callback);
    }
}

const fn = retryCount => {
    const stream = got.stream('https://httpbin.org/status/408', {
        agent: {
            https: retryCount === 0 ? new MyAgent() : undefined
        }
    });
    
    stream.retryCount = retryCount;

    stream.on('retry', (retryCount, error) => {
        console.log('Creating new retry:', retryCount);
        fn(retryCount);
    }).on('error', error => {
        console.log('Retry count:', error.request.retryCount);
    }).resume().on('end', () => {
        console.log('`end` event');
    });
};

fn(0);
```

Issue #1610, opened 2021-02-16
default searchParams not properly merged when requesting searchParams is undefined


asomethings:
#### Describe the bug

 - Node.js version: 14.15.3
 - OS & version: Ubuntu 20.04.1 LTS

#### Actual behavior

Does not merge `searchParams` when `searchParams: undefined` making a request.
But, sends default `searchParams` when 
- searchParams does not exist in request options
- searchParams is an empty object in request options
- searchParams is an empty array in request options

```typescript
import got from 'got'

const gotExtended = got.extend({ searchParams: { query: true }, responseType: 'json' })
gotExtended('http://httpbin.org/get', { searchParams: undefined })
  .then((response) => response.body)
  .then(console.log)

/** Above returns
{
  args: {},
  headers: {
    Accept: 'application/json',
    'Accept-Encoding': 'gzip, deflate, br',
    Host: 'httpbin.org',
    'User-Agent': 'got (https://github.com/sindresorhus/got)',
    'X-Amzn-Trace-Id': 'Root=1-602bd8d5-5623f1de6a109692250b3177'
  },
  origin: 'xxx.xxx.xx.xxx',
  url: 'http://httpbin.org/get'
}
*/
```

#### Expected behavior

Should pass default `searchParams` even though requested `searchParams` is undefined

```typescript
import got from 'got'

const gotExtended = got.extend({ searchParams: { query: true }, responseType: 'json' })
gotExtended('http://httpbin.org/get', { searchParams: undefined })
  .then((response) => response.body)
  .then(console.log)

/** Above should return
{
  args: { query: 'true' },
  headers: {
    Accept: 'application/json',
    'Accept-Encoding': 'gzip, deflate, br',
    Host: 'httpbin.org',
    'User-Agent': 'got (https://github.com/sindresorhus/got)',
    'X-Amzn-Trace-Id': 'Root=1-602bd8d5-5623f1de6a109692250b3177'
  },
  origin: 'xxx.xxx.xx.xxx',
  url: 'http://httpbin.org/get?query=true'
}
*/
```

#### Code to reproduce

```typescript
import got from 'got'

const gotExtended = got.extend({ searchParams: { query: true }, responseType: 'json' })
gotExtended('http://httpbin.org/get', { searchParams: undefined })
  .then((response) => response.body)
  .then(console.log)
```

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

[...]

Issue #1599, opened 2021-02-08
Upper limit on the retry backoff


Sytten:
#### What problem are you trying to solve?
Currently there is no upper limit on the backoff, meaning it grows exponentially if you want many retries which is not something you might want.

#### Describe the feature
So I would add a new optional parameter to the retry object called `backoffLimit` that would be a number of milliseconds to wait maximum between retries. This would basically do a `Math.min` between the default calculated backoff value and this parameter value. This would allow a user to put a limit of say 10 and have the following behaviour:

| Retry | 1  | 2  | 3  | 4   | 5   | 6   | 7   | 8   | 9   | 10  |
|-------|----|----|----|-----|-----|-----|-----|-----|-----|-----|
| Time  | 2s | 4s | 8s | 16s | 32s | 60s | 60s | 60s | 60s | 60s |

#### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.

sindresorhus:
Sounds ok to me, but I'll let other maintainer chime in too.

For now, you could provide your own `calculateDelay` function to achieve this.

Sytten:
Thanks! I know I can provide the method, but I didn't want to lose the checks it does for the header and there was no easy way to import the default calculate delay method.

Issue #1591, opened 2021-01-28
When transfer encoding is chunked, GOT sometimes does not download full content


adityapatadia:
#### Describe the bug

 - Node.js version: 14.15.1. GOT latest.
 - OS & version: Ubuntu 20.04

#### Actual behavior

When downloading content which does not have content-length header, i.e. chunked transfer, GOT sometimes does not download full content when `responseType: 'buffer'` is set. One example URL is here: https://cdn.sentinelassam.com/h-upload/2021/01/28/193125-whatsapp-image-2021-01-28-at-112922-am.webp

#### Expected behavior

It should download full content.

#### Code to reproduce

Not able to re-produce reliably as it happens only sometimes. I am not sure under which conditions but it certainly happens.

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

sindresorhus:
There unfortunately no actionable info here. Would you be able to try on the latest Node.js 15 release?

adityapatadia:
We could solve it by putting this code

```javascript
let lengthMatches = true;
if (response.headers && response.headers['content-length']) {
    lengthMatches = parseInt(response.headers['content-length']) == response.body.byteLength;
  }

if(response.body.byteLength != 0 && lengthMatches && response.complete){
   //everything looks okay....
  }
```

May be this can give you some idea? We will not be able to try on Node 15 because this issue is not reproducible on local machine and can only be produced when there is very high load. We will still give you info if we find something.

szmarczak:
Closing in favor of #1462

PR #1667, opened 2021-03-14
Got 12 improvements

### LINKED ISSUE #1538: Enhance pagination with parsed Link header

Paginating some GitHub API endpoints with extra logic (logging etc) when reaching internal limit of fetched pages.

#### Describe the feature

Would be nice to have more control on pagination logic. For example, [parse-link-header](https://github.com/thlorenz/parse-link-header) provides well structured format of parsed Link header.

**Raw Link header:**
```
<https://<api_url>?page=3&per_page=100>; rel="next",
<https://<api_url>?page=1&per_page=100>; rel="prev"; pet="cat",
<https://<api_url>?page=5&per_page=100>; rel="last"
```
**Parsed result:**
```javascript
{
  next: {
    page: '3',
    per_page: '100',
    rel: 'next',
    url: '<api_url>?page=3&per_page=100',
  },
  prev: {
    page: '1',
    per_page: '100',
    rel: 'prev',
    pet: 'cat',
    url: '<api_url>?page=1&per_page=100',
  },
  last: {
    page: '5',
    per_page: '100',
    rel: 'last',
    url: '<api_url>?page=5&per_page=100',
  },
}
```

If this data will be accessible, for example from [pagination.shouldContinue](https://github.com/sindresorhus/got#paginationshouldcontinue), it will help to add custom logic based on next or previous pages.

**Example**

```javascript
const got = require('got');
const limit = 10;

got.paginate('https://example.com/items', {
  searchParams: {
    limit,
  },
  pagination: {
    shouldContinue: (item, allItems, currentItems, pagination) => {
      if (pagination.next.page > limit) {
        log.warn('Reached page limit');
        return false;
      }
      return true;
    },
  },
});
```

#### Checklist

- [ ] I have read the documentation and made sure this feature doesn't already exist.

### LINKED ISSUE #1663: default maxRetryAfter is not calculated correctly
#### Describe the bug

 - Node.js version: v12.20.0
 - OS & version: macOS Mojave | 10.14.6

In got documentation sais

> If `maxRetryAfter` is set to `undefined`, it will use `options.timeout`.

But is not working well in case of use extends

#### Actual behavior

Event if we use a timeout config like `timeout: { request: 5000 }`if don't provide a `maxRetryAfter` value when  config is "normalized" the maxRetryAfter is being calculated as `Infinity`

#### Expected behavior

As per documentation `maxRetryAfter` should get `5000` as value.

#### Code to reproduce

```js
const { describe, it } = require('mocha')
const { strictEqual } = require('assert').strict
const Got = require('got')

describe.only('default maxRetryAfter', function () {
  it('should get timeout value', async function () {
    const newGot = Got.extend({ timeout: { request: 5000 } })
    strictEqual(newGot.defaults.options.retry.maxRetryAfter, 5000)
  })
})

/**
      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

Infinity !== 5000

/*
```

I think the problem is because the code to calculate default value for `maxRetryAfter` is in the method  `normalizeArguments()` and is calculating the default value before the options are merged, for this reason get timeout as undefined and the `maxRetryAfter` is getting infinity value

### LINKED ISSUE #1668: Merging `port` into defaults doesn't work with `prefixUrl`
#### Describe the bug

With Node 12, got@11.8.2.

Merging `port` into default options doesn't work when `prefixUrl` is set.

#### Code to reproduce

See https://github.com/jmm/repro-got-merge-port-to-defaults.

```js
const gotInstance = got.extend({
  mutableDefaults: true,
  prefixUrl: "http://example.com",
})

const port = 1234

gotInstance.defaults.options = gotInstance.mergeOptions(gotInstance.defaults.options, {
  port,
})

assert.equal(gotInstance.defaults.options.port, port)
```

#### Actual behavior

I expect port to be set as specified and the assertion to be true.

#### Expected behavior

Port is undefined.

---

I suspect this code: https://github.com/sindresorhus/got/blob/v11.8.2/source/core/index.ts#L1639-L1645

szmarczak:
This is a redo of #1521

Fixes #1668 (tested)
Fixes #1663 (tested)
Fixes #1538 (tested)

**If this gets merged, we need to open an issue to fix the docs. I'll do them, as well as migration notes etc.**

<strike>
**Note**: By using the `Options` class we introduce a regression! In some cases where option normalization throws `error.options` is `undefined`. ~~This could be solved via **not throwing in the `Options` constructor**, but setting a property e.g. `lastKnownError` instead. That way `error.options` would be defined. It's just a matter what we do consider more friendly.~~ Ok found a better way - let's define `error.options` on the error thrown by the `Options` constructor. Later!
</strike>
<br>
<br>

- [x] Improved option normalization (new class: `Options`)
  - [x] `new Options(url | options, options?, defaults?)`
    **Note**: Only the first `url` instance is used.
  Meaning: `got('https://a', {url: 'https://b'})` will actually make a request to `a`.
  If you expect to replace the `url` option, then you need pass the initial `url` like this:
  `got({url: 'https://a', ...otherOptions})`
  - [x] `options.merge(newOptions)` replaced `got.mergeOptions` & `Request.normalizeArguments`
    **Note**: It does NOT merge the `url` option - but it **does** merge the `searchParameters` option
  - [x] Throws when an option does not exist
  - [x] Dropped legacy URL support, use `options.url` instead (a URL instance)
  - [x] Creates a brand new object for the native `http` options
  - [x] No implicit `timeout` declaration, use `{timeout: {request: 1000}}` instead
  - [x] No implicit `retry` declaration, use `{retry: {limit: 3}}` instead
  - [x] `dnsLookupIpVersion` is now a number (`4` or `6`) or undefined
- [x] Switched to `get-stream` (getting a buffer from a stream)
- [x] Refactored `core/index.ts`
- [x] Refactored `as-promise/index.ts`
  - [x] Moved the `onCancel` handlers out of `makeRequest` in order to prevent them being called multiple times (`.destroy()` needs to be called only once)
  - [x] The `Request` class now provides `rawBody` if `request._noPipe` is `true` (used internally by Promise API)
  - [x] Switched from `for/entries` to classic `for` loop when iterating over `afterResponse` hooks in order to improve performance
        **Why**: `.entries()` sends a query to get enumerable properties + generates an array - but we don't need that since we already know the enumerable properties since it's an array and we don't need to generate a new one
  - [x] [Fixed `Request` options not being reused on retry](https://github.com/sindresorhus/got/blob/c3ca847318327152aba6e0eb3bde4b34d7cc3d5e/source/as-promise/index.ts#L137-L139) (~~needs a test~~ tested)
- [x] Moved `Response`-related stuff to `core/response.ts`
- [x] Moved error definitions to `core/errors.ts` - you now have to import them directly, meaning you cannot access them via `got.RequestError`
- [x] New error type: `RetryError` which always triggers a new retry when thrown
- [x] Moved `waitForOpenFile` to `core/utils/wait-for-open-file.ts`
- [x] Added missing `once` types for Stream API
- [x] Switched from symbols to `private`s
- [x] `redirectUrls` and `requestUrl` now give URL instances
- [x] The retry timeout in `request._beforeError` is now promisified
- [x] Renamed `request.aborted` to `request.isAborted`
- [x] Renamed the `lookup` option to `dnsLookup`
- [x] The `beforeRetry` hook now accepts only one argument: `error`
      To retrieve:
        - the underlying request: `error.request`
        - the underlying options: `error.options`
        - current attempt number: `error.request.retryCount + 1`
- [x] The `beforeRedirect` hook's first argument (`options`) is now a cloned instance of the Request options.
     This was done to make retrieving the original options possible. To retrieve the original options, use `plainResponse.request.options`.
- [x] The `redirect` event now takes two arguments: `updatedOptions` and `plainResponse`
- [x] ~~The `retry` event now takes only one argument: `error`~~ (reverted; reason: the current API is much nicer)
      ~~To get the attempt count, use `error.request.retryCount + 1`~~
- [x] The `socketPath` option has been removed. Use the `unix:` protocol instead.
- [x] The `retryWithMergedOptions` function no longer returns a `Promise`. Keep in mind that this should be the last function being executed in an `afterResponse` hook.
- [x] `error.options` is now enumerable
- [x] `defaults.handlers` don't need a default handler (`(options, next) => next(options)`) now.
- [x] Renamed `options.https` to `options.httpsOptions` 
- [x] Removed internal `_isAboutToError` function
- [x] Renamed the `searchParams` option to `searchParameters`
- [x] You can no longer set `options.agent` to `false`. To do so, you need to define all the `options.agent` properties: `http`, `https` and `http2`.
- [x] Refactored Got interface (instances, defaults, etc.)
- [x] The `init` hook now accepts a second argument: `self`, which points to an `Options` instance
- [x] The `init` hooks are ran only when passing an options object explicitly. It's no longer called on `got('https://example.com')` - in order to trigger `init` hooks, you need to call `got('https://example.com', {})`
- [x] When passing a `url` option when paginating, it now needs to be an absolute URL - the `prefixUrl` option is always reset from now on. The same when retrying in an `afterResponse` hook.
- [x] Fixed tests
  - [x] Fix `test/arguments.ts`
  - [x] Fix `test/create.ts`
  - [x] Fix `test/normalize-arguments.ts`
  - [x] Fix `test/hooks.ts`
- [x] `got.extend(...)` will throw when passing some options that don't accept `undefined` - `undefined` no longer retains the old value, as setting `undefined` explicitly may reset the option
- [x] Look at the code and update what's changed
- [ ] Created a new documentation

#### To be fixed

- [ ] #1654
- [ ] #1591 & #1462
- [ ] #1610 (might be already fixed - needs a test)
- [ ] #1137
- [ ] #1599
- [ ] #1574 - needs a test
- [ ] #1569
- [ ] #1553
- [ ] #1567 (might be already fixed - needs a test)
- [ ] #1527
- [ ] #1450 - needs a test
- [ ] #1503
- [ ] #1507
- [ ] #1492
- [ ] #1408
- [ ] #1117 (maybe, may be postponed to Got 12.x / 13)
- [ ] #1405 (docs)

#### Checklist

- [ ] I have read the documentation.


szmarczak:
So I've benchmarked option normalization... and it's great <3 130k/s when normalizing URL. But when just cloning the options... 1M/s. Blazing fast.

szmarczak:
`asPromise` scores ~3600 req/s. That's 1200 req/s slower than `core` (Stream API). I don't think we can do faster.

@Nytelife26 If you have any idea on how to improve `asPromise/index.ts`, feel free to chime in. Note we don't want to lose any of the available features Got has to offer.

Nytelife26:
> If you have any idea on how to improve asPromise/index.ts

I apologise, I have only just seen this. Do you still require assistance? If so, I will be available tonight @szmarczak

szmarczak:
> Do you still require assistance?

No, I don't require, I'm just kindly asking if you have spare time to have a quick peek ;) If that's not the case, no problem, we'll manage. I ran [a flamegraph](https://github.com/davidmarkclements/0x) and the bottleneck here comes from the native `http` module (well, streams too). I highly doubt we can improve the performance of `asPromise`.

Nytelife26:
> to have a quick peek

Yes, that's absolutely fine. I'll take a look in an hour or so when I get home.

> I highly doubt we can improve the performance of `asPromise`

I'm sure we can. Who knows until we try? We got this :)

erfanium:
May this PR fix this https://github.com/sindresorhus/got/pull/1549#issuecomment-740883268? It bothers me.

szmarczak:
@erfanium https://github.com/sindresorhus/got/pull/1549#issuecomment-740891030

szmarczak:
IOW yes

szmarczak:
Todo: do not override properties that already exist when assigning promise properties to async handler result.

szmarczak:
All tests are passing. >90% coverage.

szmarczak:
Meh, failing on Node.js 12.

szmarczak:
`cache` is failing because of #1385. Prepare for a stupid workaround.

szmarczak:
Oh, and one major thing: this PR uses TS default export.

sindresorhus:
> Fixes #1668 (needs a test)
Fixes #1663 (needs a test)
Fixes #1538 (needs a test)

Did tests get added for these?

szmarczak:
Not yet, I can get them added in an hour. Currently I'm finishing my PR in `ava/typescript`.

szmarczak:
Added the tests :)

sindresorhus:
> Created a new documentation

Can you open an issue for this?

sindresorhus:
This is looking really good! Super nice work as always. 🙌

sindresorhus:
> Renamed the searchParams option to searchParameters

I don't think we should do this though. Can you revert that? While I generally prefer full words, I think it's more important here to use the same name as `Fetch`, and also, it's not worth doing a breaking change just for that.

szmarczak:
Sure, will do this in a few hours.

szmarczak:
7b3cbfdf4d7e5264c3acba7576b4e25f082c8f97

PR #1687, opened 2021-04-15
Move to ESM

sindresorhus:


package.json
```diff
@@ -104,15 +105,16 @@
 			"test/*"
 		],
 		"timeout": "1m",
-		"typescript": {
-			"compile": "tsc",
-			"rewritePaths": {
-				"test/": "dist/test/"
```
szmarczak, line 110: Is this not needed? `dist` is still referenced in `exports`.

sindresorhus: No, that config is just for tests. We now use `ts-node` instead which does this for us.

sindresorhus:
I'm merging this to unblock other work, but let me know if there's anything else.

test/helpers/with-server.ts
```diff
@@ -1,19 +1,19 @@
 import {promisify} from 'util';
-import * as test from 'ava';
+import {ExecutionContext, Macro} from 'ava'; // eslint-disable-line ava/use-test
```
szmarczak, line 2: I also noticed that this is quite buggy. Does it also happen when doing `import type`?

sindresorhus: We just need to fix the eslint-plugin-ava rule.

Issue #1711, opened 2021-05-03
Set error codes on all errors


p-bakker:
#### What would you like to discuss?
On HTTPErrors the statusCode of the response is not exposed directly. You have to go through the HTTPError.response to get the statusCode

I would've expected the statusCode to be directly exposed on the HTTPError or otherwise set as the value of the `code` property, but neither is true

Is that as expected or is this a bug?

#### Checklist

- [x] I have read the documentation.

szmarczak:
In regards to the documentation this is expected.

p-bakker:
Tnx for the quick response.

I wasn't able to deduct this from the documentation, so thanks for clarifying

p-bakker:
Additionally: shouldn't the code property on HTTPError instances be set to 'HTTPError'? According to the docs: `Contains a code property with error class code, like ECONNREFUSED`

Drewfergusson:
Per @sindresorhus comments in #1739 it would be good to take a comprehensive approach to the code property for all error objects. The plain RequestError seems to always have a code associated with it that correlates to a Linux kind of error 

`[
			'ETIMEDOUT',
			'ECONNRESET',
			'EADDRINUSE',
			'ECONNREFUSED',
			'EPIPE',
			'ENOTFOUND',
			'ENETUNREACH',
			'EAI_AGAIN'
]` 

The other child error messages all seem to return with the code property as undefined. As also stated it would be good to follow a Node.JS convention of ERR_*. 

I might suggest

- CacheError `ERR_CACHE_ACCESS`
- ReadError `ERR_READING_RESPONSE_STREAM`
- ParseError `ERR_BODY_PARSE_FAILURE`
- UploadError `ERR_READING_REQUEST_STREAM`
- HTTPError `ERR_NON_2XX_3XX_RESPONSE`
- MaxRedirectsError `ERR_MAX_REDIRECTS`
- UnsupportedProtocolError `ERR_UNSUPPORTED_PROTOCOL`
- TimeoutError `ERR_TIMEOUT`
- CancelError `ERR_CANCEL_INVOKED`

sindresorhus:
> CancelError ERR_CANCEL_INVOKED

`ERR_CANCELED`

sindresorhus:
Otherwise, looks good to me.

@szmarczak ?

szmarczak:
`ERR_READING_REQUEST_STREAM` -> `ERR_UPLOAD`

PR #1739, opened 2021-06-02
HTTPError code set to 'HTTPError' #1711

Drewfergusson:
This is my understanding of the intention of #1711. This sets the code property on the HTTPError object to 'HTTPError' instead of leaving it as undefined. 

#### Checklist

- [X] I have included some tests (updated failing test to pass)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1705,7 +1705,7 @@ Additionaly, the errors may have `request` (Got Stream) and `response` (Got Resp
 
 #### got.RequestError
 
-When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied `code` defaults to `ERR_GEN_REQ_ERROR` All the errors below inherit this one.
+When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied, `code` defaults to `ERR_GEN_REQ_ERROR`. All the errors below inherit this one.
 
 #### got.CacheError
 
--- a/readme.md
+++ b/readme.md
@@ -1705,7 +1705,7 @@ Additionaly, the errors may have `request` (Got Stream) and `response` (Got Resp
 
 #### got.RequestError
 
-When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied, `code` defaults to `ERR_GEN_REQ_ERROR`. All the errors below inherit this one.
+When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied, `code` defaults to `ERR_GOT_REQUEST_ERROR`. All the errors below inherit this one.
 
 #### got.CacheError
 
@@ -1730,7 +1730,7 @@ When the server response code is not 2xx nor 3xx if `options.followRedirect` is 
 
 #### got.MaxRedirectsError
 
-When the server redirects you more than ten times. Includes a `response` property. Contains a `code` property with `ERR_MAX_REDIRECTS`.
+When the server redirects you more than ten times. Includes a `response` property. Contains a `code` property with `ERR_TOO_MANY_REDIRECTS`.
 
 #### got.UnsupportedProtocolError
 
--- a/readme.md
+++ b/readme.md
@@ -1705,43 +1705,44 @@ Additionaly, the errors may have `request` (Got Stream) and `response` (Got Resp
 
 #### got.RequestError
 
-When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. All the errors below inherit this one.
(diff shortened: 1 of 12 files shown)


sindresorhus:
I think it should be done on all errors at once so we can ensure we do it consistently.

And we should use a better code than `HTTPError`. It doesn't need to include the word `Error` either.

Probably best to follow the Node.js convention: https://nodejs.org/api/errors.html#nodejs-error-codes

Drewfergusson:
Hey, I was working on a solution for this but now I'm a little confused. 

I made this PR against the main branch because I saw others doing that in the closed PR's. I just realized how different the v11.8.2 branch is from the main branch. The main branch has errors broken out into a separate file and is completely missing the UnsupportedProtocolError class the v11.8.2 has.  

Am I missing something here? I searched for how to contribute but if you could point me to any docs that would be helpful 🙏

szmarczak:
> The main branch [...] is missing the UnsupportedProtocolError class the v11.8.2 has.

Correct. The logic has been moved to `options.ts` where all normalization happens. The error is later replaced by `RequestError`.

https://github.com/sindresorhus/got/blob/bc9e48ae689ceb132a192652f5ff6aeedbb4f56c/source/core/options.ts#L1168

https://github.com/sindresorhus/got/blob/bc9e48ae689ceb132a192652f5ff6aeedbb4f56c/source/core/index.ts#L509

Just omit setting `.code` for this one. Errors originating from `options.ts` need a separate discussion.

szmarczak:
> I searched for how to contribute but if you could point me to any docs that would be helpful 🙏

You can submit a PR to the `v11` branch, so another `v11` version will be released.

Note that current development for Got v12 is taking place in the `main` branch.

Drewfergusson:
@szmarczak let me know any feedback on this latest.

I realized that some of the error objects have a 'code' property passed in (but not always) and didn't want to overwrite that if it contained a more specific error code for users. So I changed the code to only add the error codes if one is not passed in. One change from our discussion was adding a default code for the parent RequestError object of `ERR_GEN_REQ_ERROR` for situations where no code is passed in.

readme.md
```diff
@@ -1705,43 +1705,44 @@ Additionaly, the errors may have `request` (Got Stream) and `response` (Got Resp
 
 #### got.RequestError
 
-When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. All the errors below inherit this one.
+When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied `code` defaults to `ERR_GEN_REQ_ERROR` All the errors below inherit this one.
```
szmarczak, line 1708: I think that `ERR_GOT_REQUEST_ERROR` would suit this a bit better. Wdyt?

readme.md
```diff
@@ -1705,43 +1705,44 @@ Additionaly, the errors may have `request` (Got Stream) and `response` (Got Resp
...
 #### got.HTTPError
 
-When the server response code is not 2xx nor 3xx if `options.followRedirect` is `true`, but always except for 304. Includes a `response` property.
+When the server response code is not 2xx nor 3xx if `options.followRedirect` is `true`, but always except for 304. Includes a `response` property. Contains a `code` property with `ERR_NON_2XX_3XX_RESPONSE` or a more specific failure code.
+
 
 #### got.MaxRedirectsError
 
-When the server redirects you more than ten times. Includes a `response` property.
+When the server redirects you more than ten times. Includes a `response` property. Contains a `code` property with `ERR_MAX_REDIRECTS`.
```
szmarczak, line 1733: What do you think about `ERR_TOO_MANY_REDIRECTS`?

readme.md
sindresorhus, line 1708:
```suggestion
When a request fails. Contains a `code` property with error class code, like `ECONNREFUSED`. If there is no specific code supplied, `code` defaults to `ERR_GEN_REQ_ERROR`. All the errors below inherit this one.
```

sindresorhus:
The readme updates needs to be applied to the TypeScript types docs comments too.

readme.md
Drewfergusson: I like it. I didn't really love using 'general' error and this is more verbose.

Drewfergusson:
Updates based on your feedback have been made!

szmarczak (approved):

sindresorhus:
I'm confused, why is this pull request targeting the v11 branch? The common practice is to always first land a change on the main branch and then a feature branch (v11) if needed.

Drewfergusson:
Yeah, with some adaptation this can be added to the v12 branch. Let me know if you want me to work on a separate PR for the v12 branch.

szmarczak:
@sindresorhus Let me know if this PR looks good. If so, I'll submit a PR for the `main` branch.

szmarczak:
I manually merged this into Got 12: f27e8d331627074e76e4a9ec85a2c8d0a0625e5f

PR #1795, opened 2021-07-22
Rename `httpsOptions` to `https`

szmarczak:
@sindresorhus TBH I'd prefer if we named this `httpsOptions` because there is `cacheOptions`. But I leave this up to you :)


sindresorhus:
It's only named `cacheOptions` because `cache` is already used for passing a cache adapter.

If you ignore the consistency argument, do you actually prefer `httpsOptions` over `https`? I like clarity, but I don't feel `httpsOptions` adds any clarity over `https` when it's part of the options. I just look at it as a namespace to cantina multiple options. Kinda like how the `retry` option is a namespace for `retry` related options. It's not named `retryOptions`.

szmarczak:
Right, let's go with `https` then.

PR #1796, opened 2021-07-23
Custom options

### LINKED ISSUE #1794: Custom options
Got v12 throws when an option does not exist. So the workaround is:

```js
	hooks: {
		init: [
			(raw, options) => {
				if ('token' in raw) {
					options.context.token = raw.token;
					delete raw.token;
				}
			}
		]
	},
```

this way `got('...', {token: 'secret'})` will set `options.context.token`. I think there should be a dedicated API for custom options. Maybe this:

```js
const instance = got.extend({
	get example() {
		// `this` returns a store (an object)
		return this.example;
	},
	set example(value) {
		// `this` returns a store (an object)
		this.example = value;
	}
});
```

or this:

```js
import got from 'got';

class MyOptions extends Options {
	get example() {
		return this._example;
	}

	set example(value) {
		this._example = value;
	}
}

const instance = got.extend({Options: MyOptions});
```

or both?

/cc @sindresorhus @mnmkng

szmarczak:
https://github.com/sindresorhus/got/blob/56ee3ac294f8637d5e6ac2467cdf2bfcc86848ec/test/create.ts#L384-L406

Fixes #1794

### DIFF
--- a/source/core/options.ts
+++ b/source/core/options.ts
@@ -1,8 +1,14 @@
+/* eslint-disable import/order */
 import {promisify, inspect} from 'util';
 import {URL, URLSearchParams} from 'url';
 import {checkServerIdentity} from 'tls';
 import {request as httpRequest} from 'http';
 import {request as httpsRequest} from 'https';
+import is, {assert} from '@sindresorhus/is';
+import lowercaseKeys from 'lowercase-keys';
+import CacheableLookup from 'cacheable-lookup';
+import http2wrapper, {ClientHttp2Session} from 'http2-wrapper';
+import parseLinkHeader from './parse-link-header.js';
 import type {Readable} from 'stream';
 import type {Socket} from 'net';
 import type {SecureContextOptions, DetailedPeerCertificate} from 'tls';
@@ -15,26 +21,18 @@ import type {
 	Agent as HttpsAgent,
 } from 'https';
 import type {InspectOptions} from 'util';
-import is, {assert} from '@sindresorhus/is';
-import lowercaseKeys from 'lowercase-keys';
-import CacheableLookup from 'cacheable-lookup';
-import http2wrapper, {ClientHttp2Session} from 'http2-wrapper';
 import type CacheableRequest from 'cacheable-request';
 import type ResponseLike from 'responselike';
 import type {IncomingMessageWithTimings} from '@szmarczak/http-timer';
 import type {CancelableRequest} from '../as-promise/types.js';
-import parseLinkHeader from './parse-link-header.js';
+import type {Promisable, OmitIndex} from '../utils.js';
 import type {PlainResponse, Response} from './response.js';
 import type {RequestError} from './errors.js';
 import type {Delays} from './timed-out.js';
 
-type Promisable<T> = T | Promise<T>;
-
 const [major, minor] = process.versions.node.split('.').map(v => Number(v)) as [number, number, number];
 
 export type DnsLookupIpVersion = undefined | 4 | 6;
-
-type Except<ObjectType, KeysType extends keyof ObjectType> = Pick<ObjectType, Exclude<keyof ObjectType, KeysType>>;
 
 export type NativeRequestOptions = HttpsRequestOptions & CacheOptions & {checkServerIdentity?: CheckServerIdentityFunction};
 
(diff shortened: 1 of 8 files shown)


szmarczak:
I would appreciate any help on how to fix this:

```ts
class Example {
    foo: number;
    bar: string;

    constructor() {
        this.foo = 1;
        this.bar = '';
    }

    [key: string]: unknown;
}

type Except<ObjectType, KeysType extends keyof ObjectType> = Pick<ObjectType, Exclude<keyof ObjectType, KeysType>>;
const test: Except<Example, 'bar'> = new Example();

console.log(test.foo + 5);

```

szmarczak:
I'm -1 on `options.yourOptionHere` because TypeScript requires stupid hacks.

szmarczak:
A much better solution that requires no hacks is:

```js
import got from './dist/source/index.js';

// Save all non-Got options to `context`
const instance = got.extend({
    hooks: {
        init: [
            (raw, options) => {
                for (const key in raw) {
                    if (!(key in options)) {
                        options.context[key] = raw[key];
                        delete raw[key];
                    }
                }
            }
        ]
    }
});

const {headers} = await instance('https://httpbin.org/anything', {
    foo: 'bar',
    hooks: {
        beforeRequest: [
            options => {
                options.headers.foo = options.context.foo;
            }
        ]
    }
}).json();

console.log(headers.Foo); //=> 'bar'
```

source/core/options.ts
```diff
@@ -857,6 +865,14 @@ export default class Options {
 			this.isStream = options.isStream!;
 		}
 
+		if ('custom' in options) {
+			this.custom = options.custom!;
+
+			for (const option of options.custom!) {
+				this[option] = undefined;
```
szmarczak, line 872: There should be an if to prevent resetting already set options.

source/create.ts
```diff
@@ -23,7 +22,7 @@ const delay = async (ms: number) => new Promise(resolve => {
 	setTimeout(resolve, ms);
 });
 
-const isGotInstance = (value: Got | ExtendOptions): value is Got => (
+const isGotInstance = (value: any): value is Got => (
```
sindresorhus, line 25:
```suggestion
const isGotInstance = (value: unknown): value is Got => (
```

szmarczak:
Note: for now custom options will **not** be visible on `console.log(request.options)`. There are two ways to fix this:

1. Store the custom options inside a private object
2. Store the custom options inside `options.context`. Meaning `options.yourOptionHere` and `options.context.yourOptionHere` would be exactly the same (yes, you'd still can do `options.yourOptionHere = 5`).
3. No fix

/cc @sindresorhus @mnmkng

szmarczak:
I've made some changes. Custom options are now stored inside `options._custom`. This has zero performance impact (done via [`Proxy`](https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy)`.

szmarczak:
LMK what you think. I'd mark this as experimental, so we can remove this in Got 13 (if it happens to be messy).

IMO `context` is [the way](https://github.com/sindresorhus/got/pull/1796#issuecomment-886092841) to go - we would have to add this to docs.

I'm ok with merging this.

mnmkng:
Looking at the code I think it adds too much complexity just for the ability to have top level custom options. The workaround with `init` hooks is good enough. Maybe just a simple documentation update "how to add custom options" would do the trick. Sorry for putting the idea into your head. It was a lot of work and I never expected this.

szmarczak:
The changes aren't that big. Let me merge some general improvements to `main`.

szmarczak:
Now the changes reflect `custom options` only.

Please note that TypeScript will error on compilation with the `init` hook approach, without the TypeScript fixes people would need to do:

```diff
+						// @ts-expect-error
                        options.context[key] = raw[key];
+						// @ts-expect-error
                        delete raw[key];
```

Unless everybody switches to `context`.

szmarczak:
The smallest `context` approach in `gh-got` looks like this: https://gist.github.com/szmarczak/26036a6ec82910fc550865e667d0712e/revisions#diff-157d9c0fb7d926d0490317599fc9d7a03e74ff169154d2062ea6cfbd4156024b

szmarczak:
@sindresorhus Let me know if we should merge this or document how to use `options.context` for custom top-level options.

test/arguments.ts
```diff
@@ -621,7 +618,6 @@ test('options have url even if some are invalid - got.extend', async t => {
 	});
 
 	await t.throwsAsync(instance('https://example.com', {
-		// @ts-expect-error
```
sindresorhus, line 624: I think this illustrates why this change is not an improvement. It makes the types looser.

sindresorhus:
> document how to use options.context for custom top-level options.

Custom options sounds nice in theory, but it seems to add too much complexity for such an edge-case. Most users will not need this.

Issue #1810, opened 2021-07-27
Support multipart/form-data encoding


octet-stream:
#### What problem are you trying to solve?

General support for spec-compatible FormData implementations.

#### Describe the feature

Current form-data encoding in Got relies on [`form-data`](https://npmjs.com/package/form-data) package implementation which is not spec-compatible and [lacks of most of FormData interface methods](https://github.com/form-data/form-data/issues/124). I suggest to add general support for this type of request body. How this can be done:

* Detect spec-compatible FormData objects in request body: this will allow to use either [`formdata-node`](https://npmjs.com/package/formdata-node), or [`formdata-polyfill`](https://npmjs.com/package/formdata-polyfill), or any other implementation as long as it follows the spec;
* Implement in-house form-data encoder or just use [`form-data-encoder`](https://npmjs.com/package/form-data-encoder) package instead;

References:

* https://github.com/node-fetch/node-fetch/issues/1167
* https://github.com/octet-stream/form-data/issues/1

If you don't mind to use `form-data-encoder` to handle this encoding, I can file a PR to bring its support.

#### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.

szmarczak:
Got v12 supports async iterators so https://www.npmjs.com/package/formdata-node should work out of the box.

szmarczak:
https://github.com/jimmywarting/FormData is a no as of now - Got doesn't support Blobs

octet-stream:
> Got v12 supports async iterators so npmjs.com/package/formdata-node should work out of the box.

Builtin encoding will be removed from formdata-node in next major release, because it must be handled by HTTP client. This suggestion is not about support for some specific package, but rather for `multipart/form-data` encoding in general.

szmarczak:
Please consider removing [`nanoid`](https://packagephobia.com/result?p=nanoid). It weights 52KB.

octet-stream:
Yeah, I will do something about this.

octet-stream:
Decided to just use crypto.randomBytes for boundary string, should be more than enough. Kind of don't want to use just Math.radmom() for this, but it would've be fine too.

Available in form-data-encoder 1.1.0, now it weights 35 KB. I will reduce the size once again when I drop CJS support.

szmarczak:
What's wrong with `Math.random()`?

octet-stream:
> What's wrong with Math.random()?

[...]

PR #1835, opened 2021-08-12
Add support for `FormData` request body

octet-stream:
This PR brings general support for `multipart/form-data` encoding and allows to use any spec-comliant FormData object as request body. This PR also introduces a new dependency [`form-data-encoder`](https://npmjs.com/package/form-data-encoder), so you might expect it to add about 37 KB of extra size into the install size of got.

Before it's done I would like to get a feedback on few more things:

- How should I deal with custom `content-type` header? The encoder will generate this header and use internally to produce body content and if user will override this header - body will not be send as of now. Should I create and use the encoder in request body anyway?
- Do I need some more tests? Perhaps for request body? The `encoder.encode()` method returns `AsyncIterableIterator` so it basically reuse the logic for async iterator body consuming.
- How would I document these changes? Should I add alternative packages (both formdata-node and formdata-polyfill) into the docs?

I will keep this PR as WIP until my questions is resolved.

Further suggestions: As I said before, the good old [`form-data`](https://npmjs.com/package/form-data) package is not spec-compliant and I think it should be deprecated. At least node-fetch will do so in near future (see refs section of this PR for more information).

Refs:
- https://github.com/sindresorhus/got/issues/1810
- https://github.com/node-fetch/node-fetch/issues/1167
- https://github.com/node-fetch/node-fetch/pull/1212

#### Checklist


octet-stream:
Ah, I forgot to mention that while tests are passing, I have issues with XO. But they not seem to be related to this PR.

szmarczak:
Thanks for the draft! Looking forward to this ^_^

> How should I deal with custom content-type header? The encoder will generate this header and use internally to produce body content and if user will override this header - body will not be send as of now.

It should be sent no matter the content type.

> Do I need some more tests? Perhaps for request body?

You read my mind :) Basically read the payload on the server and check if it's valid.

> How would I document these changes? Should I add alternative packages (both formdata-node and formdata-polyfill) into the docs?

:+1:

> the good old form-data package is not spec-compliant and I think it should be deprecated.

I agree.

/cc @sindresorhus

sindresorhus:
👍 

This will also need documentation updates.

octet-stream:
> It should be sent no matter the content type.

So I should keep header set by developer and not override it. Right?

> You read my mind :) Basically read the payload on the server and check if it's valid.

👍

> I agree.

Should I include form-data deprecation in this PR? If so, how do mention this? With `util.deprecate` or/and documentation? I'm asking in case if got has some guidelines for this case :)

> This will also need documentation updates.

Will do!

szmarczak:
> So I should keep header set by developer and not override it. Right?

Correct. Only set `content-type` if `noContentType`.

> With util.deprecate or/and documentation?

:+1: (and)

> Should I include form-data deprecation in this PR?

To be honest, I'm not sure. I'm leaving this up to @sindresorhus

documentation/2-options.md
```diff
@@ -290,6 +290,22 @@ console.log(data);
 //=> 'Hello, world!'
 ```
 
+Since Got 12 you can use spec-compliant FormData objects as request body, such as [`formdata-node`](https://github.com/octet-stream/form-data) or [`formdata-polyfill`](https://github.com/jimmywarting/FormData):
+
+```js
+import got from 'got';
+
```
szmarczak, line 297:
```suggestion
```

octet-stream:
Hmm. Do you guys have a body parser to handle FomrData bodies in your test server?

szmarczak:
> Do you guys have a body parser to handle FomrData bodies in your test server?

We don't. Feel free to include one.

octet-stream:
About `then-busboy`: To be honest, I am not familiar with express ecosystem and if there's any multipart/form-data parser that can produce a single object with files and fields. I just needed a parser that will help be to get files, read them and send back along with the fields. Maybe `multer` can handle this? It seems like they allow to upload only files with specific field names (only those you mention in .array(), .fields() and .single()), otherwise if you will use multer.any() you will get files as an array.

I was ended up trying to make a bare minimal parser, but realised that I basically re-implement my solution... Sooo I just added it for tests. It seem to work as expected.

octet-stream:
I think I gonna mark this PR as ready for a review and wait for your feedback, because I do not have any more significant changes for this feature.

octet-stream:
Looks like I have fixed XO errors related to this PR.

<details>
<summary>For some reason I also have these errors on my local machine.</summary>

image


</details>

But they don't seem to be a thing in a previous run of CI workflow.

documentation/2-options.md
```diff
@@ -290,6 +290,21 @@ console.log(data);
 //=> 'Hello, world!'
 ```
 
+Since Got 12 you can use spec-compliant FormData objects as request body, such as [`formdata-node`](https://github.com/octet-stream/form-data) or [`formdata-polyfill`](https://github.com/jimmywarting/FormData):
```
sindresorhus, line 293:
```suggestion
Since Got 12, you can use spec-compliant [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) objects as request body, such as [`formdata-node`](https://github.com/octet-stream/form-data) or [`formdata-polyfill`](https://github.com/jimmywarting/FormData):
```

octet-stream:
I think this tests failing is not related to the PR...

documentation/2-options.md
```diff
@@ -290,6 +290,21 @@ console.log(data);
...
+import {FormData} from 'formdata-node'; // or 'formdata-polyfill/esm.min.js'
+
+const form = new FormData();
+
+form.set('greeting', 'Hello, world!');
+
+const {data} = await got.post('https://httpbin.org/post', {
+	body: form
+});
+```
```
sindresorhus, line 306: The docs here needs to be added to TS as doc comment too.

octet-stream: You mean in Options class?

sindresorhus: Yes, where the other `body` option docs are.

octet-stream: Done.

szmarczak:
Big thank you @octet-stream ❤️

octet-stream:
It was a pleasure :)

tutods:
Is possible now use Got with multipart/formdata?

PR #1918, opened 2021-11-13
Fix `deflate` responses not being decompressed

iOnline247:
I think this should be deflate.

#### Checklist


szmarczak:
Thanks, this is an important fix! Can you add a test please?

iOnline247:
Admittedly, I'm new to the code base. Was just reading through and noticed an odd magic string. Is there an example test that would be a good boilerplate to follow? I'm guessing this one would be a good starting point.

https://github.com/sindresorhus/got/blob/main/test/promise.ts#L33-L41

dynamiclover:
This should be merged as a fix.

**Reasoning:**
There's no test coverage currently for deflate or it would have caught the typo. 

The CI failures seem spurious/unrelated to this change.

PR #1940, opened 2021-12-06
Export `Request` type

### LINKED ISSUE #1939: Missing Request in exported types
#### Describe the bug

The `Request` class is exported as the default export from `source/core/index`.
In `source/index`, the types from `source/core/index` are exported with an aggregation, like so: `export * from 'source/core/index.js'`: https://github.com/sindresorhus/got/blob/49c605ca85fa35b7c1b6819219cb7bd5209b4085/source/index.ts#L19
This re-export don't propagate the default export from `source/core/index.js`: https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export
Hence the `Request` type is lost in the re-export.

It is not in the list of exported type: https://github.com/sindresorhus/got/blob/main/documentation/typescript.md
So, I am not sure if this missing type is by design, or not. Feel free to close if this type is expected to be hidden.

#### Actual behavior

```typescript
import got from 'got'

let request: ReturnType<got.stream>
try {
  request = got.stream(url)
} catch(error) {
  // handle error
}
```

#### Expected behavior

```typescript
import got, {Request} from 'got'

let request: Request
try {
  request = got.stream(url)
} catch(error) {
  // handle error
}
```

#### Checklist

etnbrd:
This PR exports the `Request` class, and update the typescript documentation accordingly.

I didn't found any export related tests, but if provided some pointers where they live, I would be happy to write one.

Fixes #1939

#### Checklist

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -16,6 +16,7 @@ export {got};
 export {default as Options} from './core/options.js';
 export * from './core/options.js';
 export * from './core/response.js';
+export {default as Request} from './core/index.js';
 export * from './core/index.js';
 export * from './core/errors.js';
 export {Delays} from './core/timed-out.js';
--- a/source/index.ts
+++ b/source/index.ts
@@ -16,7 +16,7 @@ export {got};
 export {default as Options} from './core/options.js';
 export * from './core/options.js';
 export * from './core/response.js';
-export {default as Request} from './core/index.js';
+export type {default as Request} from './core/index.js';
 export * from './core/index.js';
 export * from './core/errors.js';
 export {Delays} from './core/timed-out.js';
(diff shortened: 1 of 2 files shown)


source/index.ts
```diff
@@ -16,6 +16,7 @@ export {got};
 export {default as Options} from './core/options.js';
 export * from './core/options.js';
 export * from './core/response.js';
+export {default as Request} from './core/index.js';
```
sindresorhus, line 19:
```suggestion
export type {default as Request} from './core/index.js';
```

?

I don't think we want to export the actual class as that's an implementation detail.

etnbrd: Ah yes, fine with me, thanks 🙇

szmarczak: @sindresorhus is right - currently `Request` is only meant to be used as an interface. Got does a bit more than just `new Request`:

https://github.com/sindresorhus/got/blob/a405e7cb545ef56e8f3158ea9f14fffe8fb2da6e/source/create.ts#L57-L59

PR #1951, opened 2021-12-13
Fix docs link

alextes:


alextes:
@sindresorhus
that Sindre..
fast like a ninja..

I had an awesome gif of a cat walking down a fridge vertically like a ninja to insert here but my internet is too slow right now :sob:

Issue #1093, opened 2020-02-28
Use `fastify` in tests instead of `express`


szmarczak:
They support HTTP2! :sparkles:

PR #1957, opened 2022-01-04
Add HTTP/1 and HTTP/2 timings tests

etnbrd:
This PR implements tests to make sure the `timings` and `timings.phases` values are not missing in both http/1 and http/2.
It is currently failing for an http/2 request.

#### Checklist

### DIFF
--- a/test/timings.ts
+++ b/test/timings.ts
@@ -1,14 +1,18 @@
 import test from 'ava';
 import got from '../source/index.js';
+import withServer from './helpers/with-server.js';
 
-test('http/1 timings', async t => {
-	const {timings} = await got('https://httpbin.org/anything');
+test('http/1 timings', withServer, async (t, server, got) => {
+	server.get('/', (_request, response) => {
+		response.end('ok');
+	});
+
+	const {timings} = await got('');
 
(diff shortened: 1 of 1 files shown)


test/timings.ts
```diff
@@ -0,0 +1,50 @@
+import test from 'ava';
+import got from '../source/index.js';
+
+test('http/1 timings', async t => {
+	const {timings} = await got('https://httpbin.org/anything');
```
szmarczak, line 5: Can you use a local server for this one please? https://github.com/sindresorhus/got/blob/2ac07e1b60f59e2219bd6c2809a9e40f56b146b6/test/http.ts#L35-L41

etnbrd: Just for the http/1 test? It seems the `withServer` helper only support http/1.

Would it be feasible to implement the http/2 support within `withServer`? I gave a quick look, and since I saw the implementation for an https server, I guess it would be possible?

szmarczak: This is another issue: https://github.com/sindresorhus/got/issues/1093

szmarczak:
Thanks! :tada:

PR #1959, opened 2022-01-09
Fix `nock` compatibility

huturen:
It seems that `got` does not compatible with `nock` as of version 12.

This PR makes `got` be compatible with `nock`.

**Version before 12:**
got2

**As of version 12:**
got1

**Cause:**
`nock`  implements mocking by hacking native http.request/https.request object.
The way of destructuring is not compatible with `nock`.

---

#### Checklist

### DIFF
--- a/source/core/options.ts
+++ b/source/core/options.ts
@@ -3,8 +3,8 @@ import {Buffer} from 'node:buffer';
 import {promisify, inspect} from 'node:util';
 import {URL, URLSearchParams} from 'node:url';
 import {checkServerIdentity} from 'node:tls';
-import {request as httpRequest} from 'node:http';
-import {request as httpsRequest} from 'node:https';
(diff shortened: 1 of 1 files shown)


source/core/options.ts
```diff
@@ -3,8 +3,8 @@ import {Buffer} from 'node:buffer';
 import {promisify, inspect} from 'node:util';
 import {URL, URLSearchParams} from 'node:url';
 import {checkServerIdentity} from 'node:tls';
-import {request as httpRequest} from 'node:http';
-import {request as httpsRequest} from 'node:https';
+import http from 'node:http';
+import https from 'node:https';
```
sindresorhus, line 7: Add a short comment about it, otherwise this could accidentally be regressed in the future.

szmarczak (approved):

PR #1972, opened 2022-01-30
Document `response` event in stream docs

### LINKED ISSUE #1955: How to check response status code before creating a write stream

Im trying to download images from another server using `got.stream()`. I noticed that it creates empty files if the image does not exist on the other server. I tried to check for a `200` status code on the response before creating a write stream and piping the stream to it. However the `response.status` does not exist so i checked the docs about `got.stream()` but i couldn't find anything to check for a successful response first.

How can i avoid creating empty files? This is my code (simplified):

```js
import got from 'got'
import { pipeline } from 'stream/promises'
import { createWriteStream } from 'fs'

async function main () {
  const readStream = got.stream('https://example.com/image.png')

  // how to exit early here?
  if (readStream.status !== 200) {
    return
  }

  // create file and save response
  await pipeline(readStream, createWriteStream('image.png'))
}

main()
```

#### Checklist

dedwardstech:
I've added documentation in the form of an example around the got.stream response event.

Fixes #1955
 
#### Checklist

### DIFF
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -174,7 +174,7 @@ This is emitted when a HTTP response is received.
 
 ```js
 import got from 'got';
-import { pipeline } from 'pipeline';
+import { pipeline } from 'node:stream/promises';
 import { createWriteStream } from 'node:fs';
 
 const readStream = got.stream('http://example.com/image.png', { throwHttpErrors: false });
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -164,6 +164,38 @@ Whether the socket was used for other previous requests.
 
 ## Events
 
+### `stream.on('response', ...)`
+
+#### `response`
+
+**Type: [`PlainResponse`](typescript.md#plainresponse)**
+
+This is emitted when a HTTP response is received.
+
+```js
+import got from 'got';
+import { pipeline } from 'pipeline';
+import { createWriteStream } from 'node:fs';
+
+const readStream = got.stream('http://example.com/image.png', { throwHttpErrors: false });
+
+readStream.on('response', async (response) => {
+	if (response.statusCode !== 200) {
+		console.log('failed');
+		readStream.destroy(); // destroy the stream to prevent hanging resources
+
+		return;
+	}
+
+	await pipeline(
+		readStream,
+		createWriteStream('image.png')
+	);
+
+	console.log('success');
+})
+```
+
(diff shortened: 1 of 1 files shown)


documentation/3-streams.md
```diff
@@ -164,6 +164,38 @@ Whether the socket was used for other previous requests.
...
+
+#### `response`
+
+**Type: [`PlainResponse`](typescript.md#plainresponse)**
+
+This is emitted when a HTTP response is received.
+
+```js
+import got from 'got';
+import { pipeline } from 'pipeline';
```
higoka, line 177: This should be `node:stream/promises`

dedwardstech: Oops, good catch! I'll fix it

documentation/3-streams.md
```diff
@@ -164,6 +164,37 @@ Whether the socket was used for other previous requests.
...
+This is emitted when a HTTP response is received.
+
+```js
+import {pipeline} from 'node:stream/promises';
+import {createWriteStream} from 'node:fs';
+import got from 'got';
+
+const readStream = got.stream('http://example.com/image.png', {throwHttpErrors: false});
+
+readStream.on('response', async response => {
```
sindresorhus, line 182: Using async function for an event listener is an anti-pattern as any thrown error will become an unhandled error. Instead, wrap the async stuff in an async-IIFE and use try/catch to properly handle any error.

szmarczak: If there's a missing try/catch, there will be an unhandled error regardless whether it's an async function or an async IIFE.

szmarczak (approved):

documentation/3-streams.md
sindresorhus: Hence, why I'm saying there should be a try/catch here. I think only `pipeline` could throw here, so it should be:

```js
(async () => {
	try {
		await pipeline(
	 		readStream,
	 		createWriteStream('image.png')
	 	);
	} catch {
		// handle it
	}
})();
```

Right?

szmarczak: This can be one way. Another is:

```js
readStream.on('response', async response => {
	try {
		  await pipeline(
			  readStream,
			  createWriteStream('image.png')
		  );
	} catch (error) {
		// ...
	}
});
```

szmarczak: I've modified the code. Should be fine now.

PR #1975, opened 2022-02-05
Update engines.node version

JoCat:
#### Checklist

Specified a corrective version for the package to work
"node:" prefix requires version 14.18 and above
https://nodejs.org/api/esm.html#node-imports


sindresorhus:
> Added node: import support to require(...).

That's `require`, not `import`.

JoCat:
Yes, it looks like esbuild uses require instead of import. That's why it throws an error.

Issue #2013, opened 2022-03-20
does `json` option support all json values?


samzilverberg:
I was trying to use GOT (latest) with Typescript to write some "niche" tests that send some json that probably only people trying to break your server would use.

values such as: `true`, `false`, `null`, `"123"`, `1`, `[1,2,3]`.

But noticed that GOT only supports `Record<string, any>`.

here's the source of it:

https://github.com/sindresorhus/got/blob/d95ceeafb16f306e11f5d7251a82ba073252cbee/source/core/options.ts#L1356-L1365

This means only objects and array can be passed in.
Shouldnt `json` support all valid json values?

I tried looking up documenation or issues to understand why this limitation exists but didn't find any :(

If this limitation is here to stay, then is there a way to use GOT to send the other json values?

szmarczak:
In that case value should be typed as `any`.

samzilverberg:
@szmarczak dya mean the type for the `set json(...)` function?

it can be typed better than `any` :)  , for example:
```
set json(value: Record<string, any> | number | string | boolean | undefined) { 
```

szmarczak:
https://github.com/microsoft/TypeScript/blob/45c1e5880936a56b61071d14c929674e6f595c93/lib/lib.es5.d.ts#L1066

TypeScript ships `JSON.stringify` with `any` so let's stick with that.

samzilverberg:
@szmarczak i made the proposed `any` change in https://github.com/sindresorhus/got/pull/2015 

I wanted to add some tests but wasn't sure which test file would be most appropriate.
I didn't find any exising tests of what `json` option should and shouldn't accept.

samzilverberg:
@szmarczak i'm closing this issue.

note that sindresorhus suggested to use unknown for the return value of `get json`

https://github.com/sindresorhus/got/pull/2015#discussion_r832804667

i can confirm that using `unknown` instead of `any` compiles and all tests run (on my machine).

PR #2015, opened 2022-03-22
fix: allow other json values

samzilverberg:
related to https://github.com/sindresorhus/got/issues/2013

@szmarczak proposed to use `any` for the json value type.

it works and passes all current tests on my machine.

if you prefer to avoid `any` and use types then the following also works and passes tests:

```
	get json(): Record<string, any> | string | number | boolean | undefined {
		return this._internals.json;
	}

	set json(value: Record<string, any> | string | number | boolean | undefined) {
		assert.any([is.object, is.string, is.number, is.boolean, is.null_, is.undefined], value);
		...
```

### DIFF
--- a/source/core/options.ts
+++ b/source/core/options.ts
@@ -1349,13 +1349,11 @@ export default class Options {
 
 	__Note #2__: This option is not enumerable and will not be merged with the instance defaults.
 	*/
-	get json(): Record<string, any> | undefined {
+	get json(): any {
 		return this._internals.json;
 	}
 
-	set json(value: Record<string, any> | undefined) {
-		assert.any([is.object, is.undefined], value);
-
+	set json(value: any) {
 		if (value !== undefined) {
 			assert.undefined(this._internals.body);
 			assert.undefined(this._internals.form);


szmarczak:
Thanks!

source/core/options.ts
```diff
@@ -1349,13 +1349,11 @@ export default class Options {
 
 	__Note #2__: This option is not enumerable and will not be merged with the instance defaults.
 	*/
-	get json(): Record<string, any> | undefined {
+	get json(): any {
```
sindresorhus, line 1352: @szmarczak I think it should return `unknown`.

szmarczak: Well, it was `Record<string, any>` and not `Record<string, unknown>` 🤔

sindresorhus: The existing value was incorrect too, yes.

sindresorhus: Although, they should behave the same, I prefer using `unknown` whenever possible for consistency. I only use `any` when there's absolutely no other way.

sindresorhus: https://github.com/sindresorhus/got/commit/7ed87983477decafed6a115391a6fbafd33693fd

szmarczak: :+1:

PR #2019, opened 2022-03-28
fix link to Promise on MDN

peterbe:
The link for "Promise" goes to [the Polish translation on MDN](https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Promise).  It should be without language so people get MDN in their desired language instead. 

#### Checklist


szmarczak:
Thanks! 🙈

PR #2020, opened 2022-03-29
Support `AbortController`

### LINKED ISSUE #1511: Support `AbortController`
Node.js 15 now supports `AbortController` and it makes sense for us to support it in Got. While I'm personally not a big fan of its overly verbose way of handling cancellation, it seems to become the de-facto way of handling cancellation going forward.

I think we should still retain the current `.cancel()` method if possible.

https://medium.com/@nodejs/node-js-v15-0-0-is-here-deb00750f278
https://developer.mozilla.org/en-US/docs/Web/API/AbortController
https://github.com/nodejs/node/commit/8390f8a86b

jopemachine:
Fixes #1511.

Thanks for this awesome lib! :)

I added `signal` option for supporting `AbortController`.

when `AbortController.abort` is called, throws `DOMException` in the promise of `asPromise.ts`.

About tests, I copied, pasted and fixed `cancel.ts` to `abort.ts`.

I'd appreciate it if you could let me know if I made any mistakes.

#### Checklist

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -29,6 +29,12 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 	const emitter = new EventEmitter();
 
 	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
+		firstRequest?.on('error', (error: RequestError) => {
+			if (error.code === 'ERR_GOR_ABORT_ERROR') {
+				reject(error);
+			}
+		});
+
 		onCancel(() => {
 			globalRequest.destroy();
 		});
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -29,12 +29,6 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 	const emitter = new EventEmitter();
 
 	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
-		firstRequest?.on('error', (error: RequestError) => {
-			if (error.code === 'ERR_GOR_ABORT_ERROR') {
-				reject(error);
-			}
-		});
-
 		onCancel(() => {
 			globalRequest.destroy();
 		});
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -22,13 +22,26 @@ const proxiedRequestEvents = [
 	'downloadProgress',
 ];
 
-export default function asPromise<T>(firstRequest?: Request): CancelableRequest<T> {
+export default function asPromise<T>(firstRequest?: Request, signal?: AbortSignal): CancelableRequest<T> {
 	let globalRequest: Request;
 	let globalResponse: Response;
 	let normalizedOptions: Options;
 	const emitter = new EventEmitter();
 
 	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
+		if (signal) {
+			signal.addEventListener('abort', () => {
+				globalRequest.destroy();
+
+				// To do: remove below if statement when targets node is 17.3.0 or higher.
+				if ((signal as any).reason) {
+					reject((signal as any).reason);
+				} else {
+					reject(new DOMException('This operation was aborted', 'AbortError'));
+				}
+			});
+		}
+
 		onCancel(() => {
 			globalRequest.destroy();
 		});
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -5,7 +5,6 @@ import {
 	RequestError,
 	HTTPError,
 	RetryError,
-	AbortError,
 } from '../core/errors.js';
 import Request from '../core/index.js';
 import {parseBody, isResponseOk} from '../core/response.js';
@@ -23,39 +22,13 @@ const proxiedRequestEvents = [
 	'downloadProgress',
 ];
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-const getDOMException = (errorMessage: string) => globalThis.DOMException === undefined
-	? new AbortError(errorMessage)
-	: new DOMException(errorMessage);
-
-const getAbortedReason = (signal: AbortSignal) => {
-	// To do: Remove below any castings when '@types/node' targets 18
-	const reason = (signal as any).reason === undefined
-		? getDOMException('This operation was aborted.')
-		: (signal as any).reason;
-
-	return reason instanceof Error ? reason : getDOMException(reason);
-};
-
-export default function asPromise<T>(firstRequest?: Request, signal?: AbortSignal): CancelableRequest<T> {
+export default function asPromise<T>(firstRequest?: Request): CancelableRequest<T> {
 	let globalRequest: Request;
 	let globalResponse: Response;
 	let normalizedOptions: Options;
 	const emitter = new EventEmitter();
 
 	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
-		if (signal) {
-			if (signal.aborted) {
-				globalRequest.destroy();
-				reject(getAbortedReason(signal));
-			}
-
-			signal.addEventListener('abort', () => {
-				globalRequest.destroy();
-				reject(getAbortedReason(signal));
-			});
-		}
-
 		onCancel(() => {
 			globalRequest.destroy();
 		});
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -5,7 +5,6 @@ import {
 	RequestError,
 	HTTPError,
 	RetryError,
-	AbortError,
 } from '../core/errors.js';
 import Request from '../core/index.js';
 import {parseBody, isResponseOk} from '../core/response.js';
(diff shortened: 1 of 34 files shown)


jopemachine:
It seems that `DOMException` does not exist until Node 17.
Would it be possible to create and throw `AbortError` subclassing `RequestError` instead of throwing `DOMException`?
Or should the API's target be set to node 17 or higher?

sindresorhus:
See how we did it in https://github.com/sindresorhus/p-retry/pull/59

source/as-promise/index.ts
```diff
@@ -22,13 +23,30 @@ const proxiedRequestEvents = [
...
+	: new DOMException(errorMessage);
+
+export default function asPromise<T>(firstRequest?: Request, signal?: AbortSignal): CancelableRequest<T> {
 	let globalRequest: Request;
 	let globalResponse: Response;
 	let normalizedOptions: Options;
 	const emitter = new EventEmitter();
 
 	const promise = new PCancelable<T>((resolve, reject, onCancel) => {
+		if (signal) {
```
sindresorhus, line 38: Add a TODO comment to use `signal.throwIfAborted()` when targeting Node.js 18. That would simplify the code a lot. https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted

Same with your other pull requests.

jopemachine: I think throwing error here prevents error catching using `Promise#catch`.

Shouldn't we call `reject(signal.reason);` instead? (when targeting Node v18)


> Add a TODO comment to use `signal.throwIfAborted()` when targeting Node.js 18. That would simplify the code a lot. [developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted)
> 
> Same with your other pull requests.

I think throwing error here prevents error catching using `Promise#catch`.

Shouldn't we call `reject(signal.reason);` instead? (when targeting Node v18)

sindresorhus: Actually both, if you look at the example on MDN: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted#implementing_an_abortable_api

sindresorhus:
CI is failing. You may need to just rebase from `main` branch.

source/as-promise/index.ts
jopemachine: I think the example code was incorrect.
You can see about this on https://github.com/mdn/content/issues/15348 
So we won't need to call `throwIfAborted` here.

sindresorhus (approved):

szmarczak (changes requested):
I've done some logic changes. Previously the abort signal was only affecting the native HTTP stream, which is incorrect, because it doesn't abort the retry timeout (the request would fail though).

Can you fix the tests please?

jopemachine:
> I've done some logic changes. Previously the abort signal was only affecting the native HTTP stream, which is incorrect, because it doesn't abort the retry timeout (the request would fail though).
> 
> Can you fix the tests please?

Thanks for your review and for correcting this.

1. IMHO, I still think [`asPromise`'s update](https://github.com/sindresorhus/got/blob/6bdd99af5a070288f24165c81b7c2e3b2ece860f/source/as-promise/index.ts#L46-L57) is needed to reject the promise correctly and fix tests. I would appreciate your letting me know If I got wrong.

2. I deleted `globalRequest.destroy` from `asPromise` since it should be done with [this line](https://github.com/sindresorhus/got/pull/2020/commits/945167860552e5a2236af2a8e572108e721540ac#diff-c0192b59eaa0ed83a6ac9d7faeedc53698193feb222bb9e419b2cf46378db2e7R239). Other lines are same with previous's codes.

jopemachine:
> Why do you think so?

I think the AbortError should be caught in `Promise#catch`, or `try-catch` statement.
The error emitted through `this.destroy` cannot be caught.
To catch this, I think `reject` should be called in `gotPromise` when aborted.

But I noticed that I can simplify the code using the `Request`, so I reverted [723b960](https://github.com/sindresorhus/got/pull/2020/commits/723b96030f245837dc37dff2ed878bca274a5f6e), updated `asPromise`, `AbortError`.

szmarczak:
> I think the error emitted through this.destroy cannot be caught.
> To catch this, I think reject should be called in gotPromise when aborted.

It can be caught. `destroy()` makes Got's `Request` stream emit `error` which is caught [here](https://github.com/sindresorhus/got/blob/c25af092bf792c2c7e29db848a569f2eb3e527f8/source/as-promise/index.ts#L135) and the promise rejects [here](https://github.com/sindresorhus/got/blob/c25af092bf792c2c7e29db848a569f2eb3e527f8/source/as-promise/index.ts#L132).

jopemachine:
> It can be caught. `destroy()` makes Got's `Request` stream emit `error` which is caught [here](https://github.com/sindresorhus/got/blob/c25af092bf792c2c7e29db848a569f2eb3e527f8/source/as-promise/index.ts#L135) and the promise rejects [here](https://github.com/sindresorhus/got/blob/c25af092bf792c2c7e29db848a569f2eb3e527f8/source/as-promise/index.ts#L132).

Thanks for letting me know.
I never noticed that. I think I jumped to the conclusion that the errors had not been caught just since the test failed.
I just reverted the commits, fixed the tests.

szmarczak (approved):
LGTM. The only concern I have is the use of `AbortSignal` global. I'm not sure if it breaks when someone uses `@types/node` v14 in their project.

source/core/options.ts
```diff
@@ -1484,6 +1485,36 @@ export default class Options {
 		}
 	}
 
+	/**
+	You can abort the `request` using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
+
+	**Requires Node.js 16 or later.*
```
sindresorhus, line 1491:
```suggestion
	__Requires Node.js 16 or later.__
```

source/core/index.ts
```diff
@@ -235,6 +235,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			return;
 		}
 
+		this.options.signal?.addEventListener('abort', () => {
+			this.destroy(new Error('This operation was aborted.'));
```
sindresorhus, line 239: This needs the same as what I commented on the other AbortController PRs you did. 

https://github.com/sindresorhus/p-map/blob/a5faf425ad3f871d10b3a18e30ae10d1edc78311/index.js#L3-L31

And the error needs to be documented in the docs with the other errors.

sindresorhus (changes requested):

sindresorhus:
> The only concern I have is the use of AbortSignal global. I'm not sure if it breaks when someone uses @types/node v14 in their project.

I think it would. I suggest we just use `any` for now with a comment to use `AbortSignal` when targeting Node.js 16.

jopemachine:
I think for setting the error `code` value properly, maybe it would be good to make new Error class, `AbortError` which is inherited from `RequestError`.

And because `DOMException` is not `NodeJS.ErrnoException`, I think using `DOMException` is not possible when generating the `AbortError`, so the updated code shape looks different from other PRs' ones.

jopemachine:
I'm sorry, I was confused if I should throw a custom error here or native DOMException (when possible).

> Why is it not possible?

I thought using DOMException is not possible because I had thought AbortError should inherit `RequestError`, which is not correct.

jopemachine:
I think [this statement](https://github.com/sindresorhus/got/blob/main/documentation/8-errors.md?plain=1#L10) might need to be edited since AbortError doesn't have `code` and `request` object.

sindresorhus:
> I think [this statement](https://github.com/sindresorhus/got/blob/main/documentation/8-errors.md?plain=1#L10) might need to be edited since AbortError doesn't have code and request object.

You can add a sentence below it noting the exception.

source/core/options.ts
```diff
@@ -1488,7 +1488,7 @@ export default class Options {
 	/**
 	You can abort the `request` using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController).
 
-	**Requires Node.js 16 or later.*
+	__Requires Node.js 16 or later.__
```
szmarczak, line 1491: Please revert this change

szmarczak:
> I don't think we should inherit from RequestError.

Why is that? If we strip error metadata then people won't know which request was aborted except the fact that it was aborted. I'm -1 on this.

sindresorhus:
> Why is that? If we strip error metadata then people won't know which request was aborted except the fact that it was aborted. I'm -1 on this.

That's a good point. I agree it's more important with context than the consistency of using DOMException.

source/core/errors.ts
```diff
@@ -170,3 +170,13 @@ export class RetryError extends RequestError {
 		this.code = 'ERR_RETRYING';
 	}
 }
+
+/**
+An error to be thrown when the request is aborted by AbortController.
+*/
+export class AbortError extends RequestError {
+	constructor(request: Request, error?: Error) {
+		super('This operation was aborted.', {...error, code: 'ERR_ABORTED'}, request);
```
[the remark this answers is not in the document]
jopemachine: Since `error` (abortsignal.reason) is optional here, it seems passing `error` into the second argument of the constructor cause compile error.

```
// Compile error occurs

// Argument of type 'ErrnoException | undefined' is not assignable to parameter of type 'Partial<ErrnoException & { code?: string | undefined; }>'.
// Type 'undefined' is not assignable to type 'Partial<ErrnoException & { code?: string | undefined; }>'
super('This operation was aborted.', error, request);
this.code = 'ERR_ABORTED';
```

szmarczak:
Friendly reminder: can you address https://github.com/sindresorhus/got/pull/2020#issuecomment-1172927668 please?

source/core/errors.ts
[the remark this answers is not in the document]
szmarczak: I see, thanks for clarifying. In this case you can just skip this. Like so:

```js
super('This operation was aborted.', {}, request);
this.name = 'AbortError';
this.code = 'ERR_ABORTED';
```

The reason is user accessible via `controller.signal.reason`, so I don't think we need to expose it for now. That might be for https://github.com/sindresorhus/got/issues/1953

sindresorhus:
Thanks :)

PR #2043, opened 2022-05-17
Add `response.ok`

### LINKED ISSUE #2036: Add a response.ok boolean similar to fetch API?
It's quite handy and would ease migrating to got.

It's set to false if `(res.statusCode < 200 || res.statusCode > 299)` or true if in 200-299.

IswaryaS:
Fixes #2036

Added a boolean ok to the response from got API to indicate whether status code is within 200 & 299 i.e. 2xx.

#### Checklist

- [X] I have included some tests.

### DIFF
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -359,7 +359,7 @@ Whether the response comes from cache or not.
 **Type: `boolean`**
 
 **Note:**
-> - This property need not be checked if throwHttpErrors is true.
+> - Got throws automatically when `response.ok` is `false` and `throwHttpErrors` is `true`.
 
 Whether the response was successful (status code in the range 200-299).
 
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -354,6 +354,15 @@ The server's IP address.
 
 Whether the response comes from cache or not.
 
+### `ok`
+
+**Type: `boolean`**
+
+**Note:**
+> - This property need not be checked if throwHttpErrors is true.
+
+Whether the response was successful (status code in the range 200-299).
+
 ### `statusCode`
 
 **Type: `number`**
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -358,10 +358,13 @@ Whether the response comes from cache or not.
 
 **Type: `boolean`**
 
-**Note:**
-> - Got throws automatically when `response.ok` is `false` and `throwHttpErrors` is `true`.
+Whether the response was successful
 
-Whether the response was successful (status code in the range 200-299).
+**Note:**
+> - A request is successful when the status code of the final request is `2xx` or `3xx`.
+> - When [following redirects](2-options.md#followredirect), a request is successful **only** when the status code of the final request is `2xx`.
+> - `304` responses are always considered successful.
+> - Got throws automatically when `response.ok` is `false` and `throwHttpErrors` is `true`.
 
 ### `statusCode`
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
You need to update the docs too.

source/core/response.ts
```diff
@@ -91,6 +91,11 @@ export interface PlainResponse extends IncomingMessageWithTimings {
 	The result of the request.
 	*/
 	body?: unknown;
+
+	/**
+	Whether the response status code is 2xx.
```
sindresorhus, line 96:
```suggestion
	Whether the response was successful (status code in the range 200-299).
```

sindresorhus: Should also mention that it's not necessary to check this property unless `{throwHttpErrors: false}`.

test/http.ts
```diff
@@ -375,3 +375,43 @@ test('ClientRequest can throw before promise resolves', async t => {
...
+test('status code 200 has response ok is true', withServer, async (t, server, got) => {
+	server.get('/', (_request, response) => {
+		response.statusCode = 200;
+		response.end();
+	});
+
+	const promise = got('');
+	await t.notThrowsAsync(promise);
+	const {statusCode, body, ok} = await promise;
+	t.is(ok, true);
```
sindresorhus, line 388:
```suggestion
	t.true(ok);
```

source/core/index.ts
```diff
@@ -632,6 +633,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		typedResponse.isFromCache = (this._nativeResponse as any).fromCache ?? false;
 		typedResponse.ip = this.ip;
 		typedResponse.retryCount = this.retryCount;
+		typedResponse.ok = (statusCode >= 200 && statusCode <= 299);
```
szmarczak, line 636: https://github.com/sindresorhus/got/blob/c25af092bf792c2c7e29db848a569f2eb3e527f8/source/core/response.ts#L109-L114

szmarczak (changes requested):

source/core/index.ts
IswaryaS: Should I check followRedirect & limit status code similar to this before assigning boolean to ok?

documentation/3-streams.md
```diff
@@ -354,6 +354,15 @@ The server's IP address.
 
 Whether the response comes from cache or not.
 
+### `ok`
+
+**Type: `boolean`**
+
+**Note:**
+> - This property need not be checked if throwHttpErrors is true.
```
szmarczak, line 362:
```suggestion
> - Got throws automatically when `response.ok` is `false` and `throwHttpErrors` is `true`.
```

documentation/3-streams.md
```diff
@@ -354,6 +354,15 @@ The server's IP address.
...
+
+Whether the response was successful (status code in the range 200-299).
```
szmarczak, line 364: This is not exactly correct, the range depends on `followRedirects`

szmarczak (changes requested):

documentation/3-streams.md
IswaryaS: Will update this

documentation/3-streams.md
szmarczak: 👍

IswaryaS:
@szmarczak Updated the docs :)

szmarczak (approved):

szmarczak:
Thanks! :tada:

Issue #2046, opened 2022-05-25
Add option to disable UNIX domain sockets


sindresorhus:
https://github.com/sindresorhus/got/blob/main/documentation/tips.md#unix-domain-sockets

It's a niche feature and have heard that some teams have their own custom code to prevent it, so it makes sense to have the ability to turn it off in Got.

I suggest:

- [ ] Add option to turn off support for UNIX domain sockets. Make it `false` by default.
- [ ] In Got v13 (next major release), change it to be `true` by default.

TommyDew42:
Haha, I'll pick up this one!
So when the setting is turned to `false`, the following won't work?
```js
got('unix:/var/run/docker.sock:/containers/json');
```

prsmahajan:
Is there anything to work with this? I guess, the issue is resolved, the maintainers should mark as closed.

PR #2047, opened 2022-05-25
Disable redirects to UNIX sockets

szmarczak:


source/core/index.ts
```diff
@@ -727,6 +727,11 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				const redirectBuffer = Buffer.from(response.headers.location, 'binary').toString();
 				const redirectUrl = new URL(redirectBuffer, url);
 
+				if ((url as URL).hostname !== 'unix' && (redirectUrl.protocol === 'unix:' || redirectUrl.hostname === 'unix')) {
```
sindresorhus, line 730: I think this would read better if we made a `isUnixDomainSocketURL` function and used it for both `url` and `redirectUrl`.

sindresorhus (changes requested):

source/core/index.ts
szmarczak: PTAL

source/core/index.ts
```diff
@@ -23,6 +23,7 @@ import calculateRetryDelay from './calculate-retry-delay.js';
 import Options, {OptionsError, OptionsInit} from './options.js';
 import {isResponseOk, Response} from './response.js';
 import isClientRequest from './utils/is-client-request.js';
+import isUnixSocketURL from './utils/is-unix-url.js';
```
sindresorhus, line 26: Super nitpick, but you know me..

```suggestion
import isUnixSocketURL from './utils/is-unix-socket-url.js';
```

szmarczak: PTAL

szmarczak:
@lpinca Yes, this is expected. Got has no knowledge whether to trust the `http:` resource or not. There's no difference when visiting directly that resource first, skipping the earlier `unix:` visit. Similarly, browsers have implemented CORS to prevent cross-origin requests. You can get the next `unix:` location via `error.response.headers.location`.

If that's not sufficient, I guess we could have an option to select desired behavior (#2046). Is the `http:` server `localhost`? If so, we could whitelist IPs in the `localhost` namespace.

> Also, unrelated to this PR but 2 can leak sensitive headers if the socket path is not the same. For example http://unix/foo:/ -> http://unix/bar:/.

Thanks for reporting! I've opened #2069. That's rather an unlikely occurrence, since `bar:` needs to be a malicious process or be exploitable.

lpinca:
Anyway I'm not even sure if `http:` -> `unix:` should be disabled. In my opinion it is not different from a redirect to `localhost` and that is supported in all browsers.

lpinca:
> What is the use case for this?

No real use case, was just wondering why it was disabled.

> Servers running on UNIX sockets tend to be more "important", for example `docker`.

Hmm, unconvinced. You would need a process running as root or as a member of the Docker group to access it.

Abuchtela (commented):
gh pr checkout 2047

Issue #1979, opened 2022-02-08
Consider allowing single hooks.


wtgtybhertgeghgtwtg:
#### What problem are you trying to solve?

Currently, hooks must be passed in as arrays.

```ts
const instance = got.extend({
  hooks: {
    init: [
      (plain, options) => {
        // Do something.
      }
    ],
    beforeRedirect: [
      (options, plainResponse) => {
        // Do something.
      }
    ],
    beforeRequest: [
      options => {
        // Do something.
      },
      options => {
        // Do something else.
      }
    ],
  }
});
```

It's not particularly burdensome, but it's a bit more verbose than it needs to be.

#### Describe the feature

Ideally, hooks could be passed in without wrapping them in an array.

```ts
const instance = got.extend({
  hooks: {
    // No array.
    init: (plain, options) => {
      // Do something.
    },
    // Also acceptable.
    beforeRedirect(options, plainResponse) {
      // Do something.
    },
    // Still works.
    beforeRequest: [
      options => {
        // Do something.
      },
      options => {
        // Do something else.
      }
    ],
  }
});
```

#### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.

sindresorhus:
Makes sense to me. Most users only use a single hook for each type.

TommyDew42:
This feature would make got even more user-friendly! Can I pick this up?

szmarczak:
@TommyDew42 There's no PR for this currently, so I guess you're free to pick this up.

szmarczak:
All you need is `npm install` and `npm test` (sometimes `npx xo` for manual linting).

sindresorhus:
Turned out this was not worth the complexity it would bring. See: https://github.com/sindresorhus/got/pull/2053#issuecomment-1172926091

PR #2053, opened 2022-06-05
Allow single hooks

TommyDew42:
### What this PR does?
For issue [1970](https://github.com/sindresorhus/got/issues/1979), it allows users to pass a single hook instead of alway an array of hooks to the hooks argument.

### How it is tested?
Unit tests.

#### Checklist

  - [ ] Will open a separate PR for that.

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -75,8 +75,9 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 
 				try {
 					const hooks = options.hooks.afterResponse;
+					const hooksArray = is.array(hooks) ? hooks : [hooks];
 
-					for (const [index, hook] of hooks.entries()) {
+					for (const [index, hook] of hooksArray.entries()) {
 						// @ts-expect-error TS doesn't notice that CancelableRequest is a Promise
 						// eslint-disable-next-line no-await-in-loop
 						response = await hook(response, async (updatedOptions): CancelableRequest<Response> => {
@@ -89,7 +90,7 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 
 							// Remove any further hooks for that request, because we'll call them anyway.
 							// The loop continues. We don't want duplicates (asPromise recursion).
-							options.hooks.afterResponse = options.hooks.afterResponse.slice(0, index);
+							options.hooks.afterResponse = hooksArray.slice(0, index);
 
 							throw new RetryError(request);
 						});
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -75,7 +75,7 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 
 				try {
 					const hooks = options.hooks.afterResponse;
-					const hooksArray = is.array(hooks) ? hooks : [hooks];
+					const hooksArray = [hooks].flat();
 
 					for (const [index, hook] of hooksArray.entries()) {
 						// @ts-expect-error TS doesn't notice that CancelableRequest is a Promise
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -376,7 +376,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 					}
 
 					try {
-						for (const hook of this.options.hooks.beforeRetry) {
+						const hooksArray = is.array(this.options.hooks.beforeRetry) ? this.options.hooks.beforeRetry : [this.options.hooks.beforeRetry];
+						for (const hook of hooksArray) {
 							// eslint-disable-next-line no-await-in-loop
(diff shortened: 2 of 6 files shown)


source/as-promise/index.ts
```diff
@@ -75,8 +75,9 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 
 				try {
 					const hooks = options.hooks.afterResponse;
+					const hooksArray = is.array(hooks) ? hooks : [hooks];
```
sindresorhus, line 78:
```suggestion
					const hooksArray = [hooks].flat();
```

source/core/options.ts
```diff
@@ -847,12 +847,12 @@ const cloneInternals = (internals: typeof defaultInternals) => {
 		},
 		timeout: {...internals.timeout},
 		hooks: {
-			init: [...hooks.init],
-			beforeRequest: [...hooks.beforeRequest],
-			beforeError: [...hooks.beforeError],
-			beforeRedirect: [...hooks.beforeRedirect],
-			beforeRetry: [...hooks.beforeRetry],
-			afterResponse: [...hooks.afterResponse],
+			init: is.array(hooks.init) ? [...hooks.init] : hooks.init,
```
sindresorhus, line 850: I think it's better to use the native methods when available. `Array.isArray()`.

source/as-promise/index.ts
TommyDew42: Is it purely for readability? One difference is `[hooks].flat()` create a shallow copy of `hooks`

source/core/options.ts
TommyDew42: I took a look at `is.array`. Is preference to `Array.isArray()` because we don't need logic related to `assertion` argument?

Screenshot 2022-06-06 at 7 42 55 PM

source/as-promise/index.ts
```diff
@@ -75,8 +75,9 @@ export default function asPromise<T>(firstRequest?: Request): CancelableRequest<
 
 				try {
 					const hooks = options.hooks.afterResponse;
+					const hooksArray = [hooks].flat();
```
szmarczak, line 78: Normalization should occur inside the `Options` class

source/core/index.ts
```diff
@@ -376,7 +376,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 					}
 
 					try {
-						for (const hook of this.options.hooks.beforeRetry) {
+						const hooksArray = is.array(this.options.hooks.beforeRetry) ? this.options.hooks.beforeRetry : [this.options.hooks.beforeRetry];
```
szmarczak, line 379: dtto

source/core/index.ts
```diff
@@ -760,7 +761,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				updatedOptions.prefixUrl = '';
 				updatedOptions.url = redirectUrl;
 
-				for (const hook of updatedOptions.hooks.beforeRedirect) {
+				const hooksArray = [updatedOptions.hooks.beforeRedirect].flat();
```
szmarczak, line 764: dtto

source/core/index.ts
```diff
@@ -1061,7 +1063,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 		let request: ReturnType<Options['getRequestFunction']> | undefined;
 
-		for (const hook of options.hooks.beforeRequest) {
+		const hooksArray = [options.hooks.beforeRequest].flat();
```
szmarczak, line 1066: dtto

szmarczak (changes requested):
Hooks should be an object with setters and getters:

```js
let beforeRequest;
...

return {
	get beforeRequest() {
		return beforeRequest;
	},
	set beforeRequest(value) {
		beforeRequest = [value].flat();
	},
	...
};

szmarczak:
@sindresorhus I'm not sure if the code changes are actually worth it. It's easier to put `[]` in the code instead...

sindresorhus:
> I'm not sure if the code changes are actually worth it. It's easier to put [] in the code instead...

I generally agree, but the most common way is to only have a single hook, so the convenience is kinda nice.

> Also https://github.com/microsoft/TypeScript/issues/43662 makes it extremely hard... We would be forced to constantly write as ...[].

That is a bummer indeed and not something I was aware of. I'm unsure whether it's worth it now.

TommyDew42:
Yea it's really a tradeoff between dev exp and user exp. Maybe it's close this for now before we figure out a solution that does good for both DX & UX?

szmarczak:
Eventually we could switch to single hooks for .extend and drop the array [in the typings], but then we would have different typings for .extend and for Got instance options...

sindresorhus:
Closing this for now. We can reconsider again when the TS limitation is resolved.

PR #2098, opened 2022-08-03
Add note about empty request body

mifi:
#### Checklist

I literally spent hours trying to figure out why my requests were hanging, when sending a POST, but not when sending a GET. And only when using isStream: true, not with isStream: false

### DIFF
--- a/documentation/3-streams.md
+++ b/documentation/3-streams.md
@@ -23,7 +23,8 @@ This constructor takes the same arguments as the Got promise.
 > If the `body`, `json` or `form` option is used, this stream will be read-only.
 
 **Note:**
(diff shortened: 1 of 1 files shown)


sindresorhus:
We should either just gracefully handle this or throw an error.

szmarczak:
The note is already here.

image

mifi:
Thanks! I missed that. Maybe I can add an example for the no-body use case too?

mifi:
I updated the PR. An alternative solution to avoid issues like this could be to have one `isStream` option for the request and one for the response, maybe `isRequestStream` and `isResponseStream`.

szmarczak (approved):

PR #2100, opened 2022-08-04
Don't freeze signal when freezing Options

### LINKED ISSUE #2099: Options freezing breaks abort signal
The `Options` objects [freeze the abort signal][1] if it gets passed. As a result setting the `signal` prop ends up breaking the signal under node as the builtin code sets private properties when calling `AbortController.abort()`.

This makes it impossible to pass a default abort signal to a `Got` instance as the runtime will throw when aborting.

- NodeJS `v16.15.1`
- Got `v12.3.0`

[1]: https://github.com/sindresorhus/got/blob/e032b60ff3285403b8f9627dcdd32aae39867d7a/source/core/options.ts#L2524

```javascript
import got from 'got';
const abortCtrl = new AbortController();
const got2 = got.extend({
    signal: abortCtrl.signal,
});
abortCtrl.abort();
```

```console
$ node test.mjs
node:internal/abort_controller:279
  signal[kAborted] = true;
                   ^

TypeError: Cannot assign to read only property 'Symbol(kAborted)' of object '#<AbortSignal>'
    at abortSignal (node:internal/abort_controller:279:20)
    at AbortController.abort (node:internal/abort_controller:315:5)
    at file:///tmp/test.mjs:6:11
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

svvac:
Freezing the abort signal object breaks its functionality under Node.

Fixes #2099

### DIFF
--- a/test/normalize-arguments.ts
+++ b/test/normalize-arguments.ts
@@ -1,5 +1,4 @@
 import {URL, URLSearchParams} from 'url';
-import {EventEmitter} from 'events';
 import test from 'ava';
 import got, {Options} from '../source/index.js';
 
--- a/test/normalize-arguments.ts
+++ b/test/normalize-arguments.ts
@@ -1,4 +1,5 @@
 import {URL, URLSearchParams} from 'url';
+import {EventEmitter} from 'events';
 import test from 'ava';
 import got, {Options} from '../source/index.js';
 
@@ -167,3 +168,18 @@ test('searchParams - multiple values for one key', t => {
 		['100', '200', '300'],
 	);
 });
+
+if (globalThis.AbortSignal !== undefined) {
+	test('signal does not get frozen', t => {
+		const controller = new AbortController();
+		const {signal} = controller;
+
+		const options = new Options({
+			url: new URL('http://localhost'),
+			signal,
+		});
+		options.freeze();
+
+		t.is(Object.isFrozen(options.signal), false);
+	});
+}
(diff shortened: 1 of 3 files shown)


szmarczak:
Can you add a test please?

svvac:
What shall I test? That the signal is frozen? That it can be aborted?

svvac:
Added tests :

- Asserts the signal does not get frozen when calling `Options.freeze()`
- Asserts that a signal passed as a default option to `Got.extend()` can be aborted (detection will depend on the runtime)

szmarczak (approved):

svvac:
Re: failure under node 14, would you rather have the freeze test skipped, or the signal object be mocked when `AbortController` is not defined?

szmarczak:
Skip the test like in `test/abort.ts`

szmarczak:
@svvac Please avoid force-pushing unless you need to fix merge conflicts. It really makes code review harder.

test/normalize-arguments.ts
```diff
@@ -1,4 +1,5 @@
 import {URL, URLSearchParams} from 'url';
+import {EventEmitter} from 'events';
```
szmarczak, line 2:
```suggestion
```

szmarczak:
Thanks! 🙌🏼

sindresorhus:
https://github.com/sindresorhus/got/releases/tag/v12.3.1

Issue #2103, opened 2022-08-09
Hook `beforeError` does not respect the `throwHttpErrors` setting


Supinic:
#### Describe the bug

 - Node.js version: `18.0.0`
 - OS & version:  `Windows 10`, `Raspbian Buster`, `Ubuntu 20.04`
 - Got version: `11.6.0` and higher

#### Actual behavior

The `beforeError` hook fires regardless of whether or not the `throwHttpErrors` option is set to `false` and the request fails. This has been (erroneously, it seems?) changed in #1384 - more specifically in [as-promise/index.ts](https://github.com/sindresorhus/got/pull/1384/files#diff-4fd462a09629aa458e4fad7c6ec4a0393527acc587b657fff4f3d774490881a2L150-R120). This change wasn't mentioned anywhere in the `11.6.0` release, or in other releases since.

The behaviour is correct in all versions up until `11.6.0`, where this PR was merged.

#### Expected behavior

The `beforeError` hook should not fire if the `throwHttpErrors` option is set to `false`, and the request fails.

#### Code to reproduce

```js
const got = require("got");
const opts = {
    throwHttpErrors: false,
    hooks: {
        beforeError: [
            (e) => {
                console.log("beforeError hook", e.name);
            }
        ]
    }
};

(async () => {
    const res = await got.get("https://httpstat.us/400", opts);
    console.log("Response", res.body);
})();
```

In versions prior to `11.6.0`, the request response will be logged into console. In versions since, the `beforeError` hook will take precedence, log the error from it, and the response will not be logged. 

As an alternative to a code-based fix, the documentation could be adjusted instead, as it currently contains [no reference](https://github.com/sindresorhus/got/blob/main/documentation/9-hooks.md?plain=1#L267-L299) to such behaviour.

#### Checklist
- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

szmarczak:
This unfortunately won't be easy to fix. In order to make a retry, an error is required. Then the `beforeRetry` hooks are called. If the error is still there (e.g. still getting 500) then `beforeError` is called. The error is then passed to the Promise implementation, it checks if `options.throwHttpErrors` is false. If it is, then returns `error.response` instead.

It wasn't possible to retry streams until https://github.com/sindresorhus/got/pull/1384 - to make the same logic there we would need to replace

[...]

PR #2104, opened 2022-08-09
Don't call beforeError hooks with HTTPError if !throwHttpErrors

szmarczak:
Fixes https://github.com/sindresorhus/got/issues/2103

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -793,6 +793,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			return;
 		}
 
+		// `HTTPError`s always have `error.response.body` defined.
+		// Therefore we cannot retry if `options.throwHttpErrors` is false.
+		// On the last retry, if `options.throwHttpErrors` is false, we would need to return the body,
+		// but that wouldn't be possible since the body would be already read in `error.response.body`.
 		if (options.isStream && options.throwHttpErrors && !isResponseOk(typedResponse)) {
 			this._beforeError(new HTTPError(typedResponse));
 			return;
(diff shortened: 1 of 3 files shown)


source/core/index.ts
```diff
@@ -793,6 +793,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			return;
 		}
 
+		// `HTTPError`s always have `error.response.body` defined.
+		// Therefore we cannot retry if `options.throwHttpErrors` is false.
+		// On the last retry, if `options.throwHttpErrors` is false, we would need to return the body,
+		// but that wouldn't be possible since the body would be already read in `error.response.body`.
```
sindresorhus, line 799: If I'm understanding this correctly, it cannot be read multiple times because it's a stream, right? Maybe we could buffer the stream and fake it somehow?

If not, we need to at least document the behavior.

szmarczak: > If I'm understanding this correctly, it cannot be read multiple times because it's a stream, right?

Correct.

> Maybe we could buffer the stream and fake it somehow?

That's extremely hacky and would result in double the RAM usage. I'm not saying I'm against it, I'd just prefer if we didn't do that. For example, if we get 1GB of data, RAM usage would be 2GB since we would have a string form and a buffer form.

Although we already double RAM usage for Promise API when `responseType` is set to something different to `string`.

sindresorhus: It could be stored as buffer though and then if string was needed, it could be a getter. But I agree it may not be worth the complication, so I suggest we just document the behavior.

szmarczak: If it's a getter then it won't be visible in logs :/

sindresorhus: Getters are not enumerable by default, but you can make a getter enumerable.

szmarczak: Wouldn't it look like this:

```
{
body: [ getter: body ]
}
```

?

sindresorhus: https://stackoverflow.com/questions/28072671/how-can-i-get-console-log-to-output-the-getter-result-instead-of-the-string-ge

szmarczak: Ah yes, forgot about `util.inspect.custom` - however that would be a hack on top of another hack :joy: not worth it imo

sindresorhus: I agree. I think it's enough to just do the mentioned docs changes.

Issue #2051, opened 2022-06-02
Got 12.1.0 x Typescript 4.7.2 x Node 16.15 transpilation fails with moduleResolution=Node16


ClementValot:
#### Describe the bug

 - Node.js version: 16.15.0
 - OS & version: Windows 11

#### Actual behavior

Requiring got and transpiling with Typescript 4.7 raises transpilation errors 
```
node_modules/got/dist/source/core/options.d.ts(763,22): error TS2709: Cannot use namespace 'CacheableLookup' as a type.
node_modules/got/dist/source/core/options.d.ts(764,26): error TS2709: Cannot use namespace 'CacheableLookup' as a type.
node_modules/got/dist/source/core/options.d.ts(775,21): error TS2709: Cannot use namespace 'CacheableLookup' as a type.
node_modules/got/dist/source/core/options.d.ts(776,25): error TS2709: Cannot use namespace 'CacheableLookup' as a type.
node_modules/got/dist/source/core/options.d.ts(1125,29): error TS2709: Cannot use namespace 'CacheableLookup' as a type.
```

#### Code to reproduce

package.json
```json
{
  "name": "got-ts4.7-regression",
  "version": "1.0.0",
  "type": "module",
  "dependencies": {
    "got": "^12.1.0"
  },
  "devDependencies": {
    "typescript": "^4.7.2"
  }
}
```
tsconfig.json
```json
{
  "compilerOptions": {
    "module": "Node16",
    "moduleResolution": "Node16",
    "target": "es6",
  },
  "exclude": [
    "node_modules"
  ]
}
```

"moduleResolution": "Node16" is [apparently required](https://github.com/microsoft/TypeScript/issues/33079#issuecomment-1128910470) for Typescript 4.7 support of the `exports` field in package.json

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

akowald:
Same, a little bit confused here. I cannot get this package to compile with TypeScript after the switch to ESM. Setting  `skipLibCheck` to `true` in `compilerOptions` is the only way to get this package to compile. Using TypeScript 4.7.3 and got 12.1.0.

I feel like I'm missing something, does anyone have a barebones repo with TypeScript working?

```
[...]

PR #2109, opened 2022-08-13
Upgrade Ava and @sindresorhus/tsconfig 

ClementValot:
Following discussions on https://github.com/sindresorhus/got/issues/2051

- Upgrade `ava` to V4 
- Minor packages upgrades
- Removing tsconfig property noPropertyAccessFromIndexSignature
- Upgrade `@sindresorhus/tsconfig` to 3.0.1 

#### Checklist

- [ X ] I have read the documentation.
- [ X ] I have included a pull request description of my changes.

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -45,12 +45,12 @@ export interface Progress {
 	total?: number;
 }
 
-const supportsBrotli = is.string(process.versions.brotli);
+const supportsBrotli = is.string(process.versions['brotli']);
 
 const methodsWithoutBody: ReadonlySet<string> = new Set(['GET', 'HEAD']);
 
 export type GotEventFunction<T> =
-	/**
+/**
 	`request` event to get the request object of the request.
 
 	 __Tip__: You can use `request` event to abort requests.
@@ -62,17 +62,17 @@ export type GotEventFunction<T> =
 	got.stream('https://github.com')
 		.on('request', request => setTimeout(() => request.destroy(), 50));
 	```
-	*/
+ */
 	((name: 'request', listener: (request: ClientRequest) => void) => T)
 
 	/**
 	The `response` event to get the response object of the final request.
-	*/
+	 */
 	& (<R extends Response>(name: 'response', listener: (response: R) => void) => T)
 
 	/**
 	The `redirect` event to get the response object of a redirect. The second argument is options for the next request to the redirect location.
-	*/
+	 */
 	& (<R extends Response, N extends Options>(name: 'redirect', listener: (response: R, nextOptions: N) => void) => T)
 
 	/**
@@ -103,7 +103,7 @@ export type GotEventFunction<T> =
 
 	console.log(response);
 	```
-	*/
+	 */
 	& ((name: 'uploadProgress' | 'downloadProgress', listener: (progress: Progress) => void) => T)
 	/**
 	To enable retrying on a Got stream, it is required to have a `retry` handler attached.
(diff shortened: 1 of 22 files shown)


source/core/index.ts
```diff
@@ -103,15 +103,15 @@ export type GotEventFunction<T> =
...
+	 */
 	& ((name: 'uploadProgress' | 'downloadProgress', listener: (progress: Progress) => void) => T)
 	/**
 	To enable retrying on a Got stream, it is required to have a `retry` handler attached.
 
 	When this event is emitted, you should reset the stream you were writing to and prepare the body again.
 
 	See `got.options.retry` for more information.
-	*/
+	 */
```
sindresorhus, line 114: Don't make unrelated changes

sindresorhus (changes requested):

tsconfig.json
```diff
@@ -2,12 +2,10 @@
 	"extends": "@sindresorhus/tsconfig",
 	"compilerOptions": {
 		"outDir": "dist",
-		"target": "es2020", // Node.js 14
-		"lib": [
-			"es2020"
+	    "moduleResolution": "Node",
```
sindresorhus, line 5: This should not be here

sindresorhus (changes requested):

source/core/index.ts
ClementValot: IDE auto-indent shenanigans :( 
Fixed

tsconfig.json
ClementValot: Atm got 12.3.1 has `moduleResolution=Node` inherited from `@sindresorhus/tsconfig 2.0.0`

Switching to `@sindresorhus/tsconfig 3.0.1` sets `moduleResolution` to `Node16` which breaks type checking when building got (due to issues with how some packages interact with `moduleReolution=Node16`, see debugging attempts in https://github.com/sindresorhus/got/issues/2051#issuecomment-1214196483 )

Overriding moduleResolution to Node here allows to upgrade tsconfig without breaking the package. 
It should ultimately be removed when the resolution with the problematic packages is okay. 

I added a comment in the file, does that sound reasonable?

ClementValot:
I botched unit tests because ava cannot interpret typescript transpiled with module:Node16 
I fixed it by adding module:ES2020 to ts-node. 

Quick question @szmarczak, I see you're a maintainer of https://github.com/avajs/typescript, should it be added to this project? 

Also unit test can't run properly on my setup, but there are a few that actually broke due to a FakeTimer issue, looking into it  
```
  timeout › lookup timeout

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Can\'t install fake timers twice on the same global object.',
  }

  › Object.install (node_modules/@sinonjs/fake-timers/src/fake-timers-src.js:1642:19)
  › exec (file:///test/helpers/with-server.ts:18:38)

```

sindresorhus:
> Minor packages upgrades
Removing tsconfig property noPropertyAccessFromIndexSignature

These are unrelated changes and should be done in a separate pull request.

sindresorhus:
> Quick question @szmarczak, I see you're a maintainer of https://github.com/avajs/typescript, should it be added to this project?

`ts-node` is better in our case as it means it will not need to rewrite paths and stuff.

sindresorhus:
Closing in favor of https://github.com/sindresorhus/got/commit/86308155d1bc43c635773183db908af902922549 and https://github.com/sindresorhus/got/issues/2051#issuecomment-1236513684. Help welcome getting the dependencies fixed.

PR #2112, opened 2022-08-16
Add `proxy` option

szmarczak:
We're in kinda stalemate position when it comes to proxies with Got. Many agents are unreliable and that drives off Got users, because they need to look for more and more agents. This might be better off in upper-level packages like [`got-scraping`](https://github.com/apify/got-scraping). Ideally this should be implemented in lower level, but unfortunately due to how Node.js designed the `http` module it looks kinda ugly.

@sindresorhus Please confirm do we want this in Got? If so, I'll continue with the tests & docs.

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -34,6 +34,7 @@ import {
 	CacheError,
 	AbortError,
 } from './errors.js';
+import { getProxyAgents } from './proxy/proxy.js';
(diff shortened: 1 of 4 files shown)


sindresorhus:
If you want to commit to maintaining this and the support that comes with it, I'm not against it, but I personally will ignore anything related to proxies.

PR #2120, opened 2022-08-25
Support FormData without known length

### LINKED ISSUE #2111: GOT not able to handle Passthrough streams
**Current Behaviour**
Hi, I am trying to upload the file stream to another server.
Please check the code below
```typescript
const fd = new FormData({
    readable: true
});
const stream = getFileStream();
fd.append('file', stream, {
    filename: 'upload.txt',
});
await got.post('http://localhost:3001/upload', {
    body: fd,
});
```

Hi, I am creating a file stream like the below snippet.

``` typescript
const getFileStream = () => {
    const fileStream = createReadStream(filePath);
    const passThrough = new PassThrough();
    return fileStream.pipe(passThrough);
};
```
I have to use passthrough for some reasons.

If I do something like the above I get an error
**'ERR_GOT_REQUEST_ERROR'**

But if I change **getFileStream** function like below it works basically I am not using Passthrough

``` typescript
const getFileStream = () => {
    const fileStream = createReadStream(filePath);
    const passThrough = new PassThrough();
    return fileStream;
};
```
If I do the same in Axios and node fetch it's working with passthrough.

**Expected Behaviour:**
It should upload the file

**Node Version** 
v14.18.2

octet-stream:
This PR updates fixes support for FormData instances without known length. In detail:

* Updated `form-data-encoder` dependency to version 2.1.0;
* Fixed usage of headers, returned by `FormDataEncoder` instance;
* Added a test for `got.post` requests using `FormData` without known length.

Fixes #2111

#### Checklist

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -570,7 +570,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 						headers['content-type'] = encoder.headers['Content-Type'];
 					}
 
-					headers['content-length'] = encoder.headers['Content-Length'];
+					if ('Content-Length' in encoder.headers) {
+						headers['content-length'] = encoder.headers['Content-Length'];
+					}
 
 					options.body = encoder.encode();
 				}
--- a/test/helpers/create-https-test-server.ts
+++ b/test/helpers/create-https-test-server.ts
@@ -5,7 +5,7 @@ import type {SecureContextOptions} from 'tls';
 import express from 'express';
 import pify from 'pify';
 import pem from 'pem';
-import type {CreateCSR, CreateCertificate} from '../types/pem.js';
+import type {CreateCsr, CreateCertificate} from '../types/pem.js';
 
 export type HttpsServerOptions = {
 	commonName?: string;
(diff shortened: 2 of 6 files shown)


octet-stream:
Some tests are failing, but they are not related to the PR. I don't know why is this happening.

sindresorhus:
Try to rebase from the `main` branch.

octet-stream:
Yeh, now it seem's to fail because of typescript choosing the last overload for functions.
I will try to fix this later (somehow), because I will likely have more free time this Saturday. Or should we just stick with pify 6.0.0 for this PR?

test/helpers/create-https-test-server.ts
```diff
@@ -68,7 +69,7 @@ const createHttpsTestServer = async (options: HttpsServerOptions = {}): Promise<
 
 	await pify(server.https.listen.bind(server.https))();
 
-	server.caKey = caKey;
+	server.caKey = caKey as any;
```
octet-stream, line 72: I wasn't sure what the exact meaning of this code, so I just added `as any` just in case if this assignment was without type casting to Buffer by intention. Should I convert it to buffer?

sindresorhus (approved):

source/core/index.ts
```diff
@@ -574,7 +574,9 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 						headers['content-type'] = encoder.headers['Content-Type'];
 					}
 
-					headers['content-length'] = encoder.headers['Content-Length'];
+					if ('Content-Length' in encoder.headers) {
+						headers['content-length'] = encoder.headers['Content-Length'];
+					}
```
szmarczak, line 579: Why the `if`?

octet-stream: Otherwise this header will be `undefined`, because the encoder does not include this header and I will get a `400 (Bad Request)` error in tests.

szmarczak:
Thanks! :heart:

PR #2161, opened 2022-10-10
Fix memory leak bug when re-using `abortSignal` (#2160)

### LINKED ISSUE #2160: Resource leak when (re-)using an abort signal
#### Describe the bug

 - Node.js version: v16.17.1
 - OS & version: Debian 5.10.140-1

When re-using an abort controller for multiple got requests a significant amount of memory is retained by the `AbortSignal` through got. The main reason for this (I think) is that the `Request` instance adds an event listener for the `abort`event on the passed signal where the event handler refers back to the request. The event handler is never removed so the signal retains a reference to the `Request` instance for as long as it lives.

#### Actual behavior

The code to reproduce below downloads and discards a 1MB file in a loop. The loop exists when the abort signal is aborted via ctrl-c. The same abort signal is passed to got to cancel the in-progress download. The heap profile screenshot below shows how a large chunk of memory is retained via the callback context of an event listener coming from here:
https://github.com/sindresorhus/got/blob/main/source/core/index.ts#L254

screenshot_20221010_221207

#### Expected behavior

Memory is released after the request completes.

#### Code to reproduce

```js
import got from 'got';

const abortController = new AbortController();

process.on('SIGINT', function () {
  abortController.abort();
});

async function main() {
  while (!abortController.signal.aborted) {
    await got.get(
      'https://ia902805.us.archive.org/14/items/1mbFile/1mb.mp4',
      {
        signal: abortController.signal,
      }
    ).buffer();
  }
}

main().then(() => console.log('Bye')).catch((e) => {
  if (e.constructor.name !== 'AbortError') {
    console.error(e);
  }
});
```

#### Checklist

- [*] I have read the documentation.
- [*] I have tried my code with the latest version of ~~Node.js~~ and Got.

jopemachine:
#### Checklist

This PR trying to fix #2160. (WIP)
`abortHandler` of request instance will be removed when the response end or error, abort.

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -251,9 +251,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			this.destroy(new AbortError(this));
 		}
 
-		this.options.signal?.addEventListener('abort', () => {
-			this.destroy(new AbortError(this));
-		});
+		this.abortHandler = this.abortHandler.bind(this);
+		this.options.signal?.addEventListener('abort', this.abortHandler, {once: true});
 
 		// Important! If you replace `body` in a handler with another stream, make sure it's readable first.
 		// The below is run only once.
@@ -545,6 +544,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		return this;
 	}
 
+	private abortHandler(): void {
+		this.destroy(new AbortError(this));
+	}
+
 	private async _finalizeBody(): Promise<void> {
 		const {options} = this;
 		const {headers} = options;
@@ -664,6 +667,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		response.once('end', () => {
 			this._responseSize = this._downloadedSize;
 			this.emit('downloadProgress', this.downloadProgress);
+			this.removeListener('abort', this.abortHandler);
 		});
 
 		response.once('error', (error: Error) => {
@@ -674,6 +678,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			response.destroy();
 
 			this._beforeError(new ReadError(error, this));
+			this.removeListener('abort', this.abortHandler);
 		});
 
 		response.once('aborted', () => {
@@ -684,6 +689,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 				message: 'The server aborted pending request',
 				code: 'ECONNRESET',
 			}, this));
+			this.removeListener('abort', this.abortHandler);
 		});
 
 		this.emit('downloadProgress', this.downloadProgress);


source/core/index.ts
```diff
@@ -545,6 +544,10 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		return this;
 	}
 
+	private abortHandler(): void {
```
szmarczak, line 547: nit

```suggestion
	private _abortHandler(): void {
```

source/core/index.ts
```diff
@@ -664,6 +667,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		response.once('end', () => {
 			this._responseSize = this._downloadedSize;
 			this.emit('downloadProgress', this.downloadProgress);
+			this.removeListener('abort', this.abortHandler);
```
szmarczak, line 670: This should be called on the signal instead

source/core/index.ts
```diff
@@ -674,6 +678,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			response.destroy();
 
 			this._beforeError(new ReadError(error, this));
+			this.removeListener('abort', this.abortHandler);
```
szmarczak, line 681: `response` is not guaranteed, but `request` is. However listening on `error` request event is also incorrect because of retries. I'd suggest to move this to `_destroy`.

marcelmeulemans:
There is a possible fix here if it helps: https://github.com/marcelmeulemans/got/commit/1bc49822de0d6e024a6070dc4e08839e7c3d8f06

jopemachine:
> There is a possible fix here if it helps: [marcelmeulemans@1bc4982](https://github.com/marcelmeulemans/got/commit/1bc49822de0d6e024a6070dc4e08839e7c3d8f06)

Great! I think this implementation fix #2160 correctly.
I checked this implementation removes the memory leak bug in the submited code.

스크린샷 2022-10-11 오전 8 22 30


I will close this PR due to wrong implementation, and there comes out possible fix now.

marcelmeulemans:
Would you like me to open a PR for that fix? If so any ideas for how to write a unit test for the issue?

jopemachine:
> Would you like me to open a PR for that fix? If so any ideas for how to write a unit test for the issue?

About unit test, I think you better ask for the question to this repository's maintainers.
By the way, I think we could have better code using the `?` operator like below.

https://github.com/jopemachine/got/commit/6d1223b3f7def1b666ae8878e6c1616b35bf541b

marcelmeulemans:
FYI https://github.com/sindresorhus/got/pull/2162

I considered using the `?` operator as you suggested but it seemed a shame to do the same nil check multiple times and the `if` block kinda nicely groups all abort signal code together i.m.o. If you feel strongly about it please a a comment and I will update.

PR #2162, opened 2022-10-10
Fix resource leak by removing abort event listeners on destroy

### LINKED ISSUE #2160: Resource leak when (re-)using an abort signal

marcelmeulemans:
This change removes the event listener callback that is added to the passed `AbortSignal` when the *got* instance is destroyed.

I'm not entirely sure that this fix covers all clean up points. I skipped the clean up on the redirect which I think is correct but would appreciate confirmation.

Also I'd be open to adding a test to cover this but any pointer on how to test this would be appreciated!

Fixes #2160

#### Checklist

- [*] I have read the documentation.
- [*] I have included a pull request description of my changes.

### DIFF
--- a/source/core/index.ts
+++ b/source/core/index.ts
@@ -176,6 +176,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 	private _triggerRead: boolean;
 	declare private _jobs: Array<() => void>;
 	private _cancelTimeouts: () => void;
+	private readonly _removeListeners: () => void;
 	private _nativeResponse?: IncomingMessageWithTimings;
 	private _flushed: boolean;
 	private _aborted: boolean;
@@ -199,6 +200,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		this._unproxyEvents = noop;
 		this._triggerRead = false;
 		this._cancelTimeouts = noop;
+		this._removeListeners = noop;
 		this._jobs = [];
 		this._flushed = false;
 		this._requestInitialized = false;
@@ -247,14 +249,6 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 			return;
 		}
 
-		if (this.options.signal?.aborted) {
-			this.destroy(new AbortError(this));
-		}
-
-		this.options.signal?.addEventListener('abort', () => {
-			this.destroy(new AbortError(this));
-		});
-
 		// Important! If you replace `body` in a handler with another stream, make sure it's readable first.
 		// The below is run only once.
 		const {body} = this.options;
@@ -270,6 +264,21 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 					};
 				}
 			});
+		}
+
+		if (this.options.signal) {
+			const abort = () => {
+				this.destroy(new AbortError(this));
+			};
+
+			if (this.options.signal.aborted) {
+				abort();
+			} else {
+				this.options.signal.addEventListener('abort', abort);
+				this._removeListeners = () => {
+					this.options.signal.removeEventListener('abort', abort);
+				};
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


source/core/index.ts
```diff
@@ -176,6 +176,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 	private _triggerRead: boolean;
 	declare private _jobs: Array<() => void>;
 	private _cancelTimeouts: () => void;
+	private readonly _removeListeners: () => void;
```
szmarczak, line 179:
```suggestion
	private _removeListeners: () => void;
```

source/core/index.ts
```diff
@@ -247,13 +249,20 @@ export default class Request extends Duplex implements RequestEvents<Request> {
...
+		if (this.options.signal) {
+			const abort = () => {
+				this.destroy(new AbortError(this));
+			};
 
-		this.options.signal?.addEventListener('abort', () => {
-			this.destroy(new AbortError(this));
-		});
+			if (this.options.signal.aborted) {
+				abort();
```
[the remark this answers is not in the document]
marcelmeulemans: Good catch, there would still be leak for an already aborted signal. Are you sure about the return though because the `if (is.nodeStream(body)) {` branch on line 270 is also not executed anymore? Alternatively the `addEventListener` would go in an `else`.

source/core/index.ts
marcelmeulemans: `xo` does not like this suggestion without reassigning `_removeListeners` somewhere. Should it be reset to `noop` somewhere?

source/core/index.ts
```diff
@@ -199,6 +200,7 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 		this._unproxyEvents = noop;
 		this._triggerRead = false;
 		this._cancelTimeouts = noop;
+		this._removeListeners = noop;
```
sindresorhus, line 203: @szmarczak Unrelated, but why are we assigning here instead of just giving the class fields a new value directly?

szmarczak: I guess I'm just not used to this. I agree, it looks much better if the values were given directly :)

source/core/index.ts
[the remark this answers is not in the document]
marcelmeulemans: Done

szmarczak (approved):

marcelmeulemans:
Any action required from my end here?

I thought of a unit test for this and came up with this: https://github.com/marcelmeulemans/got/commit/d150e4f77a5fd72babf199a8056313921eb6689a
It works well, but the approach is not great because it is kind of based on knowledge of how the controller is used internally. I can squash it in if you guys want.

sindresorhus:
> I can squash it in if you guys want.

That would be great 👍

sindresorhus:
Thanks :)

PR #2163, opened 2022-10-11
Improve TypeScript 4.9 compatibility

RyanCavanaugh:
TypeScript 4.9 can narrow more correctly with the `in` operator. Since this property is declared as being always present, code that only runs if it's *not* present marks that variable as `never` and causes errors. The correct definition is to mark this as optional (alternatively, you can upcast `reentry` to `symbol` in the `in` checks, but that's more invasive and less correct)


sindresorhus:
Thanks :)

PR #2187, opened 2022-11-23
fix: destroy request object after successful response (issue #1489)

### LINKED ISSUE #1489: The onCancel handler was attached after the promise settled
#### Describe the bug

 - Node.js version: 14.13
 - OS & version: MacOS Catalina

`The onCancel handler was attached after the promise settled`

#### Actual behavior

This unhandled exception is not catch by a try/catch, but I see that it happen after a timeout error like this issue #1460 

#### Checklist

nduthoit:
#### Checklist

- [x] (n/a) If it's a new feature, I have included documentation updates in both the README and the types.

Based on my testing this fixes #1489 for v11: late `ECONNRESET` errors no longer causes the "onCancel handler was attached after the promise settled" error. (I used [this gist](https://gist.github.com/nduthoit/00b0623a6b4a9a61f1e264ea1f982d33) for testing).

It mimics [what is done in v12](https://github.com/sindresorhus/got/blob/v12.5.3/source/as-promise/index.ts#L111) by calling `request.destroy()` after a successful response right before the promise is resolved.

I tried to add a test case where the server would reset the TCP connection and cause an ECONNRESET error after sending a success response but that proved to be challenging to do using an Express server. I also tried using a Python server executed from the test case but it proved to be too brittle to be a good test case. Therefore I landed on a simpler test case that ensures that the request has been destroyed once the promise has successfully resolved.

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -131,6 +131,7 @@ export default function asPromise<T>(normalizedOptions: NormalizedOptions): Canc
 					return;
 				}
 
+				request.destroy();
 				resolve(request.options.resolveBodyOnly ? response.body as T : response as unknown as T);
 			});
 
--- a/test/promise.ts
+++ b/test/promise.ts
@@ -95,3 +95,14 @@ test('promise.json() does not fail when server returns an error and throwHttpErr
 	const promise = got('', {throwHttpErrors: false});
 	await t.notThrowsAsync(promise.json());
 });
+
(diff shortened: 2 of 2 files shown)


szmarczak (approved):

sindresorhus:
@nduthoit Looks like the `v11` branch no longer builds. So you need to fix the build issues if you want us to do a release.

sindresorhus:
https://github.com/sindresorhus/got/releases/tag/v11.8.6

PR #2200, opened 2022-12-22
Do not enforce newest URI rules on URLs

xhaggi:
There are still sites in the wild that only support Latin1 encoded URLs, but GOT throws a URI malformed error even though the page can be retrieved. This PR removes the assertion made with `decodeURI` to solve the problem.

My question would be, why is `decodeURI` used here for assertion?

Samples:
* http://www.wrgag.ch/Inserat_Servicemonteur_L%FCftung_Klima.pdf

#### Checklist


xhaggi:
@szmarczak it would be nice if we could backport this fix into v11.

PR #2306, opened 2023-10-26
chore: make followRedirect option accept functions

VanTigranyan:
#### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage
 .nyc_output
 dist
 *.0x
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ coverage
 .nyc_output
 dist
 *.0x
-.idea
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "got",
-	"version": "13.0.0",
+	"version": "13.1.0",
 	"description": "Human-friendly and powerful HTTP request library for Node.js",
 	"license": "MIT",
 	"repository": "sindresorhus/got",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "got",
-	"version": "13.1.0",
+	"version": "13.0.0",
 	"description": "Human-friendly and powerful HTTP request library for Node.js",
 	"license": "MIT",
 	"repository": "sindresorhus/got",
--- a/source/core/options.ts
+++ b/source/core/options.ts
@@ -1756,7 +1756,8 @@ export default class Options {
 	}
 
 	/**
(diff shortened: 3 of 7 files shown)


VanTigranyan:
With these changes, it is possible to pass the `followRedirect` option as a function that gets a PlainResponse right before the redirect is done and returns true or false. By that, it is possible to decide whether to redirect dynamically. The option supports both plain boolean and function type, allowing users to decide for themselves which type they prefer.

source/core/options.ts
```diff
@@ -1763,12 +1763,12 @@ export default class Options {
 
 	@default true
 	*/
-	get followRedirect(): boolean {
+	get followRedirect(): boolean | ((response: PlainResponse) => boolean) {
```
sindresorhus, line 1766: The function needs to be documented (with text) here and in the readme.

.gitignore
```diff
@@ -4,3 +4,4 @@ coverage
 .nyc_output
 dist
 *.0x
+.idea
```
sindresorhus, line 7: This doens't belong here.

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "got",
-	"version": "13.0.0",
+	"version": "13.1.0",
```
sindresorhus, line 3: This should be done by the maintainers.

sindresorhus (changes requested):

source/core/options.ts
VanTigranyan: By documented you mean I should add a note how to use it?

.gitignore
VanTigranyan: Will remove

package.json
VanTigranyan: Got it. Would be nice to have a CONTRIBUTING.md with mentions what and where we should do. I'll remove this.

VanTigranyan:
@sindresorhus Made the fixes you asked for. Thanks for the quick review.

VanTigranyan:
@sindresorhus BTW: I would be happy to contribute by adding a semantic release setup that would increase the version of the lib based on commit messages and would publish automatically.

package.json
sindresorhus: I don't see the need to document basic contribution things in every project. Generally, you should try to touch the least amount of code and config.

sindresorhus:
> I would be happy to contribute by adding a semantic release setup that would increase the version of the lib based on commit messages and would publish automatically.

No thanks. I prefer the human touch to relasing and changelogs.

sindresorhus (approved):

package.json
VanTigranyan: Got it. Thanks.

PR #2313, opened 2023-11-29
Require Node.js 20

sindresorhus:


panva:
Hello @sindresorhus,

I am surprised by this change, Node.js LTS 18.x is EOL in April 2025 and it is now not supported by `got` for, based on this PR, no apparent reason.

`--loader` / `--import` test-only Node.js flag can easily be resolved by [other means](https://github.com/panva/oauth4webapi/blob/9c35f0c373e9b302ea8c45e7d23e34110c1d5128/.node_flags.sh#L13-L17).

sindresorhus:
I usually strictly follow this convention of targeting the oldest active Node.js LTS version in my packages. However, this package is the exception because the HTTP part of Node.js is consistently buggy, and I don't have time to work around issues in older Node.js releases. I you need to still support Node.js 18, I suggest staying on Got v13, which is quite stable.

sindresorhus:
> The --loader / --import test-only Node.js flag can easily be resolved by [other means](https://github.com/panva/oauth4webapi/blob/9c35f0c373e9b302ea8c45e7d23e34110c1d5128/.node_flags.sh#L13-L17).

I'm aware of the workarounds. That just one of many things. I also have to provide support for Node.js 18 issues, and that takes time away from other things.

panva:
Thank you for your thoughts @sindresorhus.

Issue #2312, opened 2023-11-29
Allow typing the body of a `RequestError` response


dcaldastp:
#### What problem are you trying to solve?

I want to type the response body of a `RequestError`. Currently because `RequestError` does not accept a generic it's impossible to type the underlying response body.

```ts
export declare class RequestError extends Error {
    code?: string;
    stack: string;
    readonly options: NormalizedOptions;
    readonly response?: Response;
    readonly request?: Request;
    readonly timings?: Timings;
    constructor(message: string, error: Partial<Error & {
        code?: string;
    }>, self: Request | NormalizedOptions);
}
```

A valid workaround at the moment is something like:

```ts
type MyErrorBodyType = { reason: string; code: number; metadata: any }

type MyRequestError = Omit<RequestError, 'response'> & {
  response: Response<MyErrorBodyType>
}
```

This way ☝️ the `body` would be typed properly.

#### Describe the feature

I would like to be able to inject a type to the `Request`, like so 👇 

```ts
export declare class RequestError<T> extends Error {
    code?: string;
    stack: string;
    readonly options: NormalizedOptions;
    readonly response?: Response<T>;
    readonly request?: Request;
    readonly timings?: Timings;
    constructor(message: string, error: Partial<Error & {
        code?: string;
    }>, self: Request | NormalizedOptions);
}
```

whereby `T` is the type fo pass along to the respective response body.

#### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.

PR #2325, opened 2024-01-27
feat: allow typing the body of a RequestError response

katsanva:
#### Checklist

Made `RequestError`/`HTTPError` to be generic eg.

```ts
await got('https://google.com', { responseType: 'json' })
  .catch((err: HTTPError<{ code: number, message: string}>) => console.log(err.response.body.message));
```

Resolves https://github.com/sindresorhus/got/issues/2312

### DIFF
--- a/source/core/errors.ts
+++ b/source/core/errors.ts
@@ -16,13 +16,13 @@ function isRequest(x: unknown): x is Request {
 An error to be thrown when a request fails.
 Contains a `code` property with error class code, like `ECONNREFUSED`.
 */
-export class RequestError extends Error {
+export class RequestError<T = unknown> extends Error {
 	input?: string;
 
 	code: string;
 	override stack!: string;
 	declare readonly options: Options;
-	readonly response?: Response;
+	readonly response?: Response<T>;
 	readonly request?: Request;
 	readonly timings?: Timings;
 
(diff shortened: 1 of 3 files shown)


source/core/errors.ts
```diff
@@ -89,8 +89,8 @@ An error to be thrown when the server response code is not 2xx nor 3xx if `optio
 Includes a `response` property.
 */
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export class HTTPError extends RequestError {
-	declare readonly response: Response;
+export class HTTPError<T = any> extends RequestError<T> {
```
sindresorhus, line 92:
```suggestion
export class HTTPError<T = unknown> extends RequestError<T> {
```

katsanva: This will be a breaking change, isn't it?
eg this code will no more work
```ts
await got('https://google.com', { responseType: 'json' })
  .catch((err: HTTPError) => console.log(err.response.body.foo)) // 'err.response.body' is of type 'unknown'.ts(18046)

```

sindresorhus: Right. Add a code comment at least, like

> TODO: Change this to `unknown` in the next major version.

Haylane-png (approved):

PR #2327, opened 2024-01-29
feat: add original error property

### LINKED ISSUE #2010: Add original error property

@szmarczak asked me to create a separate issue (https://github.com/sindresorhus/got/issues/1865#issuecomment-993725043).

Could you add a property on the `got` Error object like `original` that stores the original Error if there was one for example from `socks-proxy-agent`? It would really help with more involved error handling scenarios.

#### Describe the feature

More background is at https://github.com/sindresorhus/got/issues/1865.

The `cause` property could be used https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#differentiate_between_similar_errors since it is standard.

#### Checklist

katsanva:
Add root error as `RequestError['cause']`

Resolves #2010 

#### Checklist

### DIFF
--- a/documentation/8-errors.md
+++ b/documentation/8-errors.md
@@ -21,7 +21,7 @@ Read the article [here](async-stack-traces.md).
 
 > [!NOTE]
 > - The error codes may differ when the root error has a `code` property set.
-> - Root error will be propagated as is via [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property.
+> - The root error will be propagated as is via the [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property.
 
 ### `RequestError`
 
--- a/documentation/8-errors.md
+++ b/documentation/8-errors.md
@@ -19,8 +19,9 @@ All Got errors contain various metadata, such as:
 
 Read the article [here](async-stack-traces.md).
 
-**Note:**
+> [!NOTE]
 > - The error codes may differ when the root error has a `code` property set.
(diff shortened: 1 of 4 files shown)


test/error.ts
```diff
@@ -359,6 +361,21 @@ test.skip('the old stacktrace is recovered', async t => {
...
 
+test('should wrap got cause', async t => {
+	const error = await t.throwsAsync<RequestError>(got('https://github.com', {retry: {limit: 0}, timeout: {request: 1}}));
+	const cause = error?.cause as TimeoutError;
+	t.is(error?.code, cause.code);
+	t.is(error?.message, cause.message);
+});
+
+test('should wrap non-got cause', async t => {
+	const error = await t.throwsAsync<RequestError>(got('https://github.com', {retry: {limit: 0}, timeout: {read: 1}, agent: {https: new SocksProxyAgent('socks://your-name%40gmail.com:abcdef12345124@br41.nordvpn.com')}}));
```
sindresorhus, line 372: Can you do this test without adding two dev dependencies? Every added dependency adds a maintenance burden.

documentation/8-errors.md
```diff
@@ -19,8 +19,9 @@ All Got errors contain various metadata, such as:
 
 Read the article [here](async-stack-traces.md).
 
-**Note:**
+> [!NOTE]
 > - The error codes may differ when the root error has a `code` property set.
+> - Root error will be propagated as is via [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property.
```
sindresorhus, line 24:
```suggestion
> - The root error will be propagated as is via the [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property.
```

PR #2336, opened 2024-03-03
catch error in fall back to utf8

archvlad:
To reproduce error fetch big file and then read body of the response. You will receive the error but you won't be able to catch it with try-catch.

```javascript
import got from "got";

try {
  const res = await got("https://mp4-download.com/8k-5-MP4");
  console.log(res.body);
} catch (error) {
  console.log("Catch error:", error.message);
}
```

So what I add is try-catch block to `response.rawBody.toString()` as it is in `parseBody` method from `got\dist\source\core\response.js`

#### Checklist

### DIFF
--- a/source/as-promise/index.ts
+++ b/source/as-promise/index.ts
@@ -7,7 +7,7 @@ import {
 	type RequestError,
 } from '../core/errors.js';
 import Request from '../core/index.js';
-import {parseBody, isResponseOk, type Response} from '../core/response.js';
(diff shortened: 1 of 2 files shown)


sindresorhus:
This needs a test.

archvlad:
The problem occurred when I get a body larger than 512MB, causing `Buffer.toString()` to throw an error. So, I added test that creates buffer with size 512 MB and then sends it to the client.

archvlad:
Thank you for accepting my PR. I have little experience in participating in open source, so this is very valuable to me.

Issue #1126, opened 2020-03-20
Throw helpful error messages


kirillgroshkov:
#### What problem are you trying to solve?

I'm using Github API. On error (non-200 http status) `got` throws an HTTPError which is by default printed like this:
```
HTTPError: Response code 422 (Unprocessable Entity)
    at EventEmitter.<anonymous> (.../node_modules/got/dist/source/as-promise.js:118:31)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  name: 'HTTPError'
}
```

It gives me no information from the response body to be able to debug the issue (e.g on the local machine, or from the server logs in production).

#### Describe the feature

What I'd like to see is that it'd print the `response.body`, e.g by including it in Error.message (maybe limiting the maximum length, for sanity). Now I need to add this code to achieve the needed result:

```typescript
await got(...)
  .catch(err => {
    console.log((err as HTTPError).response.body)
  })
```

Then it prints like this:
```
{"message":"Reference already exists","documentation_url":"https://developer.github.com/v3/git/refs/#create-a-reference"}
```

And suddenly the error message becomes very useful.

I'm aware that I can use Hooks, extend my `got` instance with some error-handling hook to achieve that (that what I'm going to try now). But wouldn't everyone benefit from such feature enabled by default in `got`? Or behind a non-default configuration flag?
...

#### Checklist

- [x] I have read the documentation and made sure this feature doesn't already exist.

szmarczak:
https://github.com/sindresorhus/got/blob/master/readme.md#errors It's your responsibility to catch the error and read the useful information. Options aren't logged on purpose.

@sindresorhus I think each error message should contain the request method and its URL.

sindresorhus:
Agreed

yeo-yeo:
FWIW, I had the same problem as @kirillgroshkov today and would have found the proposed functionality useful and intuitive

kyler-hyuna:
Encountered this today. Error is thrown and no useful information can be derived from the error thrown. Not even the body returned from the request.

LeonardoRick:
I agree too. I'm starting to use the lib and that was my first unconfortable usecase. Would be awesome to receive formated erros by default.

Hazmi35:
The problem is we can't identify what kind of actual errors are really sent from the API, for example, 400 Bad Request could mean an invalid token but it could be mean something else too.

felixfbecker:
[...]

PR #2339, opened 2024-03-14
pass msg body to the error

fin-dogedev:
This PR addresses an issue where the server response is obscured when encountering a 400 status code. My colleague and I encountered difficulties due to the library's behavior of hiding the server response in such cases. This situation is particularly problematic as some servers provide valuable error messages within the response body for 400 status codes.

The changes included in this PR aim to enhance the handling of 400 response status codes by ensuring that the server response is properly surfaced and accessible to users. By exposing the full server response, we can better diagnose and address issues encountered during API interactions, leading to improved usability and developer experience.

#### Checklist

### DIFF
--- a/source/core/errors.ts
+++ b/source/core/errors.ts
@@ -96,7 +96,7 @@ export class HTTPError<T = any> extends RequestError<T> {
 	declare readonly timings: Timings;
 
 	constructor(response: PlainResponse) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
This is intentional. We could maybe make it opt-in with an option for now.

https://github.com/sindresorhus/got/issues/1126

PR #2353, opened 2024-05-29
feat: improve types with generic extend

### LINKED ISSUE #1117: How to correctly type an instance returned by `got.extend`

Given the following client:

```ts
const jsonClient = got.extend({
  responseType: 'json',
  resolveBodyOnly: true
});
```

Doing:
```ts
const response = await jsonClient('https://httpbin.org/json');
```
hints TypeScript to believe that `response` is of type `Response<string>` while it should be `unknown`.

How can I correctly type `jsonClient` to take into account my options?

#### Checklist

spence-s:
#### Checklist

This changes the way types work in Got by making got a generic function of the passed in options.

Fixes #1117 and opens up to some new generic patterns that may make the TS experience better. It doesn't look like there are any specific tests for types, but this was light enough change that it didn't take any code changes.

There shouldn't be really any "breaking" changed here, the types work a little different, but that shouldn't be noticeable to most I don't think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,8 @@
 	},
 	"ava": {
 		"files": [
-			"test/*"
+			"test/*",
+			"!test/*.types.ts"
 		],
 		"timeout": "10m",
 		"extensions": {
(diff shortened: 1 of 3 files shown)


sindresorhus:
It would be useful with some kind of type test that proves that this change actually works. Could just do some https://github.com/mmkal/expect-type assertions in one of the test files.

PR #2365, opened 2024-07-11
Make get-stream a devDependency

bjornstar:
I tried updating `get-stream` to the latest version and realized it was only being used in one place with a TODO to replace the usage so I went ahead and did that and moved it into devDependencies.

#### Checklist


source/core/index.ts
```diff
@@ -877,11 +876,8 @@ export default class Request extends Duplex implements RequestEvents<Request> {
 
 		try {
 			// Errors are emitted via the `error` event
-			const rawBody = await getStreamAsBuffer(from);
-
-			// TODO: Switch to this:
-			// let rawBody = await from.toArray();
-			// rawBody = Buffer.concat(rawBody);
+			const fromArray = await from.toArray();
+			const rawBody = fromArray.every(b => isBuffer(b)) ? Buffer.concat(fromArray) : Buffer.from(fromArray.join(''));
```
sindresorhus, line 880: Why are you checking every chunk for whether it's a buffer? I don't think it can be mixed.

bjornstar: You're probably right, should be safe to just check the first item. Updated!

Issue #2371, opened 2024-08-07
Abort error reasi


tal-rofe98:
#### Describe the bug

 - Node.js version: 
 - OS & version: 

#### Actual behavior

...

#### Expected behavior

...

#### Code to reproduce

```js
...
```

#### Checklist

- [ ] I have read the documentation.
- [ ] I have tried my code with the latest version of Node.js and Got.

PR #2388, opened 2024-11-02
fix: correct support for AbortSignal:timeout()

katsanva:
See #2371

This changes the rejected value to TimeotError when `AbortSignal:timeout()` is initially aborted.

As stated in [docs](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static#return_value):

> The signal will abort with its [AbortSignal.reason](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) property set to a TimeoutError [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) on timeout, or an AbortError [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) if the operation was user-triggered.

#### Checklist

### DIFF
--- a/test/abort.ts
+++ b/test/abort.ts
@@ -325,6 +325,7 @@ test('support setting the signal as a default option', async t => {
 	t.true(signalHandlersRemoved(), 'Abort signal event handlers not removed');
 });
 
+const timeoutErrorCode = 23;
 // See https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static
 test('support AbortSignal.timeout()', async t => {
 	const signal = AbortSignal.timeout(1);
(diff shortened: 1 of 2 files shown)


test/abort.ts
```diff
@@ -324,3 +324,54 @@ test('support setting the signal as a default option', async t => {
...
 });
+
+// See https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static
+test('support AbortSignal.timeout()', async t => {
+	const signal = AbortSignal.timeout(1);
+
+	const p = got('http://example.com', {signal});
+
+	await t.throwsAsync(p, {
+		code: 23,
```
sindresorhus, line 335: Would be better to use a constant here than a magic number.

tal-rofe98:
@sindresorhus 
This should have been marked as breaking change..

Issue #2248, opened 2023-04-06
Connect timeout fired even though underlying socket is actually connected


thegedge:
#### Describe the bug

 - Node.js version: 18.4.0
 - OS & version: Debian bullseye (we're using the `node:18.4.0-bullseye` docker image)

At scale, when the event loop is incredibly busy and/or blocked by larger sync tasks, timeouts can fire when the underlying event has occurred at a lower level.

One example of this is the `connect` timeout. Although the underlying socket connects very fast, the `setTimeout` that aborts the request fires before the I/O phase of the event loop has a chance to let the request know the connection was actually established.

#### Actual behavior

Request is aborted with `Timeout awaiting 'connect' for 50ms`

#### Expected behavior

Connection is successfully established

#### Code to reproduce

No easy way to produce. We started observing this at scale.

Here's the patch we've applied locally: https://github.com/sindresorhus/got/compare/main...thegedge:got:main

And here's a timeline of the `Timeout awaiting 'connect' for XYZms` errors from patching:

![timeline showing 1 connect timeout every minute and a deploy that eliminated these events](https://user-images.githubusercontent.com/833576/230252008-523a456c-43c3-4cd8-9fe1-f8fd17e3ce2d.png)

I understand we can go either way here. Ideally we wouldn't have to go through the event loop and could just ask the underlying socket "hey, are you connected" before aborting, but alas.

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

paulrutter:
@thegedge we are seeing similar behaviour. What version of got did you encounter this on?

We're seeing it since moving to latest 14.x,coming from latest 11.x.

thegedge:
We're on 11.8.5 right now, and I'm going to guess that was the version at the time of writing this too (can't yet upgrade much beyond that until we can properly deal with a mix of CJS and ESM, but newer versions of node is making that easier!).

paulrutter:
I'm using import statements to load got into a cjs environment in node 20.x, that allows us to upgrade to latest got again.

But now we're seeing intermittent timeouts we didn't see before.

paulrutter:
Did you create a PR for this fix, because it seems pretty straightforward and would solve the busy event loop issue.

thegedge:
I did not. I posted the fix in my own fork, but was hoping for some discussion or a "send us a PR" from the maintainers before firing that off. I guess that never happened, and this fell off my radar 🙈

[...]

PR #2399, opened 2025-01-30
Fix timeout error firing even though it's cancelled

paulrutter:
This is fixes https://github.com/sindresorhus/got/issues/2248, as supplied by [@thegedge](https://github.com/thegedge) earlier. See the original issue for a description of the issue and the proposed solution.

It's hard to add a unit test for this behavior, as it involves blocking the event loop for a certain amount of time.
Let me know if this is something that would be considered or if there are any concerns.

#### Checklist

### DIFF
--- a/source/core/timed-out.ts
+++ b/source/core/timed-out.ts
@@ -71,7 +71,7 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 	const {host, hostname} = options;
 
 	const timeoutHandler = (delay: number, event: string): void => {
-		// Use queueMicroTask to allow for any cancelled events to be handled first, 
+		// Use queueMicroTask to allow for any cancelled events to be handled first,
 		// to prevent firing any TimeoutError unneeded when the event loop is busy or blocked
 		queueMicrotask(() => {
 			if (!handled.has(event)) {
--- a/source/core/timed-out.ts
+++ b/source/core/timed-out.ts
@@ -51,7 +51,7 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 	request[reentry] = true;
 	const cancelers: Array<typeof noop> = [];
 	const {once, unhandleAll} = unhandler();
-	const handled: Record<string, boolean> = {};
+	const handled: Map<string, boolean> = new Map<string, boolean>();
 
 	const addTimeout = (delay: number, callback: (delay: number, event: string) => void, event: string): (typeof noop) => {
 		const timeout = setTimeout(callback, delay, delay, event) as unknown as NodeJS.Timeout;
@@ -59,7 +59,7 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 		timeout.unref?.();
 
 		const cancel = (): void => {
-			handled[event] = true;
+			handled.set(event, true);
 			clearTimeout(timeout);
 		};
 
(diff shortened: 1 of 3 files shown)


paulrutter:
I do see some failed tests, so probably this change is not entirely backwards compatible.
https://github.com/sindresorhus/got/actions/runs/13049814941/job/36407379410

source/core/timed-out.ts
```diff
@@ -51,13 +51,15 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 	request[reentry] = true;
 	const cancelers: Array<typeof noop> = [];
 	const {once, unhandleAll} = unhandler();
+	const handled: Record<string, boolean> = {};
```
sindresorhus, line 54: Should use a `Map`.

source/core/timed-out.ts
```diff
@@ -69,7 +71,11 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 	const {host, hostname} = options;
 
 	const timeoutHandler = (delay: number, event: string): void => {
-		request.destroy(new TimeoutError(delay, event));
+		setTimeout(() => {
```
sindresorhus, line 74: Needs a short code comment why it's wrapped in this.

`queueMicrotask` may be a better choice here.

sindresorhus (changes requested):

source/core/timed-out.ts
```diff
@@ -69,7 +71,13 @@ export default function timedOut(request: ClientRequest, delays: Delays, options
 	const {host, hostname} = options;
 
 	const timeoutHandler = (delay: number, event: string): void => {
-		request.destroy(new TimeoutError(delay, event));
+		// Use queueMicroTask to allow for any cancelled events to be handled first,
+		// to prevent firing any TimeoutError unneeded when the event loop is busy or blocked
+		queueMicrotask(() => {
```
thegedge, line 76: I don't think `queueMicrotask` will work.

CleanShot 2025-01-30 at 10 03 36@2x

[source](https://miro.medium.com/v2/resize:fit:4800/format:webp/1*2yXbhvpf1kj5YT-m_fXgEQ.png)

Ignoring the fact that the above image isn't a perfect representation of what's happening (in Node, other runtimes may do something even more different), what we need to happen is to go around the outer loop. Since we're in the "timer callbacks" phase when we enter this handler, by queueing up another `setTimeout` we're going to do one full outer loop. This means we process I/O events (like connections!) and guarantee whether or not the connection truly was established.

thegedge: Given we've had this patch running in production for a long time, I'm inclined to say it's just the tests that need some tweaking 🙂 

I took a super quick look, and two things stand out to me:
1. `socket timeout` appears to set a custom `setTimeout` on the stream, which may not play nicely with this.
2. `send timeout` ticks a clock. Not sure how that works, but maybe we need two calls to tick to ensure the additional callback is called.

paulrutter: I also look into the tests, and the point is that (even though it works) it _does_ break backwards compatibility as before the `TimeoutError` would be thrown synchronously, and now it no longer is returned directly.

We can tune the tests so that is passes, but not sure if that would be acceptable or that we would need a different approach.

thegedge: Ah, fair point :)

Of course @sindresorhus can make the final call, I'd be happy to offer my two cents:

1. The original handler was already part of a `setTimeout`, and hence "async". There were no guarantees that it would trigger the timeout exactly as specified.
2. Doing another `setTimeout` does mean we extend the time for real connection timeouts a bit, since we have to go through the event loop one more time, but we were okay with that tradeoff (because it was sufficiently rare).
3. If backwards compatibility is a must, adding configuration is always an option!

paulrutter: Thanks for your input, appreciated.
I think it's up to @sindresorhus to make a fair judgement on this.

sindresorhus: I don't think anyone should have relied on it being sync. So I'm fine with you just adjusting the tests.

paulrutter: Ok, i will look into this 👍 thanks

paulrutter: I adjusted the tests so all pass now.

paulrutter: The tests are now fixed @sindresorhus @thegedge

Issue #2444, opened 2026-04-13
`got.stream()` hangs when response contains `Set-Cookie` and `cookieJar` is set


erri120:
#### Describe the bug

 - Node.js version: 22
 - OS & version: Linux

When using `got.stream()` with a `cookieJar` and the server responds with a `Set-Cookie header`, the stream never emits 'end' and the request hangs indefinitely.

#### Actual behavior

Stream hangs indefinitely. Removing `cookieJar` or removing `Set-Cookie` from the response both resolve the hang.

#### Expected behavior

Stream completes normally and the cookie is stored in the jar.

#### Code to reproduce

```js
test('stream reads a cookie and completes', withServer, async (t, server, got) => {
	server.get('/', (_request, response) => {
		response.setHeader('set-cookie', 'hello=world');
		response.end('body');
	});

	const cookieJar = new toughCookie.CookieJar();

	await new Promise<void>((resolve, reject) => {
		const stream = got.stream({cookieJar});
		stream.on('error', reject);
		stream.resume();
		stream.on('end', resolve);
	});

	const cookie = cookieJar.getCookiesSync(server.url)[0];
	t.is(cookie?.key, 'hello');
	t.is(cookie?.value, 'world');
});
```

#### Checklist

- [x] I have read the documentation.
- [x] I have tried my code with the latest version of Node.js and Got.

PR #2445, opened 2026-04-13
Add repro test for #2444

erri120:


erri120:
The test I added stalls because of #2444. I don't think the test runner is set up to handle timeouts. (nvm, failed after 10 minutes as expected).

PR #2447, opened 2026-04-20
Fix httpbin response examples in JSON and form docs

Bojun-Vvibe:
**Repo:** sindresorhus/got (⭐ 14000)
**Type:** docs
**Files changed:** 2
**Lines:** +7/-7

## What
This change corrects the `httpbin.org/anything` examples in the README and options documentation so they reference the response fields that `httpbin` actually returns. The JSON example now reads from `json`, and the form example now reads from `form`, with the expected output updated to match.

## Why
The previous snippets were misleading for readers who copy-paste examples while learning Got’s request body helpers. `httpbin` echoes raw bodies in `data`, parsed JSON in `json`, and form payloads in `form`, so the old examples could make correct Got usage look broken.

## Testing
Verified the rendered diff manually and ran `git diff --check` successfully. No automated test run was needed because this is a docs-only change.

## Risk
Low / Documentation-only patch with no runtime behavior changes.

### DIFF
--- a/documentation/2-options.md
+++ b/documentation/2-options.md
@@ -381,13 +381,13 @@ JSON **request** body. If set, the `content-type` header defaults to `applicatio
 ```js
 import got from 'got';
 
-const {data} = await got.post('https://httpbin.org/anything', {
+const {json} = await got.post('https://httpbin.org/anything', {
 	json: {
 		hello: 'world'
 	}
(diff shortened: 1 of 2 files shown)


sindresorhus:
Fully AI-generated PRs are not welcome.

PR #2456, opened 2026-06-17
fix: apply backoffLimit cap to total delay including noise

JSap0914:
#### Checklist

---

### Bug

`calculateRetryDelay` applies `Math.min` only to the exponential base component, leaving the noise contribution uncapped:

```ts
// Before (buggy)
const noise = Math.random() * retryOptions.noise;
return Math.min(((2 ** (attemptCount - 1)) * 1000), retryOptions.backoffLimit) + noise;
```

This allows the returned delay to exceed `backoffLimit` by up to `noise` ms on every invocation where the base already hit the cap.  With the default `noise` of `100` and a `backoffLimit` of `500`, **100 % of retries produce delays of 500–600 ms** — always above the documented upper bound.

### What the documentation says

From `documentation/7-retry.md`:

> #### `backoffLimit`
> The upper limit of the **`computedValue`**.
> By default, the `computedValue` is calculated in the following way:
>
> ```ts
> ((2 ** (attemptCount - 1)) * 1000) + noise
> ```

`backoffLimit` is documented as the upper limit of `base + noise`, not of `base` alone.

### Fix

Move `Math.min` to wrap the full expression so that the cap applies to the total delay:

```ts
// After (fixed)
const noise = Math.random() * retryOptions.noise;
return Math.min(((2 ** (attemptCount - 1)) * 1000) + noise, retryOptions.backoffLimit);
```

### Verification

New offline test file `test/calculate-retry-delay.ts` (runs without a live server):

```
npx xo source/core/calculate-retry-delay.ts test/calculate-retry-delay.ts  # no errors
npx tsc --noEmit                                                             # clean
NODE_OPTIONS='--import=tsx/esm' npx ava test/calculate-retry-delay.ts
```

Result — all 9 tests pass, including the regression test that previously failed:

```
✔ returns 0 when attemptCount exceeds limit
✔ returns 1 for RetryError regardless of other conditions
✔ returns 0 when method is not allowed
✔ returns 0 when neither error code nor status code matches
✔ respects retryAfter header value
✔ returns 0 when retryAfter exceeds computedValue
✔ returns 0 for status 413 without retryAfter
✔ delay never exceeds backoffLimit (noise included)   ← red before, green after
✔ exponential backoff grows with attempt count

9 tests passed
```

> Note: this fix was developed with AI assistance.

### DIFF
--- a/source/core/calculate-retry-delay.ts
+++ b/source/core/calculate-retry-delay.ts
@@ -36,7 +36,7 @@ const calculateRetryDelay: Returns<RetryFunction, number> = ({
 	}
 
 	const noise = Math.random() * retryOptions.noise;
-	return Math.min(((2 ** (attemptCount - 1)) * 1000), retryOptions.backoffLimit) + noise;
+	return Math.min(((2 ** (attemptCount - 1)) * 1000) + noise, retryOptions.backoffLimit);
 };
 
 export default calculateRetryDelay;
--- a/test/calculate-retry-delay.ts
+++ b/test/calculate-retry-delay.ts
@@ -1,0 +1,162 @@
+import test from 'ava';
+import calculateRetryDelay from '../source/core/calculate-retry-delay.js';
+import type {RetryOptions} from '../source/core/options.js';
+
+const makeError = (overrides: Record<string, unknown> = {}) => ({
+	name: 'RequestError',
+	code: 'ECONNRESET',
+	options: {method: 'GET'},
+	response: null,
+	...overrides,
+});
+
+const defaultRetryOptions: RetryOptions = {
+	limit: 10,
+	methods: ['GET'],
+	statusCodes: [408, 413, 429, 500, 502, 503, 504, 521, 522, 504],
+	errorCodes: [
+		'ETIMEDOUT',
+		'ECONNRESET',
+		'EADDRINUSE',
+		'ECONNREFUSED',
+		'EPIPE',
+		'ENOTFOUND',
+		'ENETUNREACH',
+		'EAI_AGAIN',
+	],
+	calculateDelay: () => 0,
+	backoffLimit: Number.POSITIVE_INFINITY,
+	noise: 100,
+	maxRetryAfter: undefined,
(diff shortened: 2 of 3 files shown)


sindresorhus:
CI is failing

JSap0914:
**Investigation update:** Added a missing test fix in d46ecef.

The existing `respects backoffLimit` test used `backoffLimit: 0`, which incidentally worked with the old buggy formula:

- **Old formula:** `Math.min(1000, 0) + noise = 0 + noise` → delay ≈ 0–100 ms (positive → retries happen)
- **New formula:** `Math.min(1000 + noise, 0) = 0` → delay = 0 → `if (backoff)` is falsy → no retry → 408 thrown

So the test was failing 3/3 on this branch, 0/3 on main — **deterministic, not flaky** — but for the wrong reason: the test was quietly depending on the broken behavior (noise escaping the cap) to keep the delay non-zero and therefore allow retries.

**Doc verdict:** `documentation/7-retry.md` is unambiguous:

```
computedValue = ((2 ** (attemptCount - 1)) * 1000) + noise
backoffLimit  = The upper limit of the computedValue.
```

Noise is part of `computedValue`; `backoffLimit` caps the whole thing. `Math.min(base + noise, backoffLimit)` is correct.

**Fix:** Changed the test from `backoffLimit: 0` (which with the correct formula means 'cap total to 0 = never retry') to `backoffLimit: 200` (a realistic cap: base 1000ms >> 200ms limit → delay capped to 200ms, well under the `< 400ms` assertion). Full `test/retry.ts` suite: 39/39 pass.

JSap0914:
After further review I'm withdrawing this. While `Math.min(base + noise, backoffLimit)` matches the documented definition of `backoffLimit` for typical values, it changes behavior at the `backoffLimit: 0` edge (delay becomes exactly 0, which the retry engine treats as falsy and stops retrying, whereas the previous code retried with the noise component). The practical difference for normal limits is only the noise term, so this isn't worth the behavioral edge change. Closing; apologies for the noise.

PR #2460, opened 2026-06-28
Preserve request body on cross-origin 307 and 308 redirects

### LINKED ISSUE #2459: 307 & 308 redirect should preserve body per RFC
`got` starting from version 15 started dropping request body when followRedirect: true

The comment is wrong in the source code
https://github.com/sindresorhus/got/blob/e5e645a7d6deeec02933bf474727a541775772c7/source/core/index.ts#L1088-L1089
and the shouldDropBody flag 
https://github.com/sindresorhus/got/blob/e5e645a7d6deeec02933bf474727a541775772c7/source/core/index.ts#L1093
should be false when http status code is 307 or 308 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/307
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/308

all other major http clients (including `ky`) I have tested retains the body on POST method when encountering 308 redirect.

sindresorhus:
307 and 308 must preserve the request method and body per RFC 9110, even across origins. Got was dropping the body on cross-origin redirects, producing a POST with an empty body, which matched neither the spec nor a GET conversion and diverged from every other client (browsers, undici, curl, ky all preserve the body and strip only credentials).

Fixes #2459

### DIFF
--- a/documentation/2-options.md
+++ b/documentation/2-options.md
@@ -772,7 +772,9 @@ Optionally, pass a function to dynamically decide based on the response object.
 #### **Note:**
 > - If a `303` is sent by the server in response to any request type (POST, DELETE, etc.), Got will request the resource pointed to in the location header via GET.\
 >  This is in accordance with the [specification](https://tools.ietf.org/html/rfc7231#section-6.4.4). You can optionally turn on this behavior also for other redirect codes - see [`methodRewriting`](#methodrewriting).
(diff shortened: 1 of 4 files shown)


tavanh2025123-netizen (commented):
9d2473

tavanh2025123-netizen (commented):
1890

